PR #3, opened 2020-03-06
Added metadata and build spec for publishing to SAR

svenngjeto:


template.yaml
```diff
@@ -5,6 +5,15 @@ Description: >
 
   Sample SAM Template for AWS
 
+Metadata:
+  AWS::ServerlessRepo::Application:
+    Name: nva-publication-api
```
janbj: NvaPublicationApi?

PR #5, opened 2020-03-12
Enabled ACCEPT_EMPTY_STRING_AS_NULL_OBJECT in ObjectMapper

janbj:


publication-commons/src/main/java/no/unit/nva/PublicationHandler.java
```diff
@@ -89,10 +89,11 @@ public static ObjectMapper createObjectMapper() {
         return new ObjectMapper()
                 .registerModule(new ProblemModule())
                 .registerModule(new JavaTimeModule())
-                .configure(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS, false)
-                .configure(SerializationFeature.INDENT_OUTPUT, true)
-                .setSerializationInclusion(JsonInclude.Include.NON_NULL)
-                .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+                .enable(DeserializationFeature.ACCEPT_EMPTY_STRING_AS_NULL_OBJECT)
+                .enable(SerializationFeature.INDENT_OUTPUT)
+                .disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES)
```
axthosarouris: I think for this one, it is better if the author of each Java bean decides himself about this with the tag `@JsonIgnoreProperties(ignoreUnknown = true)` but I do not have strong opinions about it.

publication-commons/src/main/java/no/unit/nva/PublicationHandler.java
```diff
@@ -89,10 +89,11 @@ public static ObjectMapper createObjectMapper() {
...
+                .disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS)
```
axthosarouris: is this for the milliseconds issue? Put is as a comment (simpler) or refactor as a fluent method

publication-commons/src/main/java/no/unit/nva/PublicationHandler.java
```diff
@@ -89,10 +89,11 @@ public static ObjectMapper createObjectMapper() {
...
+                .setSerializationInclusion(Include.NON_NULL);
```
axthosarouris: add comment `Ignore null fields` to make it more obvious

publication-commons/src/main/java/no/unit/nva/PublicationHandler.java
janbj: No, this to format as date time instead of unix timestamp

publication-commons/src/main/java/no/unit/nva/PublicationHandler.java
janbj: Yes, we can do it like that when you create the new lib

PR #13, opened 2020-03-30
Upgraded to datamodel v0.4.0

janbj:
Upgraded dependencies to datamodel v0.4.0 and updated test data.

### DIFF
--- a/fetch-publication/src/test/resources/publication.json
+++ b/fetch-publication/src/test/resources/publication.json
@@ -1,111 +1,114 @@
 {
-  "type": "Publication",
-  "createdDate": "2020-03-05T12:52:15.887Z",
-  "status": "Draft",
-  "handle": "http://example.org/handle/123",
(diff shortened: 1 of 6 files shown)


fetch-publication/src/test/resources/publication.json
```diff
@@ -1,111 +1,114 @@
...
-      }
-    ]
+    "contributors" : [ {
+      "type" : "Contributor",
+      "identity" : {
+        "type" : "Identity",
+        "id" : "http://example.org/person/123",
+        "name" : "Navnesen, Navn",
+        "nameType" : "Personal",
+        "orcId" : "orc123",
```
brinxmat: General comment, I feel that these should always be URIs and that we can probably do away with the naming of them — they're additional identifiers and we have a preference for ids from ARP and then ORCID and then "others".

janbj: Create a task to update datamodel?

PR #17, opened 2020-04-14
Np 754 get by owner returns only latest version

axthosarouris:


publication-commons/src/main/java/no/unit/publication/service/impl/DynamoDBPublicationService.java
```diff
@@ -1,5 +1,7 @@
 package no.unit.publication.service.impl;
 
+import static no.unit.publication.PublicationHandler.ENVIRONMENT_VARIABLE_NOT_SET;
```
axthosarouris, line 3: automatic re-arrangement of imports

PR #18, opened 2020-04-16
Np 777 update datamodel

nraanes:


template.yaml
```diff
@@ -106,8 +106,8 @@ Resources:
                 httpMethod: POST
                 type: AWS_PROXY
               responses:
-                '200':
-                  description: OK
+                '202':
+                  description: ACCEPTED
```
janbj, line 110: We can use a more human readable description here, for example "Publication modified successfully".

PR #19, opened 2020-04-21
Publish Publication feature

janbj:


publication-commons/src/main/resources/features/publish_publication.feature
```diff
@@ -0,0 +1,30 @@
+Feature: Publish a Publication
+
+  Scenario: The User publishes a Publication
+    Given the User wants to publish a Publication
```
brinxmat, line 4: Valid credentials and publishing permission?

publication-commons/src/main/resources/features/publish_publication.feature
```diff
@@ -0,0 +1,30 @@
...
+    When they set the Accept header to "application/json"
+    And they set the Authentication header to a Bearer token with their <credentials>
```
brinxmat, line 6: This makes the scenario a scenario outline, the example table is missing

publication-commons/src/main/resources/features/publish_publication.feature
```diff
@@ -0,0 +1,30 @@
...
+    And they request POST /publication/{identifier}/publish
+    Then they receive a response with status code 202
+    And they see that the response Content-type is "application/json"
```
brinxmat, line 9: Perhaps a location header that corresponds to the location of the published version of the document?

publication-commons/src/main/resources/features/publish_publication.feature
```diff
@@ -0,0 +1,30 @@
...
+    And they see the response body is a JSON object with the message "Publication is being published. This may take a while."
+
+  Scenario: The User checks status on a published Publication
+    Given the User has published the Publication
```
brinxmat, line 13: Does this imply idempotency? If so, I suspect that the method we should use is PUT, not POST

publication-commons/src/main/resources/features/publish_publication.feature
```diff
@@ -0,0 +1,30 @@
...
+
+  Scenario: The User checks status on a published and indexed Publication
+    Given the User has published the Publication
+    And the Publication is indexed
+    When they set the Accept header to "application/json"
+    And they set the Authentication header to a Bearer token with their <credentials>
+    And they request POST /publication/{identifier}/publish
+    Then they receive a response with status code 200
+    And they see that the response Content-type is "application/json"
+    And they see the response body is a JSON object with the message "Publication is published."
```
brinxmat, line 30: Error states such as invalid data need to be reported — in our current case, the Publication must have an owner, a title and a content (be that a file or a link)

publication-commons/src/main/resources/features/publish_publication.feature
janbj: For example?

publication-commons/src/main/resources/features/publish_publication.feature
janbj: Yes, you are right.

publication-commons/src/main/resources/features/publish_publication.feature
janbj: Who has permissions, the owner and the curator?

publication-commons/src/main/resources/features/publish_publication.feature
janbj: Example credentials are useless I think. I will change it.

publication-commons/src/main/resources/features/publish_publication.feature
janbj: Those are the only requirements?

publication-commons/src/main/resources/features/publish_publication.feature
brinxmat: A question for the PO…

publication-commons/src/main/resources/features/publish_publication.feature
brinxmat: I think in this case, just returning the URI for the publication is enough — it is there that the resource will show changes in status

publication-commons/src/main/resources/features/publish_publication.feature
```diff
@@ -0,0 +1,74 @@
+Feature: Publish a Publication
+
+  Scenario: A User without required permissions tries to publish a Publication
+    Given a User wants to publish a Publication
+    But they are missing the required permissions
```
brinxmat, line 5: If And/But are interchangeable, then I guess this is fine (I have never used it, and the documentation doesn't explicitly mention this usage)

publication-commons/src/main/resources/features/publish_publication.feature
```diff
@@ -0,0 +1,74 @@
...
+    When they set the Accept header to "application/json"
+    And they set the Authentication header to a Bearer token with their credentials
+    And they request PUT /publication/{identifier}/publish
+    Then they receive a response with status code 202
+    And they see that the response Content-type is "application/json"
+    And they see that the response Location header is the URI to the Publication
+    And they see the response body has a field "message" with the value "Publication is being published. This may take a while."
+
+    Examples:
+      | User |
```
brinxmat, line 26: Formatting :)

publication-commons/src/main/resources/features/publish_publication.feature
```diff
@@ -0,0 +1,74 @@
...
+  Scenario: The Owner publishes a Publication with a malformed identifier
+    Given the Owner wants to publish a Publication
+    When they set the Accept header to "application/json"
+    And they set the Authentication header to a Bearer token with their credentials
+    And they request PUT /publication/{malformed identifier}/publish
+    Then they receive a response with status code 400
+    And they see that the response Content-Type header is "application/problem+json"
+    And they see that the response body is a problem.json object
+    And they see the response body has a field "title" with the value "Bad Request"
+    And they see the response body has a field "status" with the value "400"
```
brinxmat, line 40: Should the detail contain the URI and the specific error?

publication-commons/src/main/resources/features/publish_publication.feature
```diff
@@ -0,0 +1,74 @@
...
+
+  Scenario: The Owner checks status on a published Publication
```
brinxmat, line 42: I think this is a misnamed — they're resubmitting the request on an already submitted request. While this has no effect, in terms of semantics, I'd look at the location and check the status if I was "checking"

publication-commons/src/main/resources/features/publish_publication.feature
```diff
@@ -0,0 +1,74 @@
...
+    But the Publication is still not indexed
+    When they set the Accept header to "application/json"
+    And they set the Authentication header to a Bearer token with their credentials
+    And they request PUT /publication/{identifier}/publish
+    Then they receive a response with status code 202
+    And they see that the response Content-type is "application/json"
+    And they see that the response Location header is the URI to the Publication
+    And they see the response body has a field "message" with the value "Publication is being published. This may take a while."
+
+  Scenario: The Owner checks status on a published and indexed Publication
```
brinxmat, line 53: Similarly…

publication-commons/src/main/resources/features/publish_publication.feature
```diff
@@ -0,0 +1,74 @@
...
+  Scenario: The Owner publishes an invalid Publication
+    Given the Owner wants to publish an invalid Publication
+    When they set the Accept header to "application/json"
+    And they set the Authentication header to a Bearer token with their credentials
+    And they request PUT /publication/{identifier}/publish
+    Then they receive a response with status code 500
+    And they see that the response Content-Type header is "application/problem+json"
+    And they see that the response body is a problem.json object
+    And they see the response body has a field "title" with the value "Internal Server Error"
+    And they see the response body has a field "status" with the value "500"
```
brinxmat, line 73: Is this a server error or a bad request?

publication-commons/src/main/resources/features/publish_publication.feature
brinxmat: I guess this requires a feature for checking the Location URI and asserting that the status has changed

publication-commons/src/main/resources/features/publish_publication.feature
janbj: Good question, it's not really part of the request but stored in the DB

publication-commons/src/main/resources/features/publish_publication.feature
```diff
@@ -0,0 +1,86 @@
...
+    And they see that the response body is a Publication JSON object
+    And they see that the response body has a field "status" with the value "PUBLISHED"
+    And they see that the response body has a field "indexedDate" with a value "{datetime}"
+
+  Scenario: The Owner publishes an invalid Publication
+    Given the Owner wants to publish an invalid Publication
+    When they set the Accept header to "application/json"
+    And they set the Authentication header to a Bearer token with their credentials
+    And they request PUT /publication/{identifier}/publish
+    Then they receive a response with status code 422
```
brinxmat, line 81: I like your thinking, but we are not processing an entity here. I think 409 is a better option, or 424

publication-commons/src/main/resources/features/publish_publication.feature
```diff
@@ -0,0 +1,86 @@
...
+    And they see that the response Content-Type header is "application/problem+json"
+    And they see that the response body is a problem.json object
+    And they see the response body has a field "title" with the value "Unprocessable Entity"
+    And they see the response body has a field "status" with the value "422"
+    And they see the response body has a field "detail" with the value "Publication is missing data required for publishing."
```
brinxmat, line 86: I think we should give an explicit response containing a validation report, noting what needs to be fixed…I guess it is enough for now to say it contains the string "The publication cannot be published because the following fields are not populated" 
And the missing fields are listed

PR #21, opened 2020-04-24
Np 768 Refactor before I begin on publish publication

janbj:


publication-commons/src/test/java/no/unit/nva/publication/service/impl/DynamoDBPublicationServiceTest.java
```diff
@@ -111,27 +133,89 @@ public void callingConstructorWithAllEnv() {
         new DynamoDBPublicationService(client, objectMapper, environment);
     }
 
+    @Test
+    public void canGetPublication() throws Exception {
```
axthosarouris, line 137: This test although simple, is not easy to understand with the first reading. The title could be a mode descriptive and the variable names as well. By itself the test is not a problematic. But these small difficulties accumulate when one edits someone else's code or reviews code (especially long PRs).  

The line `Publication publication = publication();` is not informative of the role of this object in the test. I have to figure it out myself that it is the expected object.
The name "publicationFromDb" is not exactly correct. It is not necessarily what was in the database, but what the service retrieved. These are not necessarily the same thing. Otherwise we wouldn't be testing. So it is the "retrievedPublicaion"

It may seem nitpicking but clarity in code will help us understand our own code better and make refactoring and reviewing easier. 

It is not necessary to change the test, but ideally code should be **very** easy to read. 

```suggestion
    public void getPublicationReturnsExistingPublicationWhenInputIsExistingIdentifier() throws Exception {
    Publication storedPublication = publication();
    publication.createPublication(storedPublication);
    Publication retrievedPublication = publication.getPublication(storedPublication.getIdentifier());
    assertEquals(retrievedPublication,savedPublication):
```

publication-commons/src/test/java/no/unit/nva/publication/service/impl/DynamoDBPublicationServiceTest.java
```diff
@@ -111,27 +133,89 @@ public void callingConstructorWithAllEnv() {
...
+        publication.setOwner(newOwner);
+        Publication updatedPublication = publicationService.updatePublication(
+                publication.getIdentifier(), publication);
+        assertEquals(newOwner, updatedPublication.getOwner());
+    }
+
+    @Test
+    public void updateExistingCustomerChangesModifiedDate() throws Exception {
+        Publication publication = publication();
+        Publication createdPublication = publicationService.createPublication(publication);
```
axthosarouris, line 170: :+1 "createdPublication"

janbj: What do you mean?

PR #23, opened 2020-04-28
Np 826 publish publication api

janbj:


template.yaml
```diff
@@ -206,6 +270,30 @@ Resources:
             Path: /by-owner
             Method: get
             RestApiId: !Ref NvaPublicationApi
+  NvaPublishPublicationFunction:
```
janbj, line 273: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

You must specify a value that is greater than or equal to 128, and it must be a multiple of 64. You cannot specify a size larger than 3008. Error at Resources/NvaPublishPublicationFunction/Properties/MemorySize

publication-commons/src/main/java/no/unit/nva/publication/service/impl/DynamoDBPublicationService.java
```diff
@@ -231,4 +236,20 @@ private void allFieldsAreNonNull(String owner, URI publisherId) {
             return Optional.empty();
         }
     }
+
+    @Override
+    public PublishPublicationStatus publishPublication(UUID identifier) throws ApiGatewayException {
+        Publication publicationToPublish = getPublication(identifier);
+        if (validatePublicationPublished(publicationToPublish)) {
```
axthosarouris, line 243: rename to publishPublicationIsValid ?

PR #33, opened 2020-05-05
Np 865 using layers for publication commons

janbj:


build.gradle
```diff
@@ -1,5 +1,6 @@
 plugins {
     id 'java'
+    id 'java-library'
```
axthosarouris, line 3: do you need both 'java' and 'java-library'?

publication-commons/build.gradle
```diff
@@ -2,7 +2,22 @@ repositories {
     maven { url "https://s3.eu-central-1.amazonaws.com/dynamodb-local-frankfurt/release" }
 }
 
+def jacksonVersion = '2.10.2'
+def awsSdkVersion = '1.11.714'
+
 dependencies {
+    api group: 'com.github.BIBSYSDEV', name: 'nva-commons', version: 'v0.3.0'
```
axthosarouris, line 9: Are you sure that implementation does not do the job?

janbj: Yes, I didn't get access to transitive dependencies when using implementation

build.gradle
janbj: Maybe not?

PR #34, opened 2020-05-05
Keeping identifier from input

janbj:


publication-commons/src/main/java/no/unit/nva/publication/service/impl/DynamoDBPublicationService.java
```diff
@@ -84,15 +84,13 @@ public DynamoDBPublicationService(AmazonDynamoDB client, ObjectMapper objectMapp
...
-        UUID identifier = UUID.randomUUID();
         try {
-            publication.setIdentifier(identifier);
             Item item = publicationToItem(publication);
             table.putItem(item);
         } catch (Exception e) {
             throw new DynamoDBException(ERROR_WRITING_TO_TABLE, e);
         }
-        return getPublication(identifier);
+        return getPublication(publication.getIdentifier());
```
janbj, line 93: This is only a solution while I try to think of a better way to handle internal properties on Publication.

PR #41, opened 2020-05-20
NP-978: DynamoDBPublicationService not allowed access patterns feature description

axthosarouris:
First version of features. of DynamoDbPublicationService.

There should be no endpoint or method that would allow the reading or the inserting in the database of any file without providing a username.
We assume that the username comes from the claims

### DIFF
--- a/publication-commons/src/test/resources/features/NotAllowedAccessToDynamoWithoutUserCredentials.feature
+++ b/publication-commons/src/test/resources/features/NotAllowedAccessToDynamoWithoutUserCredentials.feature
@@ -2,8 +2,6 @@ Feature:
   DynamoDBPublicationService does not allow reading, inserting or modifying an entry without
   the caller of the service specifying the username who is performing the action.
 
-  Rule: User performing CRUD is always explicitly specified.
-
   Background:
     Given that PublicationService exists
 
--- a/publication-commons/src/test/resources/features/NotAllowedAccessToDynamoWithoutUserCredentials.feature
+++ b/publication-commons/src/test/resources/features/NotAllowedAccessToDynamoWithoutUserCredentials.feature
@@ -1,0 +1,44 @@
+Feature:
+  DynamoDBPublicationService does not allow reading, inserting or modifying an entry without
+  the caller of the service specifying the username who is performing the action.
+
+
+  Background:
+    Given that PublicationService exists
+
+  Scenario: PublicationService requires username for reading a publication
+    Given that PublicationService provides a READ method for reading a single publication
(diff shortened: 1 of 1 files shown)


publication-commons/src/test/resources/features/NotAllowedAccessToDynamoWithoutUserCredentials.feature
```diff
@@ -0,0 +1,44 @@
...
+
+
+  Background:
+    Given that PublicationService exists
+
+  Scenario: PublicationService requires username for reading a publication
+    Given that PublicationService provides a READ method for reading a single publication
+    And that READ method is being called by a module or an application
+    When the READ call does not include a username
+    Then the module or the application cannot call the READ method
```
janbj, line 13: Or the read only produces results with a match on username?

axthosarouris: This will be on the features regarding the returning results

brinxmat: If the Publication is PUBLISHED, then anyone can read it

publication-commons/src/test/resources/features/NotAllowedAccessToDynamoWithoutUserCredentials.feature
```diff
@@ -0,0 +1,76 @@
...
+  Scenario: PublicationService requires non empty username for reading a publication
+    Given that PublicationService provides a READ method for reading a single publication
+    And that READ method is being called by a module or an application
+    When the READ call includes a username as a parameter
+    And the parameter is null or empty
+    Then the READ method returns an error message that empty values are not allowed
+
+
+  Scenario: PublicationService requires username for listing a set of publications
+    Given that PublicationService provides a LIST method for reading a set of publications
```
brinxmat, line 24: Again, if the thing. is published, it. can be listable, c.f. listing publications on public profile

axthosarouris:
As we discussed before with @brinxmat, the publication service is the way to perform CRUD operations regarding publication data and metadata. Since at the moment we have a Lambda function that can access and modify everything, we should be very strict with our implementation.

My suggestion is that we do not allow CRUD operations on the database to be performed without specifying the username of the user or "Anonymous" if it is an anonymous user. In this way, we will never try to read/write without making an explicit decision about who is reading/writing. 

By "not allow" I mean if someone tries to access data without specifying the user, one gets a compilation error and not a runtime error.

PR #42, opened 2020-05-21
Np 978 read published material 

axthosarouris:
Feature description for reading published publications.

This feature describes what happens only in the case of a published publication.

### DIFF
--- a/publication-commons/src/test/resources/features/AccessedToPublishedMaterial.feature
+++ b/publication-commons/src/test/resources/features/AccessedToPublishedMaterial.feature
@@ -11,7 +11,6 @@ Feature:
 
   2. We assume that when a user is logged in, he/she has read access to all material that is published.
 
-
   Background:
     Given that DynamoDBPublicationService exists
     And the DynamoDBPublicationService has a READ method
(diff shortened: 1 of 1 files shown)


axthosarouris:
For the moment I do not know whether CURATORS or other roles should have the right to edit published publications.

publication-commons/src/test/resources/features/AccessedToPublishedMaterial.feature
```diff
@@ -0,0 +1,64 @@
+Feature:
+  1. DynamoDBPublicationService allows a user to read any document that is marked as published and
+  therefore it is publicly available.
+
+  2. DynamoDBPublicationService does not allow any user to modify or delete any document that is marked
```
brinxmat, line 5: While I would think this is a natural assumption, any owner, curator, editor or Admin can modify a published publication associated with their institution. This is because there needs to be an opportunity of correcting mistakes.

Consider also that deletion of resources with status PUBLISHED is considered to be different to other statuses, where a soft/logical deletion takes place in the former case and a physical deletion in the latter. This is not implemented as far as I can see.

Note that this means that the major distinction between the roles is one of access scope to documents in the workflow, rather than one of CRUD privileges. I will try to codify this statement in a document.

publication-commons/src/test/resources/features/AccessedToPublishedMaterial.feature
```diff
@@ -0,0 +1,64 @@
...
+  as published.
+
+  Notes:
+  1. By "published" we mean that a document has been published in a Journal, Conference proceedings,
```
brinxmat, line 9: Published in the sense of NVA means that it has status PUBLISHED, there is no requirement that it has simultaneously been published elsewhere — the data may be so-called "grey literature"

publication-commons/src/test/resources/features/AccessedToPublishedMaterial.feature
```diff
@@ -0,0 +1,64 @@
...
+  or is part of a book and any logged in user can access it.
+
+  2. We assume that when a user is logged in, he/she has read access to all material that is published.
```
brinxmat, line 12: We assume that when a user is logged in or has role Anonymous, they have read access to all material that is published.

publication-commons/src/test/resources/features/AccessedToPublishedMaterial.feature
```diff
@@ -0,0 +1,64 @@
...
+    Then the READ method returns the publication with ID "PubId"
+
+
+  Scenario Outline: User is forbidden to update/delete a published publication
+    Given that "PubId" is an existing publication ID
+    And that the publication with ID "PubId" has been published
+    And the owner of the publication is "theOwner"
+    When a call to the <action> method requests to update the publication with Id "PubId"
+    And the call to the <action> method makes the request on behalf of <callerUser>
+    Then the <action> method returns a response indicating that this operation is now allowed
```
brinxmat, line 55: s/now/not/

publication-commons/src/test/resources/features/AccessedToPublishedMaterial.feature
```diff
@@ -0,0 +1,64 @@
...
+  Scenario: An anonymous user reads a published publication
+    Given that "PubId" is an existing publication ID
+    And that the publication with ID "PubId" has been published
+    And the owner of the publication is "theOwner"
+    When a call to the READ method requests the publication with Id "PubId"
+    And the call to the READ method makes the request on behalf of an Anonymous user
+    Then the READ method returns the publication with ID "PubId"
+
+
+  Scenario Outline: User is forbidden to update/delete a published publication
```
brinxmat, line 49: The table for who can and cannot modify/delete a published resource needs to be defined. I will do this now.

PR #51, opened 2020-05-27
Setting identifier in createPublication

janbj:


axthosarouris:
can you also write a test that checks that we put an identifier?

PR #82, opened 2020-08-27
Remove duplicate Publications in response

janbj:


publication-commons/src/main/java/no/unit/nva/publication/service/impl/DynamoDBPublicationService.java
```diff
@@ -227,7 +227,8 @@ private void validateIdentifier(UUID identifier, Publication publication) throws
             throw new DynamoDBException(ERROR_READING_FROM_TABLE, e);
         }
 
-        return parseJsonToPublicationSummaries(items);
+        List<PublicationSummary> publications = parseJsonToPublicationSummaries(items);
+        return filterOutOlderVersionsOfPublications(publications);
```
axthosarouris, line 231: Could we have a test for that?

axthosarouris (commented):
test missing?

brinxmat (commented):
+1 on axthosarouris' comment

publication-commons/src/main/java/no/unit/nva/publication/service/impl/DynamoDBPublicationService.java
janbj: The method already has tests.

axthosarouris: I actually this is the problem which explains why `filterOutOlderVersionsOfPublications` is protected and not private. The module is the DynamoDBPublicationService and it is through the public methods of this module that we need to test the module functionality. 
The fact that we changed the functionality of a public method and no tests failed and no tests were changed is problematic.
`filterOutOlderVersionsOfPublications` is an implementation detail. This method should be private and there should be a test with a name somethiing like `listPublishedPublicationsByDateReturnsTheMostRecentVersionOfEachPublication`

brinxmat (approved):

axthosarouris (approved):

PR #85, opened 2020-09-21
Moved definition of dynamodb table that contains stuff defined in NVA…

svenngjeto:
…, table name is to be discussed.  Definition of table should be complete. Table will be deleted from NVA-infrstructure repo after finished QA


cloudformation/nva_resources_dynamodb.yml
```diff
@@ -0,0 +1,105 @@
+AWSTemplateFormatVersion: '2010-09-09'
```
axthosarouris, line 1: Shouldn't that be in the service template file? How it is going to be deployed?

axthosarouris (approved):

PR #87, opened 2020-09-29
Dynamodb inline

svenngjeto:
Definition of nva_publications is inline in template, table and index names as parameters with default values

### DIFF
--- a/template.yaml
+++ b/template.yaml
@@ -509,7 +509,7 @@ Resources:
               KeyType: RANGE
           Projection:
             ProjectionType: ALL
-        - IndexName: ByStatusPublishedDate
+        - IndexName: !Ref PublishedPublicationsIndexName
           KeySchema:
             - AttributeName: status
               KeyType: HASH
--- a/template.yaml
+++ b/template.yaml
@@ -300,7 +300,7 @@ Resources:
         Variables:
           ALLOWED_ORIGIN: '*'
           TABLE_NAME: !Ref PublicationsTable
-          BY_PUBLISHER_INDEX_NAME: !Ref PublicationsTable
+          BY_PUBLISHER_INDEX_NAME: !Ref PublicationsByOwnerIndexName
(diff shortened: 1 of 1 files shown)


axthosarouris (approved):
Approved on the condition that !Ref PublicationsTable returns the name and not the ARN of the table.

template.yaml
```diff
@@ -299,14 +299,14 @@ Resources:
       Environment:
         Variables:
           ALLOWED_ORIGIN: '*'
-          TABLE_NAME: !Ref PublicationsTableName
-          BY_PUBLISHER_INDEX_NAME: !Ref PublicationsByOwnerIndexName
+          TABLE_NAME: !Ref PublicationsTable
+          BY_PUBLISHER_INDEX_NAME: !Ref PublicationsTable
```
norrs, line 303: Intended?

svenngjeto: Nope..

template.yaml
```diff
@@ -66,7 +66,7 @@ Resources:
       DefinitionBody:
         swagger: '2.0'
         info:
-          title: NVA Publication API
+          title: NVA custom Publication API
```
norrs, line 69: Probably dont need the `custom` here either @svenng ?

norrs (approved):
nitpick, looks good.

template.yaml
```diff
@@ -299,14 +299,14 @@ Resources:
       Environment:
         Variables:
           ALLOWED_ORIGIN: '*'
-          TABLE_NAME: !Ref PublicationsTableName
+          TABLE_NAME: !Ref PublicationsTable
```
janbj, line 302: Maybe we should make the table name and index names internal to simplify stack setup?

EDIT: I guess it helps with default values.

janbj (approved):

template.yaml
svenngjeto: yes, using defaults without overriding them is poor man's constants

norrs:
@svenngjeto : Should we maybe see if we can get this merged? :)

PR #89, opened 2020-10-10
NP-1672 Add handling for Content-Range header

brinxmat:
Please refer to RFC 7231 whenever you define a new API so that we avoid having to make changes like this in future.

If a Content-Range header is present on a PUT request, according to RFC 7231 §4.3.4, the service should return 400 BAD REQUEST as the request indicates that the request contains partial content.

### DIFF
--- a/modify-publication/src/main/java/no/unit/nva/publication/modify/ModifyPublicationHandler.java
+++ b/modify-publication/src/main/java/no/unit/nva/publication/modify/ModifyPublicationHandler.java
@@ -7,6 +7,7 @@ import no.unit.nva.api.PublicationResponse;
 import no.unit.nva.api.UpdatePublicationRequest;
 import no.unit.nva.model.Publication;
 import no.unit.nva.publication.RequestUtil;
+import no.unit.nva.publication.modify.exception.PartialContentException;
 import no.unit.nva.publication.service.PublicationService;
 import no.unit.nva.publication.service.impl.DynamoDBPublicationService;
 import nva.commons.exceptions.ApiGatewayException;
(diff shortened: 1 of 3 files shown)


modify-publication/src/main/java/no/unit/nva/publication/modify/ModifyPublicationHandler.java
```diff
@@ -20,6 +21,7 @@
 import java.util.UUID;
 
 import static nva.commons.utils.JsonUtils.objectMapper;
+import static org.apache.http.HttpHeaders.CONTENT_RANGE;
 
 public class ModifyPublicationHandler extends ApiGatewayHandler<UpdatePublicationRequest, PublicationResponse> {
```
norrs, line 26: Could maybe ApiGatewayHandler itself do the verifying?

default method and ensure we call super?

like: if method.PUT && !validateNotPartialContentReuqest  ; bad boy.

norrs (changes requested):
I guess we might end up with adding more default validations later if API gateway from Amazon lets em thru. We shouldnt need to remember to check this manually for each handler we have.

modify-publication/src/main/java/no/unit/nva/publication/modify/ModifyPublicationHandler.java
axthosarouris: Maybe we should have have a PutRequestHander and a (GetRequestHandler etc.) where each one of them checks for valid requests in their context (PUT,GET,POST, etc) requests. So we can void the clutter of 
```
if(method.name==something){
 validate()
}

axthosarouris (approved):
Approved under the condition that we create a task for validating requests.

norrs:
@brinxmat Link to tech debt if not fixing it right away.

janbj (approved):

PR #96, opened 2020-11-16
Data type for Digital Object Identifier (DOI)

norrs:
To be used across NVA to represent a Doi.

### DIFF
--- a/publication-commons/src/main/java/no/unit/nva/doi/models/Doi.java
+++ b/publication-commons/src/main/java/no/unit/nva/doi/models/Doi.java
@@ -55,15 +55,13 @@ public abstract class Doi {
      */
     public URI toUri() {
         try {
-            URI uri = createDoi();
-            uri.toURL(); // validate
-            return uri;
-        } catch (MalformedURLException | URISyntaxException e) {
+            return createDoi();
+        } catch (URISyntaxException e) {
             throw new IllegalStateException(ERROR_PROXY_URI_MUST_BE_A_VALID_URL, e);
         }
     }
 
-    private URI createDoi() throws URISyntaxException {
+    protected URI createDoi() throws URISyntaxException {
         return new URI(getProxy().getScheme(),
             getProxy().getUserInfo(),
             getProxy().getHost(),
--- a/publication-commons/src/main/java/no/unit/nva/doi/models/Doi.java
+++ b/publication-commons/src/main/java/no/unit/nva/doi/models/Doi.java
@@ -12,6 +12,7 @@ import java.util.List;
  */
 public abstract class Doi {
 
+    public static final String HANDLE_DOI_PREFIX = "10.";
     public static final String ERROR_PROXY_URI_MUST_BE_A_VALID_URL = "Proxy URI must be a valid URL";
     protected static final String DOI_ORG = "doi.org";
     protected static final String HANDLE_STAGE_DATACITE_ORG = "handle.stage.datacite.org";
--- a/publication-commons/src/main/java/no/unit/nva/doi/models/Doi.java
+++ b/publication-commons/src/main/java/no/unit/nva/doi/models/Doi.java
@@ -57,6 +57,7 @@ public abstract class Doi {
      * @see java.net.URL#toExternalForm
      */
     @JacocoGenerated
+    @SuppressWarnings("PMD.UseStringBufferForStringAppends") // since we copy JDK method!
     private String extractSchemeWithAuthorityAndHost() {
         try {
             var proxyUrl = getProxy().toURL();
--- a/publication-commons/src/main/java/no/unit/nva/doi/models/Doi.java
+++ b/publication-commons/src/main/java/no/unit/nva/doi/models/Doi.java
@@ -2,8 +2,8 @@ package no.unit.nva.doi.models;
 
 import java.net.MalformedURLException;
 import java.net.URI;
-import java.util.List;
 import java.net.URISyntaxException;
+import java.util.List;
 
 /**
  * Doi class for working with Dois.
@@ -13,14 +13,18 @@ import java.net.URISyntaxException;
 public abstract class Doi {
 
     public static final String ERROR_PROXY_URI_MUST_BE_A_VALID_URL = "Proxy URI must be a valid URL";
-    public static final String DOI_ORG = "doi.org";
-    public static final String HTTPS = "https://";
-    public static final String HANDLE_STAGE_DATACITE_ORG = "handle.stage.datacite.org";
-    public static final String DX_DOI_ORG = "dx.doi.org";
+    protected static final String DOI_ORG = "doi.org";
+    protected static final String HANDLE_STAGE_DATACITE_ORG = "handle.stage.datacite.org";
+    protected static final String DX_DOI_ORG = "dx.doi.org";
     public static final List<String> VALID_PROXIES = List.of(DOI_ORG, DX_DOI_ORG, HANDLE_STAGE_DATACITE_ORG);
+    protected static final String HTTPS = "https";
+    protected static final String HTTP = "http";
+    public static final List<String> VALID_SCHEMES = List.of(HTTPS, HTTP);
     protected static final char PATH_SEPARATOR = '/';
     protected static final String PATH_SEPARATOR_STRING = String.valueOf(PATH_SEPARATOR);
-    public static final URI DOI_PROXY = URI.create(HTTPS.concat(DOI_ORG).concat(PATH_SEPARATOR_STRING));
+    protected static final String SCHEMA_SEPARATOR = "://";
+    public static final URI DEFAULT_DOI_PROXY = URI.create(
+        HTTPS.concat(SCHEMA_SEPARATOR).concat(DOI_ORG).concat(PATH_SEPARATOR_STRING));
 
     public static ImmutableDoi.Builder builder() {
         return ImmutableDoi.builder();
@@ -31,8 +35,7 @@ public abstract class Doi {
     public abstract String getSuffix();
 
     public URI getProxy() {
-        // default value
-        return DOI_PROXY;
+        return DEFAULT_DOI_PROXY;
     }
 
     /**
--- a/publication-commons/src/main/java/no/unit/nva/doi/models/Doi.java
+++ b/publication-commons/src/main/java/no/unit/nva/doi/models/Doi.java
@@ -1,0 +1,45 @@
+package no.unit.nva.doi.models;
+
+import java.net.URI;
+
+/**
+ * Doi class for working with Dois.
+ *
+ * <p>Use {@link Doi#builder()} for constructing a new Doi instance.
+ */
+public abstract class Doi {
+
+    public static final String DOI_PROXY = "https://doi.org/";
+    private static final String FORWARD_SLASH = "/";
+
+    public static ImmutableDoi.Builder builder() {
+        return ImmutableDoi.builder();
+    }
+
+    public abstract String getPrefix();
+
+    public abstract String getSuffix();
+
+    public String getProxy() {
+        // default value
+        return DOI_PROXY;
+    }
+
+    /**
+     * Represents the DOI with ${prefix}/${suffix}.
+     *
+     * @return prefix/suffix (DOI identifier)
+     */
+    public String toIdentifier() {
+        return getPrefix() + FORWARD_SLASH + getSuffix();
+    }
+
+    /**
+     * Represents the DOI as an URI, this includes proxy, prefix and suffix.
+     *
+     * @return DOI as URI with proxy, prefix and suffix.
+     */
+    public URI toId() {
+        return URI.create(getProxy() + getPrefix() + FORWARD_SLASH + getSuffix());
+    }
+}
--- a/publication-commons/src/main/java/no/unit/nva/doi/models/Doi.java
+++ b/publication-commons/src/main/java/no/unit/nva/doi/models/Doi.java
@@ -1,6 +1,5 @@
 package no.unit.nva.doi.models;
 
-import java.net.MalformedURLException;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.List;
--- a/publication-commons/src/main/java/no/unit/nva/doi/models/Doi.java
+++ b/publication-commons/src/main/java/no/unit/nva/doi/models/Doi.java
@@ -2,7 +2,6 @@ package no.unit.nva.doi.models;
 
 import java.net.MalformedURLException;
 import java.net.URI;
-import nva.commons.utils.JacocoGenerated;
 
 /**
  * Doi class for working with Dois.
@@ -56,7 +55,6 @@ public abstract class Doi {
      * @return scheme with authority and host and forward slash at the end.
      * @see java.net.URL#toExternalForm
      */
-    @JacocoGenerated
     @SuppressWarnings("PMD.UseStringBufferForStringAppends") // since we copy JDK method!
     private String extractSchemeWithAuthorityAndHost() {
         try {
--- a/publication-commons/src/main/java/no/unit/nva/doi/models/Doi.java
+++ b/publication-commons/src/main/java/no/unit/nva/doi/models/Doi.java
@@ -47,7 +47,7 @@ public abstract class Doi {
      *
      * @return DOI as URI with proxy, prefix and suffix.
      */
-    public URI toId() {
+    public URI toUri() {
         String schemeWithAuthorityAndHost = extractSchemeWithAuthorityAndHost();
         return URI.create(schemeWithAuthorityAndHost + getPrefix() + PATH_SEPARATOR + getSuffix());
     }
--- a/publication-commons/src/main/java/no/unit/nva/doi/models/Doi.java
+++ b/publication-commons/src/main/java/no/unit/nva/doi/models/Doi.java
@@ -71,8 +71,6 @@ public abstract class Doi {
                     + FORWARD_SLASH;
             return schemeWithAuthorityAndHost;
         } catch (MalformedURLException e) {
-            // This should not really ever happen because argument proxy has been validated in
-            // ImmutableDoi#validateProxyUri.
             throw new IllegalStateException(ERROR_PROXY_URI_MUST_BE_A_VALID_URL, e);
         }
     }
--- a/publication-commons/src/main/java/no/unit/nva/doi/models/Doi.java
+++ b/publication-commons/src/main/java/no/unit/nva/doi/models/Doi.java
@@ -13,6 +13,7 @@ public abstract class Doi {
 
     public static final String HANDLE_DOI_PREFIX = "10.";
     public static final String ERROR_PROXY_URI_MUST_BE_A_VALID_URL = "Proxy URI must be a valid URL";
+    public static final String DOI_URI_SYNTAX = "<hypertextTransferProtocol>://<proxy>/<prefix>/<suffix>";
     protected static final String DOI_ORG = "doi.org";
     protected static final String HANDLE_STAGE_DATACITE_ORG = "handle.stage.datacite.org";
     protected static final String DX_DOI_ORG = "dx.doi.org";
--- a/publication-commons/src/main/java/no/unit/nva/doi/models/Doi.java
+++ b/publication-commons/src/main/java/no/unit/nva/doi/models/Doi.java
@@ -17,8 +17,9 @@ public abstract class Doi {
     public static final String HANDLE_STAGE_DATACITE_ORG = "handle.stage.datacite.org";
     public static final String DX_DOI_ORG = "dx.doi.org";
     public static final List<String> VALID_PROXIES = List.of(DOI_ORG, DX_DOI_ORG, HANDLE_STAGE_DATACITE_ORG);
-    protected static final String PATH_SEPARATOR = "/";
-    public static final URI DOI_PROXY = URI.create(HTTPS.concat(DOI_ORG).concat(PATH_SEPARATOR));
+    protected static final char PATH_SEPARATOR = '/';
+    protected static final String PATH_SEPARATOR_STRING = String.valueOf(PATH_SEPARATOR);
+    public static final URI DOI_PROXY = URI.create(HTTPS.concat(DOI_ORG).concat(PATH_SEPARATOR_STRING));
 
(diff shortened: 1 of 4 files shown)


publication-commons/src/main/java/no/unit/nva/doi/models/Doi.java
```diff
@@ -0,0 +1,77 @@
...
+public abstract class Doi {
+
+    public static final URI DOI_PROXY = URI.create("https://doi.org/");
+    public static final String ERROR_PROXY_URI_MUST_BE_A_VALID_URL = "Proxy URI must be a valid URL";
+    private static final String FORWARD_SLASH = "/";
+
+    public static ImmutableDoi.Builder builder() {
+        return ImmutableDoi.builder();
+    }
+
```
norrs, line 21: Do we want copyOf here?

```
public static Doi copyOf(Doi doi) {
  return immutableDoi.copyOf(doi)
}
```

publication-commons/src/main/java/no/unit/nva/doi/models/Doi.java
```diff
@@ -0,0 +1,76 @@
...
+    public String toIdentifier() {
+        return getPrefix() + FORWARD_SLASH + getSuffix();
+    }
+
+    /**
+     * Represents the DOI as an URI, this includes proxy, prefix and suffix.
+     *
+     * @return DOI as URI with proxy, prefix and suffix.
+     */
+    public URI toId() {
```
norrs, line 44: could maybe also add `toDoi()` alias? @brinxmat

janbj (dismissed):

norrs:
Added withDoi(URI doi) to builder and removed some ugly hacks and rather use the long URI constructor.

publication-commons/src/main/java/no/unit/nva/doi/models/ImmutableDoi.java
```diff
@@ -0,0 +1,369 @@
...
+            return doi.getPath().chars().filter(ch -> ch == FORWARD_SLASH).count() != 2;
+        }
+
+        private static void checkNotIsSet(boolean isSet, String name) {
+            if (isSet) {
+                throw new IllegalStateException("Builder of Doi is strict, attribute is already set: ".concat(name));
+            }
+        }
+
+        private boolean proxyIsSet() {
```
brinxmat, line 340: These in essence replace nonNull(field), do they not?

publication-commons/src/main/java/no/unit/nva/doi/models/ImmutableDoi.java
```diff
@@ -0,0 +1,369 @@
...
+    }
+
+    /**
+     * Builds instances of type {@link ImmutableDoi ImmutableDoi}. Initialize attributes and then invoke the {@link
+     * #build()} method to create an immutable instance.
+     *
+     * <p><em>{@code Builder} is not thread-safe and generally should not be stored in a field or collection,
+     * but instead used immediately to create instances.</em>
+     */
+    public static final class Builder {
```
brinxmat, line 221: This has sufficient complexity that I am lead to believe that there are simpler ways of achieving the same; that said, I am unsure what you are trying to achieve from the code.

publication-commons/src/test/java/no/unit/nva/doi/models/ImmutableDoiTest.java
```diff
@@ -0,0 +1,294 @@
...
+    public static final String URI_VALID_EMAILTO_BUT_INVALID_URL = "emailto:nope@example.net";
+    private static final URI STAGE_DOI_PROXY = URI.create("https://handle.stage.datacite.org/");
+    private static final String DEMO_PREFIX = "10.5072";
+    public static final String EXAMPLE_PREFIX = DEMO_PREFIX;
+    public static final String EXAMPLE_IDENTIFIER = EXAMPLE_PREFIX + FORWARD_SLASH + EXAMPLE_SUFFIX;
+    private static final URI EXAMPLE_PROXY = STAGE_DOI_PROXY;
+    public static final URI EXAMPLE_DOI = URI.create(EXAMPLE_PROXY + EXAMPLE_IDENTIFIER);
+    private static final String EXAMPLE_PREFIX_2 = "10.16903";
+
+    @Test
```
brinxmat, line 32: Forventer at blant annet, følgende ikke valideres som DOI:

`            "https://example.org/local.irrigation/burlapSack",
            "https://example.org/wanderlust/freakout",
            "https://doi.org/11.9821/corgismack",
            "ftp://doi.org/murkyWater/fretwork"`

Og at withIdentifier ikke aksepterer disse som verdier.

brinxmat (changes requested):

publication-commons/src/main/java/no/unit/nva/doi/models/ImmutableDoi.java
norrs: So the base is based on org.immutables which we have inhoused.
I trust the checks are faster from the generated code I used as a base.

publication-commons/src/test/java/no/unit/nva/doi/models/ImmutableDoiTest.java
```diff
@@ -0,0 +1,294 @@
...
+    private static class AnotherPojoDoi extends Doi {
+
+        private final URI proxy;
+        private final String prefix;
+        private final String suffix;
+
+        public AnotherPojoDoi(URI proxy, String prefix, String suffix) {
+            this.proxy = proxy;
+            this.prefix = prefix;
+            this.suffix = prefix;
```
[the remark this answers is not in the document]
norrs: Nice catch, fixed https://github.com/BIBSYSDEV/nva-publication-api/pull/96/commits/4c3d6cc7afeefe3da9caf5213260a3b500209bc8

publication-commons/src/main/java/no/unit/nva/doi/models/Doi.java
```diff
@@ -0,0 +1,76 @@
+package no.unit.nva.doi.models;
```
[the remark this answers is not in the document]
norrs: What do you mean about commons here? 

Is this not a data class, hence package models that is a convention we use for POJOs in other places in our project? 
Or what is `models` used for then?

publication-commons/src/test/java/no/unit/nva/doi/models/ImmutableDoiTest.java
norrs: https://github.com/BIBSYSDEV/nva-publication-api/pull/96/commits/2fe80f388dc107333cfa18ae88b4273e0c03c374 and https://github.com/BIBSYSDEV/nva-publication-api/pull/96/commits/3e033e023c96cf3329d2cfe4ccb8c2ac758e73d8 covers it?

publication-commons/src/main/java/no/unit/nva/doi/models/Doi.java
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * Doi class for working with Dois.
+ *
+ * <p>Use {@link Doi#builder()} for constructing a new Doi instance.
+ */
+public abstract class Doi {
+
+    public static final URI DOI_PROXY = URI.create("https://doi.org/");
+    public static final String ERROR_PROXY_URI_MUST_BE_A_VALID_URL = "Proxy URI must be a valid URL";
+    private static final String FORWARD_SLASH = "/";
```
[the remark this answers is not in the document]
norrs: Fixed in https://github.com/BIBSYSDEV/nva-publication-api/pull/96/commits/982d6d95a58a6b512c38179045007f7c0448eae8

publication-commons/src/main/java/no/unit/nva/doi/models/Doi.java
norrs: Same as I added `Doi#builder` , easier and looks nicer. Then you end up using:

```
Doi.copyOf(..)
or 
Doi.builder()

instead of having to type the ImmutableDoi name, like ImmutableDoi.builder()
``

Comment (norrs) [reply publication-commons/src/main/java/no/unit/nva/doi/models/Doi.java:32]: It hides the `implementation` behind the Doi class.

File: publication-commons/src/test/java/no/unit/nva/doi/models/ImmutableDoiTest.java (hunk 5 above)
Comment (norrs) [reply publication-commons/src/test/java/no/unit/nva/doi/models/ImmutableDoiTest.java:42]: Together with cfbf0f01cdabe7aa0d1f3064a9c6ff475949aa8a I believe.

File: publication-commons/src/main/java/no/unit/nva/doi/models/Doi.java (hunk 7 above)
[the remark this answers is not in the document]
Comment (norrs) [reply publication-commons/src/main/java/no/unit/nva/doi/models/Doi.java:1]: Note that we use `models` quite many other places.. so I find it weird to put it under `commons`.

File: publication-commons/src/main/java/no/unit/nva/doi/models/Doi.java (hunk 2 above)
Comment (norrs) [reply publication-commons/src/main/java/no/unit/nva/doi/models/Doi.java]: Fixed.

File: publication-commons/src/main/java/no/unit/nva/doi/models/Doi.java [hunk 9]
```diff
@@ -0,0 +1,76 @@
...
+import java.net.URI;
+
+/**
+ * Doi class for working with Dois.
+ *
+ * <p>Use {@link Doi#builder()} for constructing a new Doi instance.
+ */
+public abstract class Doi {
+
+    public static final URI DOI_PROXY = URI.create("https://doi.org/");
```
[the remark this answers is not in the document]
norrs: Refactored a tiny bit, it now also show that this is the default proxy if you don't specify one.

publication-commons/src/test/java/no/unit/nva/doi/models/ImmutableDoiTest.java
norrs: Also added https://github.com/BIBSYSDEV/nva-publication-api/pull/96/commits/004083546b6399a149f6429c272d0d6b2cef39e4 to test for all known valid proxies and schemes.

publication-commons/src/main/java/no/unit/nva/doi/models/ImmutableDoi.java
```diff
@@ -0,0 +1,396 @@
...
+
+    /**
+     * Copy the current immutable object by setting a value for the {@link Doi#getProxy() proxy} attribute. An equals
+     * check used to prevent copying of the same value by returning {@code this}.
+     *
+     * @param value A new value for proxy
+     * @return A modified copy of the {@code this} object
+     */
+    public final ImmutableDoi withProxy(URI value) {
+        URI newValue = Objects.requireNonNull(value, MESSAGE_NON_NULL_ARGUMENT_FOR_PARAMETER_PROXY);
```
axthosarouris, line 123: These methods are problematic because they replicate the logic in the builder. It could be much simpler if the builder allowed overwriting values, copy would work by creating a builder and validation happened in build. 
```
Doi existingDoi;
doi.copy().withSuffix(newSuffix).build().
```
Also for the time being I do not see the need to have two classes for Doi. One class where one creates a Doi and cannot mutate the fields (i.e., no setters) would be OK.

publication-commons/src/main/java/no/unit/nva/doi/models/ImmutableDoi.java
```diff
@@ -0,0 +1,396 @@
...
+    }
+
+    /**
+     * Copy the current immutable object by setting a value for the {@link Doi#getPrefix()} () prefix} attribute. An
+     * equals check used to prevent copying of the same value by returning {@code this}.
+     *
+     * @param value A new value for prefix
+     * @return A modified copy of the {@code this} object
+     */
+    public final ImmutableDoi withPrefix(String value) {
```
axthosarouris, line 138: ditto.

publication-commons/src/main/java/no/unit/nva/doi/models/ImmutableDoi.java
```diff
@@ -0,0 +1,396 @@
...
+
+    /**
+     * Copy the current immutable object by setting a value for the {@link Doi#getSuffix()} () suffix} attribute. An
+     * equals check used to prevent copying of the same value by returning {@code this}.
+     *
+     * @param value A new value for suffix
+     * @return A modified copy of the {@code this} object
+     */
+    public final ImmutableDoi withSuffix(String value) {
+        String newValue = Objects.requireNonNull(value, MESSAGE_NON_NULL_ARGUMENT_FOR_PARAMETER_SUFFIX);
```
axthosarouris, line 154: ditto

publication-commons/src/main/java/no/unit/nva/doi/models/ImmutableDoi.java
```diff
@@ -0,0 +1,396 @@
...
+         * Doi#getProxy() proxy}.</em>
+         *
+         * @param proxy The value for proxy
+         * @return {@code this} builder for use in a chained invocation
+         */
+
+        public final Builder withProxy(URI proxy) {
+            checkNotIsSet(proxyIsSet(), "proxy");
+            var newProxyValue = createUriWithoutPathQueryAndFragmentWithSuffixForwardSlash(proxy);
+            validateProxyUri(newProxyValue);
```
axthosarouris, line 254: validation should be in one method called `validate()` that would be called in the constructor or build()

publication-commons/src/main/java/no/unit/nva/doi/models/ImmutableDoi.java
axthosarouris: Why is it necessary to check in the builder if something has been set already? Why to make the builder "immutable" also?

publication-commons/src/test/java/no/unit/nva/doi/models/ImmutableDoiTest.java
```diff
@@ -0,0 +1,372 @@
...
+    void equalsReturnsTrueForTwoDifferentInstancesButIdenticalObjects() {
+        var doi = createDoi();
+        var doi2 = createDoi();
+
+        assertThat(doi, is(equalTo(doi2)));
+        assertThat(doi, is(equalTo(doi)));
+    }
+
+    @Test
+    void toIdThenReturnsDoiIdWhichIsProxyPrefixAndSuffix() {
```
axthosarouris, line 126: here you call toUri

publication-commons/src/test/java/no/unit/nva/doi/models/ImmutableDoiTest.java
```diff
@@ -0,0 +1,372 @@
...
+        String randomSuffix = createRandomSuffix();
+        var doi = createDoi(randomSuffix);
+        URI expectedUri = URI.create(EXAMPLE_PROXY + EXAMPLE_PREFIX + FORWARD_SLASH + randomSuffix);
+        assertThat(doi.toUri(), is(equalTo(expectedUri)));
+    }
+
+    @Test
+    void toIdWhereBuilderWithProxyWithoutSuffixSlashReturnsCorrectIdURI() {
```
axthosarouris, line 134: ditto

publication-commons/src/test/java/no/unit/nva/doi/models/ImmutableDoiTest.java
```diff
@@ -0,0 +1,372 @@
...
+
+    @Test
+    void toIdWithAnotherSubClassOfDoiWithInvalidProxyUriThenThrowsIllegalStateException() {
+        var doi = getAnotherPojoDoi(URI.create(URI_VALID_EMAILTO_BUT_INVALID_URL));
+        var actualException = assertThrows(IllegalStateException.class, doi::toUri);
+        assertThat(actualException.getMessage(), is(equalTo(Doi.ERROR_PROXY_URI_MUST_BE_A_VALID_URL)));
+    }
+
+    @Test
+    void withProxyWithInvalidUrlAsUriThrowsIllegalStateException() {
```
axthosarouris, line 160: remove these methods and use the pattern. I do not think the performance improvements on copying dois are critical. Besides, do we have such an example where we create a DOI from another DOI? 
```
doi.copy().withBlabla().build()
public Builder copy(){

}

**Review by axthosarouris — CHANGES_REQUESTED** [review]

File: publication-commons/src/main/java/no/unit/nva/doi/models/Doi.java (hunk 1 above)
Comment (norrs) [reply publication-commons/src/main/java/no/unit/nva/doi/models/Doi.java:33]: We removed `copyOf` as we dont need that functionality.

File: publication-commons/src/main/java/no/unit/nva/doi/models/ImmutableDoi.java (hunk 3 above)
Comment (norrs) [reply publication-commons/src/main/java/no/unit/nva/doi/models/ImmutableDoi.java:302]: For avoid calling same methods twice and validating proxy as proxy argument is a URI when we only use scheme, authority and port. Also note proxy has a default argument if not provided and then is not set in the builder.

File: publication-commons/src/main/java/no/unit/nva/doi/models/ImmutableDoi.java [hunk 17]
```diff
@@ -0,0 +1,396 @@
+package no.unit.nva.doi.models;
```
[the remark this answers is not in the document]
norrs: I have written this by hand, it is different then the auto generated code as it does not contain the validation.

publication-commons/src/test/java/no/unit/nva/doi/models/ImmutableDoiTest.java
norrs: Removed.

publication-commons/src/test/java/no/unit/nva/doi/models/ImmutableDoiTest.java
norrs: Updated test names.

publication-commons/src/test/java/no/unit/nva/doi/models/ImmutableDoiTest.java
norrs: Updated test names.

publication-commons/src/main/java/no/unit/nva/doi/models/ImmutableDoi.java
norrs: Fixed and moved into build() on builder.

publication-commons/src/main/java/no/unit/nva/doi/models/ImmutableDoi.java
norrs: Removed `copyOf` functionality

publication-commons/src/main/java/no/unit/nva/doi/models/ImmutableDoi.java
norrs: Removed `copyOf` functionality

publication-commons/src/main/java/no/unit/nva/doi/models/ImmutableDoi.java
norrs: Removed `copyOf` functionality

norrs:
@axthosarouris @brinxmat PTAL

axthosarouris (changes requested):

axthosarouris:
General comment. In this class, we build a DOI object using a host, a prefix and a suffix. My understanding is that we always handle existing DOIs as a type of identity that we never construct ourselves. I have not worked with Datacite.  When we receive a Doi from them, do we receive a host a prefix and a suffix or do we receive a URL? 

To my understanding if I had a URL that it was a DOI and I wanted to create a Doi object, I would have to extract the host prefix and suffix, from the URL to feed it to the Doi class in order for the Doi class to build the URL again. Why not use the URL directly?

norrs:
Their APIs are working with identifiers, hence a need for the type. URI class should always be used instead of URL.

norrs:
This is good enough for me for the initial start of having a Doi data type. Leaving the readers the options to create better versions later.

publication-commons/src/main/java/no/unit/nva/doi/models/Doi.java
```diff
@@ -0,0 +1,75 @@
...
+
+    public static final String HANDLE_DOI_PREFIX = "10.";
+    public static final String ERROR_PROXY_URI_MUST_BE_A_VALID_URL = "Proxy URI must be a valid URL";
+    protected static final String DOI_ORG = "doi.org";
+    protected static final String HANDLE_STAGE_DATACITE_ORG = "handle.stage.datacite.org";
+    protected static final String DX_DOI_ORG = "dx.doi.org";
+    public static final List<String> VALID_PROXIES = List.of(DOI_ORG, DX_DOI_ORG, HANDLE_STAGE_DATACITE_ORG);
+    protected static final String HTTPS = "https";
+    protected static final String HTTP = "http";
+    public static final List<String> VALID_SCHEMES = List.of(HTTPS, HTTP);
```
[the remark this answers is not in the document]
janbj: What use case do we have that requires a DOI to be anything else than a URL type object that can be validated?

axthosarouris (approved):

brinxmat (approved):

publication-commons/src/main/java/no/unit/nva/doi/models/Doi.java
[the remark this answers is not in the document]
norrs: @janbj : FYI, https://github.com/BIBSYSDEV/nva-publication-api/pull/96#issuecomment-730971822 

URL#equals etc does DNS lookups etc which is not nice. Its a long term common thing to avoid URL ( https://bugs.java.com/bugdatabase/view_bug.do?bug_id=4434494 ) . It has sadly other annoying things as well when it comes to percent encoding etc.

publication-commons/src/main/java/no/unit/nva/doi/models/ImmutableDoi.java
norrs: We replaced the bitwise optimization and does null checking instead. 

To move on and getting delivery of user story: https://github.com/BIBSYSDEV/nva-publication-api/pull/96#issuecomment-731022543

publication-commons/src/test/java/no/unit/nva/doi/models/ImmutableDoiTest.java
norrs: Lets add more validation if needed in the `Builder#build` if we notice we need more validation: https://github.com/BIBSYSDEV/nva-publication-api/pull/96#issuecomment-731022543

PR #98, opened 2020-11-17
Added PublicationDtoTestDataGenerator

janbj:


axthosarouris (approved):

axthosarouris (approved):

norrs (changes requested):
@janbj The test generator need to return `this` after creation a publication, then you want a method for `as(Doi)Publication(Dto?)` to return the DTOs that were created.

Good idea to also provide a way to clear recorded publications in the test generator.

This so you can access the generated objects that have been created via the test generator.

axthosarouris (approved):

norrs (approved):

norrs:
Acknowledged a bit too fast, I notice this is not what we want. 

The idea with the test generator is we want a recorder of simply adding good and bad objects related to the test that we can retrieve again to do assertions against. 

Idea is that the test generator should do the "complicated" of random generate both valid type of different objects and invalid different type of objects, including test data with different locale etc.

PR #102, opened 2020-11-19
Add CFN linter

norrs:
Use a cloudformation linter. 

It deletes the line where we use the `envsubst` for settings some variables for avoiding linting errors on that particular line.

### DIFF
--- a/template.yaml
+++ b/template.yaml
@@ -41,10 +41,6 @@ Parameters:
     Type: String
     Description: Reference to index containing publications by owner
     Default: ByPublisher
-  PublicationsDoiRequestsByStatusIndexName:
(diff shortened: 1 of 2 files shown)


template.yaml
```diff
@@ -41,10 +41,6 @@ Parameters:
     Type: String
     Description: Reference to index containing publications by owner
     Default: ByPublisher
-  PublicationsDoiRequestsByStatusIndexName:
```
janbj, line 44: Removed on purpose?

norrs: Linter says it is not used by anything, I could not see it in use either.

janbj (approved):

PR #103, opened 2020-11-19
Obligatory fields

axthosarouris:
The scope of this PR is to make the doi-request chain work, and provide some information about the missing fields when a failure occurs with as little changes as possible.

Description:

1. Check the incoming events if they are valid and provide a message of the missing mandatory fields.
2. If there is no DoiRequest ignore the Publication update. (This needs fixing because there might be publications with pre-existing DOI, but belongs to another PR.
3) Update tests to check for the cases of missing fields.


publication-testing/src/main/java/no/unit/nva/publication/doi/dto/PublicationStreamRecordTestDataGenerator.java
```diff
@@ -111,20 +110,6 @@ public DynamodbStreamRecord asDynamoDbStreamRecord() {
         return toDynamodbStreamRecord(event);
     }
 
-    private void updateDoiRequest(DoiRequest doiRequest, ObjectNode event) {
```
axthosarouris, line 114: moved lower due to autoformating

publication-doi-commons/src/main/java/no/unit/nva/publication/doi/PublicationMapper.java
```diff
@@ -91,23 +95,70 @@ public PublicationMapping fromDynamodbStreamRecord(DynamodbStreamRecord streamRe
      * @return Publication doi.Publication
      */
     public Publication fromDynamodbStreamRecordDao(DynamodbStreamRecordImageDao dao) {
+
         return Builder.newBuilder()
-            .withId(transformIdentifierToId(namespacePublication, dao.getIdentifier()))
-            .withInstitutionOwner(URI.create(dao.getPublisherId()))
+            .withId(transformIdentifierToId(namespacePublication, dao))
```
axthosarouris, line 100: read the JsonPointers and if the field exists add the value, else put null. Validation will not happen here.

publication-doi-commons/src/main/java/no/unit/nva/publication/doi/dto/DoiRequest.java
```diff
@@ -4,21 +4,27 @@
 import java.time.Instant;
 import java.util.Objects;
 import nva.commons.utils.JacocoGenerated;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
-public class DoiRequest {
+public class DoiRequest extends Validatable {
```
axthosarouris, line 10: `Validatable` allows us to check for mandatory fields.

publication-doi-commons/src/main/java/no/unit/nva/publication/doi/dto/DoiRequest.java
```diff
@@ -51,6 +57,12 @@ public int hashCode() {
         return Objects.hash(getStatus(), getModifiedDate());
     }
 
+    @Override
+    @JacocoGenerated
+    protected Logger logger() {
```
axthosarouris, line 62: required by `Validatable`

config/pmd/ruleset.xml
```diff
@@ -21,6 +21,8 @@
     <exclude name="SystemPrintln"/>
     <exclude name="GuardLogStatement"/>
     <exclude name="AvoidPrintStackTrace"/>
+    <!--Allow arrays as parameters. -->
+    <exclude name="UseVarargs"/>
```
axthosarouris, line 25: We use arrays in `Validatable` class.

publication-doi-commons/src/main/java/no/unit/nva/publication/doi/dto/PublicationDate.java
```diff
@@ -9,13 +9,18 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
-public class PublicationDate {
+public class PublicationDate extends Validatable {
 
-    public static final JsonPointer YEAR_JSON_POINTER = JsonPointer.compile("/date/m/year/s");
-    public static final JsonPointer MONTH_JSON_POINTER = JsonPointer.compile("/date/m/month/s");
-    public static final JsonPointer DAY_JSON_POINTER = JsonPointer.compile("/date/m/day/s");
+    public static final JsonPointer YEAR_JSON_POINTER = JsonPointer.compile("/year/s");
```
axthosarouris, line 17: Update pointers to look into the PublicationDate object.

publication-doi-commons/src/main/java/no/unit/nva/publication/doi/dynamodb/dao/DynamodbStreamRecordImageDao.java
```diff
@@ -122,7 +122,7 @@ public Builder withDynamodbStreamRecordImage(JsonNode rootNode) {
             publicationInstanceType = textFromNode(
                 rootNode,
                 jsonPointers.getEntityDescriptionReferenceTypeJsonPointer());
-            publicationReleaseDate = rootNode.at(jsonPointers.getEntityDescriptionMapJsonPointer());
+            publicationReleaseDate = rootNode.at(jsonPointers.getEntityDescriptionDateJsonPointer());
```
axthosarouris, line 125: PublicationReleaseDatePointer was pointing to EntityDescription and could not parse the date.

publication-doi-commons/src/main/java/no/unit/nva/publication/doi/dynamodb/dao/DynamodbStreamRecordJsonPointers.java
```diff
@@ -37,15 +38,15 @@ public String getValue() {
 
     private static final String ENTITY_DESCRIPTION_REFERENCE_TYPE =
         "/entityDescription/m/reference/m/publicationInstance/m/type/s";
-    private static final String ENTITY_DESCRIPTION_MAP = "/entityDescription/m";
+
```
axthosarouris, line 41: entity description pointer was used only for the publication date and it was replaced.

publish-publication/src/test/java/no/unit/nva/publication/publish/PublishPublicationHandlerTest.java
```diff
@@ -81,6 +80,14 @@ public void canPublishPublication() throws Exception {
         assertEquals(expected, actual);
     }
 
+    @Test
```
axthosarouris, line 83: test moved due to auto-formating

publication-doi-commons/src/main/java/no/unit/nva/publication/doi/dto/DoiRequest.java
```diff
@@ -4,21 +4,27 @@
...
      *
-     * @param status        doi request status
-     * @param modifiedDate  modified date of doi request
+     * @param status       doi request status
+     * @param modifiedDate modified date of doi request
      */
     public DoiRequest(
         @JsonProperty("status") DoiRequestStatus status,
         @JsonProperty("modifiedDate") Instant modifiedDate) {
+        super();
```
axthosarouris, line 27: PMD

publication-doi-commons/src/main/java/no/unit/nva/publication/doi/dto/Publication.java
```diff
@@ -48,6 +59,7 @@ public Publication(@JsonProperty("id") URI id,
                        @JsonProperty("status") PublicationStatus status,
                        @JsonProperty("contributors") List<Contributor> contributors,
                        @JsonProperty("publication_date") PublicationDate publicationDate) {
+        super();
```
axthosarouris, line 62: PMD

publication-doi-commons/src/test/java/no/unit/nva/publication/doi/dto/ValidatableTest.java
```diff
@@ -0,0 +1,185 @@
...
+import static org.hamcrest.core.StringContains.containsString;
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import nva.commons.utils.JacocoGenerated;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.function.Executable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+class ValidatableTest {
```
norrs, line 13: ditto above about Validatable.

publication-doi-commons/src/main/java/no/unit/nva/publication/doi/dto/Validatable.java
```diff
@@ -0,0 +1,88 @@
...
+
+import static nva.commons.utils.attempt.Try.attempt;
+import java.lang.reflect.Field;
+import java.util.Arrays;
+import java.util.Objects;
+import nva.commons.utils.attempt.Failure;
+import nva.commons.utils.attempt.Try;
+import org.slf4j.Logger;
+
+public abstract class Validatable {
```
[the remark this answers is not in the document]
norrs: I find is very odd we should create our own code for supporting `@MandatoryField` annotation, when we have native language constructs that we could use to our benefit instead. 
(optionals)

norrs (changes requested):
I stopped reading because I think the `Validatable` class is wrong approach, doesn't feel right.

norrs:
Two things summarized as I understood it from the 15 minutes meeting today:

1) Can we leverage existing libraries for bean validation? JSR380
1.1) Are we okey with introducing more reflection usage which is problematic for when building native binaries which we want to look into?
1.2) If we want JSR380, are we ready to maybe look into quarkus etc instead of having to solve the problems others have solved for us?
2) Could org.immutables be used here for DTOs? If fields are present, they should be set. Only optional if it has a optional getter.  This also works fine with native binaries. DAOs and DTOs should be simple data objects.

publication-doi-commons/src/main/java/no/unit/nva/publication/doi/dynamodb/dao/DynamodbStreamRecordJsonPointers.java
axthosarouris: Not at the moment. The Json pointers are basically the physical model hard-coded and we should move away from them because any updating of the physical model (storage model) would mean that we have to go into the code and change the pointers manually. This PR is a fix making the Doi-request chain work.

publication-doi-commons/src/test/java/no/unit/nva/publication/doi/PublicationMapperTest.java
```diff
@@ -52,21 +53,21 @@ void fromDynamoStreamRecordNewImage() {
...
     @Test
-    void fromDynamoStreamRecordOldImage() {
-        DynamodbStreamRecordJsonPointers jsonPointers = new DynamodbStreamRecordJsonPointers(
-            DynamodbImageType.OLD);
-        Builder daoBuilder = Builder.createValidPublication(FAKER, jsonPointers, StreamViewType.OLD_IMAGE.toString());
-        var dynamodbStreamRecord = daoBuilder.build().asDynamoDbStreamRecord();
+    void fromDynamoStreamRecordOldImage() throws JsonProcessingException {
+        String dynamoStreamRecordString = IoUtils.stringFromResources(
+            Path.of(DYNAMO_STREAM_EVENT_WITH_OLD_IMAGE_JSON));
+        DynamodbEvent.DynamodbStreamRecord recordWithOldImage = objectMapper.readValue(dynamoStreamRecordString,
```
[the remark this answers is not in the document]
axthosarouris: The test-generator produces now a lot of invalid objects (after the insertion of validation) and I want to limit the changes to the absolute min. This test has no clear expectations and  is only testing that an old image is present. There is no real need to read from the test-generator.

publication-doi-commons/src/main/java/no/unit/nva/publication/doi/dto/Validatable.java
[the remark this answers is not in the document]
axthosarouris: Right now these classes follow not such pattern. I want to make the chain to work with as few changes in the code as possible. Right now the test generators create invalid objects during testing and I want to avoid refactoring them too much

publication-doi-commons/src/test/java/no/unit/nva/publication/doi/PublicationMapperTest.java
```diff
@@ -84,36 +85,44 @@ void fromDynamodbStreamRecordDaoThenReturnFullyMappedPublicationDto() {
...
         var publication = new PublicationMapper(EXAMPLE_NAMESPACE).fromDynamodbStreamRecordDao(
             dao);
 
         assertThat(publication.getContributor(), hasSize(0));
     }
 
-    private PublicationStreamRecordTestDataGenerator createDynamoDbStreamRecordWithoutContributorIdentityNames() {
-        return PublicationStreamRecordTestDataGenerator.Builder.createValidPublication(FAKER, jsonPointers)
+    private Builder validPublication() {
+        return Builder.createValidPublication(FAKER, new DynamodbStreamRecordJsonPointers(DynamodbImageType.NEW));
```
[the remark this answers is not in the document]
axthosarouris: No method is using any other kind of pointers.

axthosarouris:
> Two things summarized as I understood it from the 15 minutes meeting today:
> 
>     1. Can we leverage existing libraries for bean validation? JSR380
>        1.1) Are we okey with introducing more reflection usage which is problematic for when building native binaries which we want to look into?
>        1.2) If we want JSR380, are we ready to maybe look into quarkus etc instead of having to solve the problems others have solved for us?
> 
>     2. Could org.immutables be used here for DTOs? If fields are present, they should be set. Only optional if it has a optional getter.  This also works fine with native binaries. DAOs and DTOs should be simple data objects.

Immutables is a possible way to go, but I would like to look more into it, because I have the impression that we should include the generated classes in the classpath.

PR #104, opened 2020-11-22
NP-1828: add checks for mandatory fields

axthosarouris:
This is a re-write of the PR #103 which is now closed. These are the absolutely minimum changes for making the  `DynamoDbFanoutPublicationDtoProducer` work.  About half of the changed files are just json test files. 

I chose to not use the  JSR380 validation because it does not cover the case where a field is present but invalid. For example a Publication with a non-null but invalid DoiRequest will be invalid. In any case, should we choose to use this library the refactoring would be easy with no changes in tests because all new tests have been written on handler level. 

I think the idea of using Immutables should be tested but it is out of scope of this PR which is to make the doi-request-chain work with minimal changes. My main concern right now with using Immutables in this PR, is the dependency to the generated classes and how this will affect the other modules. However, given that the new tests are written on a handler  level, the introduction of Immutables will be easy if we choose to go that way.

### DIFF
--- a/publication-doi-commons/src/main/java/no/unit/nva/publication/doi/dto/DoiRequest.java
+++ b/publication-doi-commons/src/main/java/no/unit/nva/publication/doi/dto/DoiRequest.java
@@ -24,6 +24,7 @@ public class DoiRequest extends Validatable {
         this.modifiedDate = modifiedDate;
     }
 
+    @Override
     public void validate() {
         requireFieldIsNotNull(status, "DoiRequest.status");
         requireFieldIsNotNull(modifiedDate, "DoiRequest.modifiedDate");
--- a/publication-doi-commons/src/main/java/no/unit/nva/publication/doi/dto/DoiRequest.java
+++ b/publication-doi-commons/src/main/java/no/unit/nva/publication/doi/dto/DoiRequest.java
@@ -7,6 +7,8 @@ import nva.commons.utils.JacocoGenerated;
 
 public class DoiRequest extends Validatable {
 
+    public static final String DOI_REQUEST_STATUS_FIELD_INFO = "DoiRequest.status";
+    public static final String DOI_REQUEST_MODIFIED_DATE_FIELD_INFO = "DoiRequest.modifiedDate";
     private final DoiRequestStatus status;
     private final Instant modifiedDate;
 
@@ -26,8 +28,8 @@ public class DoiRequest extends Validatable {
 
     @Override
     public void validate() {
-        requireFieldIsNotNull(status, "DoiRequest.status");
-        requireFieldIsNotNull(modifiedDate, "DoiRequest.modifiedDate");
+        requireFieldIsNotNull(status, DOI_REQUEST_STATUS_FIELD_INFO);
+        requireFieldIsNotNull(modifiedDate, DOI_REQUEST_MODIFIED_DATE_FIELD_INFO);
     }
 
     public DoiRequestStatus getStatus() {
--- a/publication-doi-commons/src/main/java/no/unit/nva/publication/doi/dto/DoiRequest.java
+++ b/publication-doi-commons/src/main/java/no/unit/nva/publication/doi/dto/DoiRequest.java
@@ -5,7 +5,7 @@ import java.time.Instant;
 import java.util.Objects;
 import nva.commons.utils.JacocoGenerated;
 
-public class DoiRequest {
+public class DoiRequest extends Validatable {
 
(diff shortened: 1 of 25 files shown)


publication-doi-commons/src/test/java/no/unit/nva/publication/doi/PublicationMapperTest.java
```diff
@@ -52,13 +55,13 @@ void fromDynamoStreamRecordNewImage() {
     }
 
     @Test
-    void fromDynamoStreamRecordOldImage() {
-        DynamodbStreamRecordJsonPointers jsonPointers = new DynamodbStreamRecordJsonPointers(
-            DynamodbImageType.OLD);
-        Builder daoBuilder = Builder.createValidPublication(FAKER, jsonPointers, StreamViewType.OLD_IMAGE.toString());
-        var dynamodbStreamRecord = daoBuilder.build().asDynamoDbStreamRecord();
+    void fromDynamoStreamRecordOldImage() throws JsonProcessingException {
```
axthosarouris, line 58: Failing test due to bug in TestData generator.

publication-doi-commons/src/test/java/no/unit/nva/publication/doi/dto/PublicationDateTest.java
```diff
@@ -18,27 +18,12 @@
...
     @Test
     void testJsonNodeConstructor() {
-        var actual = new PublicationDate(getPublicationWithDate());
+        var actual = PublicationDate.fromJsonNode(getPublicationWithDate());
         assertThat(actual.getYear(), is(equalTo("1999")));
         assertThat(actual.getMonth(), is(equalTo("07")));
         assertThat(actual.getDay(), is(equalTo("09")));
     }
 
-    @Test
```
axthosarouris, line 27: Removed test with no clear expectations. The is test is also irrelevant now as PublicationDate cannot be created unless there is a year.

publication-event-dtopublicationdoi-producer/build.gradle
```diff
@@ -8,12 +8,12 @@ dependencies {
     }
 
     implementation group: 'com.github.BIBSYSDEV', name: 'nva-commons', version: project.ext.nvaCommonsVersion
-
     implementation group: 'com.amazonaws', name: 'aws-lambda-java-core', version: '1.2.1'
     implementation group: 'com.amazonaws', name: 'aws-lambda-java-events', version: '3.2.0'
-    implementation group: 'software.amazon.awssdk', name: 'eventbridge', version: project.ext.awsSdk2Version
-    implementation group: 'software.amazon.awssdk', name: 'url-connection-client', version: project.ext.awsSdk2Version
-    implementation group: 'software.amazon.awssdk', name: 'sqs', version: project.ext.awsSdk2Version
+
```
axthosarouris, line 13: Removed unused dependencies and added necessary dependency for PMD

publication-doi-commons/src/main/java/no/unit/nva/publication/doi/dto/DoiRequest.java
```diff
@@ -5,24 +5,31 @@
 import java.util.Objects;
 import nva.commons.utils.JacocoGenerated;
 
-public class DoiRequest {
+public class DoiRequest extends Validatable {
```
axthosarouris, line 8: Validatable is now a much simple class with only one very simple method.

brinxmat (approved):

publication-doi-commons/src/test/java/no/unit/nva/publication/doi/PublicationMapperTest.java
norrs: From what I can see the reason why it was failing it due to passing in wrong instance of jsonPointer where it is looking for newImage and not oldImage as the stream view type is set to.

publication-doi-commons/src/main/java/no/unit/nva/publication/doi/dto/Validatable.java
```diff
@@ -0,0 +1,23 @@
+package no.unit.nva.publication.doi.dto;
+
+import static java.util.Objects.isNull;
+
+public abstract class Validatable {
```
norrs, line 5: Probably want to use composeability for this instead of inheritance, but good enough for now.

norrs (commented):
Ideally would be nice if someone else gives a test of how you want a test generator, instead of us having to maintain such long json documents as test resources.

PR #106, opened 2020-11-23
Do not merge: Example of immutables (Sharing knowledge only)

norrs:
@axthosarouris if you wanted to see how it looked like. 

Be free to close it, did it for my own exercise, shared here for learning sharing.

### DIFF
--- a/publication-doi-commons/src/main/java/no/unit/nva/immutables/Builded.java
+++ b/publication-doi-commons/src/main/java/no/unit/nva/immutables/Builded.java
@@ -1,0 +1,16 @@
+package no.unit.nva.immutables;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Target;
+import org.immutables.value.Value;
+
+/**
+ * Builded annotation will generate builder which produces private implementations of abstract value type.
+ */
+@Target(ElementType.TYPE)
+@Value.Style(
+    typeBuilder = "BuilderFor_*",
+    defaultAsDefault = true, // java 8 default methods will be automatically turned into @Value.Default
+    visibility = Value.Style.ImplementationVisibility.PRIVATE,
+    builderVisibility = Value.Style.BuilderVisibility.PACKAGE) // We will extend builder to make it public
+public @interface Builded {}
--- a/publication-doi-commons/src/main/java/no/unit/nva/immutables/Tupled.java
+++ b/publication-doi-commons/src/main/java/no/unit/nva/immutables/Tupled.java
@@ -1,0 +1,16 @@
+package no.unit.nva.immutables;
+
+import org.immutables.value.Value;
+
+/**
+ * Tupled annotation will be used to generate simple tuples in reverse-style, having construction methods of all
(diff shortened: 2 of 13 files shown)


publication-doi-commons/src/main/java/no/unit/nva/immutables/Builded.java
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * Builded annotation will generate builder which produces private implementations of abstract value type.
+ */
+@Target(ElementType.TYPE)
+@Value.Style(
+    typeBuilder = "BuilderFor_*",
+    defaultAsDefault = true, // java 8 default methods will be automatically turned into @Value.Default
+    visibility = Value.Style.ImplementationVisibility.PRIVATE,
+    builderVisibility = Value.Style.BuilderVisibility.PACKAGE) // We will extend builder to make it public
+public @interface Builded {}
```
norrs, line 16:
```
// Custom annotation for builder with private immutable implementation.
@Builded public interface Record {
  long id();
  String name();

  default String notes() { // Works as default attribute!
    return "";
  }

  // Then we extend package-private builder with public nested builder and expose all
  // public methods as methods of Record.Builder.
  class Builder extends BuilderFor_Record {}
}
// ...
Record record = new Record.Builder()
    .id(123L)
    .name("Named Record")
    .notes("Oh, nothing interesting, surely!")
    .build();
```

publication-doi-commons/src/main/java/no/unit/nva/immutables/Tupled.java
```diff
@@ -0,0 +1,16 @@
...
+ * annotations.
+ */
+@Value.Style(// Tupled annotation will serve as both immutable and meta-annotated style annotation
+    typeAbstract = "*Def",
+    typeImmutable = "*",
+    allParameters = true, // all attributes will become constructor parameters
+    // as if they are annotated with @Value.Parameter
+    visibility = Value.Style.ImplementationVisibility.PUBLIC, // Generated class will be always public
+    defaults = @Value.Immutable(builder = false)) // Disable copy methods and builder
+public @interface Tupled {}
```
norrs, line 16:
```
// Look, custom annotation instead of @Value.Immutable
// and the style is also attached!
@Tupled interface RgbDef {
  double red();
  double green();
  double blue();
}
// ...
Rgb color = Rgb.of(0.4, 0.3, 1.0);
````

norrs:
The annoying thing tho is you have to `compileClasses` before they are detected automatically, but I guess there is some trick for getting intellij annotation processor to play better. 

Is it slow? 

The good part is write less code and get a lot for free. 

See https://immutables.github.io/style.html for the way you can customize everything, ie you can keep  `new Foo.Builder()` etc if you want. 

Left this as a PR so you can easily click view diff, be free to close the PR. 

cc @brinxmat @janbj if someone wanted to see how immutables work. I did it as an exercise to get it playing along with gradle. 

IDE requires some enabling: https://immutables.github.io/apt.html

PR #120, opened 2020-12-09
Build with new nva-commons

janbj:


publication-fanout-handler/src/main/java/no/unit/nva/publication/events/PublicationFanoutHandler.java
```diff
@@ -0,0 +1,2 @@
+package no.unit.nva.publication.events;public class PublicationFanoutHandler {
```
janbj, line 1: Ignore this for now

axthosarouris (approved):

PR #122, opened 2020-12-10
NP-1494: remove validation from Fanout handler

axthosarouris:


janbj (approved):
Looks ok if intention is to move validation elsewhere.

PR #127, opened 2020-12-16
Created DeletePublicationHandler

janbj:


delete-publication/src/test/java/no/unit/nva/publication/delete/DeletePublicationHandlerTest.java
```diff
@@ -0,0 +1,104 @@
...
+                .build();
+        doThrow(InputException.class).when(publicationService).markPublicationForDeletion(identifier);
+
+        handler.handleRequest(inputStream, outputStream, context);
+
+        GatewayResponse<Problem> gatewayResponse = toGatewayResponseProblem();
+        assertEquals(HttpStatus.SC_BAD_REQUEST, gatewayResponse.getStatusCode());
+    }
+
+    private GatewayResponse<Void> toGatewayResponse() throws JsonProcessingException {
```
axthosarouris, line 94: Use the GatewayResponse.fromOutputStream() method. You need to declare the outputStream to be a ByteArrayOutputStream. It is OK to do so since we are using the ByteArrayOutputStream::toString method and not the generic OutputStream::toString method.

delete-publication/src/test/java/no/unit/nva/publication/delete/DeletePublicationHandlerTest.java
```diff
@@ -0,0 +1,104 @@
...
+        return objectMapper.readValue(outputStream.toString(),
+                PARAMETERIZED_GATEWAY_RESPONSE_VOID_RESPONSE_TYPE);
+    }
+
+    private GatewayResponse<Problem> toGatewayResponseProblem() throws JsonProcessingException {
```
axthosarouris, line 99: ditto

delete-publication/src/test/java/no/unit/nva/publication/delete/DeletePublicationHandlerTest.java
```diff
@@ -0,0 +1,104 @@
...
+
+import static java.util.Collections.singletonMap;
+import static nva.commons.handlers.ApiGatewayHandler.ALLOWED_ORIGIN_ENV;
+import static nva.commons.utils.JsonUtils.objectMapper;
+import static org.apache.http.HttpStatus.SC_INTERNAL_SERVER_ERROR;
+import static org.junit.Assert.assertEquals;
+import static org.mockito.Mockito.doThrow;
+import static org.mockito.Mockito.when;
+
+public class DeletePublicationHandlerTest {
```
axthosarouris, line 37: We need tests that go through the whole thing. I.e. a handler connected to a service with a local database and sending queries. The tests do not describe the behavior of the handler.

* Does the handler return accepted when the publication is Draft?
* What does the handler return when nhe publicaiton is Public?
* What does it return when the publication does not exist?
* What does it return when the publication is already Drafted for Deletion?
* What does it return when the user is not the owner?   

I think these tests should be in this test class. But I need Rurik's opinion on this

axthosarouris (changes requested):

delete-publication/src/test/java/no/unit/nva/publication/delete/DeletePublicationHandlerTest.java
brinxmat: Agreed

brinxmat (approved):

brinxmat (approved):

PR #153, opened 2021-01-20
Np 1979 transform resource to publication

axthosarouris:


publication-commons/src/main/java/no/unit/nva/publication/service/impl/ResourceService.java
```diff
@@ -186,6 +179,22 @@ public Resource markPublicationForDeletion(Resource resource)
...
 
+    private static List<Resource> queryResultToResourceList(QueryResult result) {
+        return result.getItems()
+            .stream()
+            .map(resultValuesMap -> parseAttributeValuesMap(resultValuesMap, ResourceDao.class))
+            .map(ResourceDao::getResource)
+            .collect(Collectors.toList());
+    }
+
+    private Resource fetchEventuallyConsistentResource(Resource newResource) {
```
brinxmat, line 190: Hmm. In my mind, this action seems like something we shouldn't be doing as the response should be 202 + location header

publication-commons/src/test/java/no/unit/nva/publication/service/impl/ResourceServiceTest.java
```diff
@@ -592,8 +598,20 @@ private void assertThatUpdateFails(Resource resourceUpdate) {
...
+        assertThatNewEntityDescriptionDiffersOnlyInTitle(resource.getEntityDescription(), newEntityDescription);
+        return resource.copy().withEntityDescription(newEntityDescription).build();
+
+
+    }
+
+    private void assertThatNewEntityDescriptionDiffersOnlyInTitle(EntityDescription oldEntityDescription,
+                                                                  EntityDescription newEntityDescription) {
+        Diff diff = javers.compare(oldEntityDescription, newEntityDescription);
+        assertThat(diff.getChanges().size(),is(equalTo(1)));
```
brinxmat, line 614: You probably want to assert that the diff is the title

storage-model/src/test/java/no/unit/nva/publication/storage/model/ResourceTest.java
```diff
@@ -40,27 +84,190 @@ public void copyContainsAllFields() {
...
+            .withDescription(randomString())
+            .withAlternativeTitles(alternativeTitles)
+            .withContributors(sampleContributor())
+            .withLanguage(SAMPLE_LANGUAGE)
+            .withMainTitle(randomString())
+            .withMetadataSource(randomUri())
+            .withNpiSubjectHeading(randomString())
+            .withTags(List.of(randomString()))
+            .withReference(reference)
+
```
brinxmat, line 170: Remove space

brinxmat (changes requested):

publication-commons/src/main/java/no/unit/nva/publication/service/impl/ResourceService.java
axthosarouris: I agree, my intention was to not break the current contract. Maybe we can create a bug issue and fix it later?

brinxmat (approved):

PR #157, opened 2021-01-22
Np 1979 migrate mark for deletion

axthosarouris:


publication-commons/src/main/java/no/unit/nva/publication/service/impl/ResourceService.java
```diff
@@ -304,52 +326,40 @@ private Resource sendUpdateRequest(UpdateItemRequest updateRequest) {
...
             .withKey(dao.primaryKey())
             .withUpdateExpression(updateExpression)
-            .withConditionExpression("#resource.#status <> " + PublicationStatus.PUBLISHED.toString())
+            .withConditionExpression(conditionExpression)
             .withExpressionAttributeNames(expressionNamesMap)
             .withExpressionAttributeValues(expressionValuesMap)
             .withReturnValues(ReturnValue.ALL_NEW);
     }
 
-    private ConcurrentHashMap<String, String> publishUpdateExpressionNamesMap() {
```
axthosarouris, line 327: wrote the queries as strings for better readability. I find the cognitive cost of understanding the queries to be higher than the benefit of reusing some strings. In case of a refactoring tests will fail and the developer will need to update every query separately as indicated by the tests.

publication-commons/src/main/java/no/unit/nva/publication/service/impl/ResourceServiceUtils.java
```diff
@@ -35,10 +35,10 @@
         + " AND "
         + SORT_KEY_NAME_PLACEHOLDER + " = " + SORT_KEY_VALUE_PLACEHOLDER;
 
-    public static final Map<String, String> PRIMARY_KEY_PLACEHOLDERS_AND_ATTRIBUTE_NAMES_MAPPING =
-        primaryKeyAttributeNamesMapping();
+    public static final Map<String, String> PRIMARY_KEY_EQUALITY_CONDITION_ATTRIBUTE_NAMES =
```
axthosarouris, line 38: Trying to use terminology closer to Dynamo's standard terminology

publication-commons/src/main/java/no/unit/nva/publication/service/impl/ResourceServiceUtils.java
```diff
@@ -35,10 +35,10 @@
...
+        primaryKeyEqualityConditionAttributeNames();
 
-    public static final String KEY_EXISTS_CONDITION = keyExistsCondition();
+    public static final String KEY_NOT_EXISTS_CONDITION = keyNotExistsCondition();
```
axthosarouris, line 41: The actual condition is that the key does not exist. See line 112

brinxmat (approved):

PR #160, opened 2021-01-25
Np 1979 doi request dao

axthosarouris:


storage-model/src/main/java/no/unit/nva/publication/storage/model/daos/Dao.java
```diff
@@ -66,7 +54,7 @@ public final void setPrimaryKeyPartitionKey(String key) {
 
     public abstract void setData(R data);
 
-    @JsonProperty(BY_TYPE_CUSTOMER_STATUS_INDEX_PARTITION_KEY_NAME)
```
axthosarouris, line 69: Moved to the interface

brinxmat (approved):

PR #163, opened 2021-01-26
Np 1979 create doi request handler

axthosarouris:


storage-model/src/main/java/no/unit/nva/publication/storage/model/DatabaseConstants.java
```diff
@@ -20,6 +20,7 @@
 
     public static final String CUSTOMER_INDEX_FIELD_PREFIX = "Customer";
     public static final String STATUS_INDEX_FIELD_PREFIX = "Status";
+    public static final String RESOURCE_INDEX_FIELD_PREFIX = "Resource";
```
axthosarouris, line 23: Used in the JoinByResource index: Customer:<CustomerId>:Resource:<ResourceId>

storage-model/src/main/java/no/unit/nva/publication/storage/model/DoiRequest.java
```diff
@@ -30,20 +31,24 @@
     private final Instant createdDate;
     @JsonProperty("customerId")
     private final URI customerId;
-    @JsonAlias("date")
+    @JsonProperty("owner")
     private final String owner;
+    @JsonProperty("resourceTitle")
```
axthosarouris, line 36: Duplicate the resource title in the doi request to facilitate the front-end view of the DoiRequests without joining the Resource with the DoiRequest

storage-model/src/main/java/no/unit/nva/publication/storage/model/daos/DoiRequestDao.java
```diff
@@ -25,6 +25,19 @@ public DoiRequestDao(DoiRequest doiRequest) {
         this.data = doiRequest;
     }
 
+    public static DoiRequestDao queryObject(URI publisherId, String owner, SortableIdentifier doiRequestIdentifier) {
+        DoiRequest doi = new DoiRequest(
+            doiRequestIdentifier,
+            doiRequestIdentifier,
```
axthosarouris, line 31: Null Resource Identifier is not allowed. Subsequent PR has a fix on this with a private constructor and more static generator methods

create-doi-request/build.gradle
```diff
@@ -0,0 +1,41 @@
+repositories {
+    maven { url "https://s3.eu-central-1.amazonaws.com/dynamodb-local-frankfurt/release" }
+}
+
+dependencies {
+
+    compileOnly project(":storage-model")
```
brinxmat, line 7: Choose compile/test following each other, or block of compile/test. Annotate to explain why, because it is still confusing.

create-doi-request/src/main/java/no/unit/nva/doirequest/create/CreateDoiRequestHandler.java
```diff
@@ -0,0 +1,80 @@
...
+
+    public CreateDoiRequestHandler(DoiRequestService requestService, Environment environment) {
+        super(CreateDoiRequest.class, environment, LOGGER);
+        this.doiRequestService = requestService;
+    }
+
+    @Override
+    protected Void processInput(CreateDoiRequest input, RequestInfo requestInfo, Context context)
+        throws ApiGatewayException {
+        URI customerId = requestInfo.getCustomerId().map(URI::create).orElse(null);
```
brinxmat, line 44: I have a feeling that these are in some ways default for API-GW methods and should be provided by the library…probably wrong about this, but I mention it anyway to see if it triggers something useful in your mind

publication-commons/src/test/java/no/unit/nva/publication/service/impl/DoiRequestServiceTest.java
```diff
@@ -0,0 +1,158 @@
...
+import nva.commons.apigateway.exceptions.NotFoundException;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.function.Executable;
+
+public class DoiRequestServiceTest extends ResourcesDynamoDbLocalTest {
+
+    public static final String NOT_THE_RESOURCE_OWNER = "someOther@owner.org";
+    public static final String SOME_USER = "some@user.com";
+    public static final URI SOME_PUBLISHER = URI.create("https://some-publicsher.com");
```
brinxmat, line 37: some-publisher.example.org to ensure we know this is example data

brinxmat (changes requested):

create-doi-request/src/main/java/no/unit/nva/doirequest/create/CreateDoiRequestHandler.java
axthosarouris: The only thing we can do is the conversion of the String to a URI in the commons. But what happens when the value is absent is maybe the responsibility of the caller.  One maybe wishes to throw an exception, another may wish to continue doing something else.

create-doi-request/build.gradle
axthosarouris: fixed

publication-commons/src/test/java/no/unit/nva/publication/service/impl/DoiRequestServiceTest.java
axthosarouris: fixed

brinxmat (approved):

PR #164, opened 2021-01-27
Np 1979 update doi request status when publishing

axthosarouris:
Most changes is import changes due to repackaging


publication-commons/src/main/java/no/unit/nva/publication/service/impl/ResourceService.java
```diff
@@ -158,16 +165,16 @@ public void updateOwner(String identifier, UserInstance oldOwner, UserInstance n
             .collect(Collectors.toList());
     }
 
-    public Publication publishPublication(Publication dto)
-        throws InvalidPublicationException, NotFoundException {
-        return publishResource(Resource.fromPublication(dto)).toPublication();
+    public void publishPublication(UserInstance userInstance, SortableIdentifier resourceIdentifer)
```
axthosarouris, line 168: Cannot return the updated publication anymore because we update the resource and the doi-request at the same time in one transaction and dynamo does not return the updated values.

publication-commons/src/main/java/no/unit/nva/publication/service/impl/ResourceService.java
```diff
@@ -218,22 +225,150 @@ private Void waitBeforeFetching() throws InterruptedException {
...
         ResourceDao fetchedDao = parseAttributeValuesMap(getResult.getItem(), ResourceDao.class);
         return fetchedDao.getData();
     }
 
-    private Resource publishResource(Resource resource) throws NotFoundException, InvalidPublicationException {
-        Resource existingResource = getResource(resource);
-        if (PublicationStatus.PUBLISHED.equals(existingResource.getStatus())) {
-            return existingResource;
+    @SuppressWarnings("rawtypes")
+    private void publishResource(UserInstance userInstance, SortableIdentifier resourceIdentifier)
```
axthosarouris, line 240: update the resource and the possibly present doi-request

storage-model/src/main/java/no/unit/nva/publication/storage/model/daos/JoinWithResource.java
```diff
@@ -40,7 +40,7 @@ default String getByResourcePartitionKey() {
     @JsonProperty(BY_RESOURCE_INDEX_SORT_KEY_NAME)
     default String getByResourceSortKey() {
         return
-            this.getType()
+            this.getOrderedType()
```
axthosarouris, line 43: orderedType is "a:DoiRequest" and "b:Resource" so that we enforce an ordering in the "ByResource" index.

storage-model/src/main/java/no/unit/nva/publication/storage/model/daos/ResourceDao.java
```diff
@@ -22,11 +27,24 @@ public ResourceDao(Resource resource) {
...
+        return new ResourceDao(resource);
+    }
+
     public static String constructPrimaryPartitionKey(URI customerId, String owner) {
         return String.format(PRIMARY_KEY_PARTITION_KEY_FORMAT, Resource.TYPE,
             orgUriToOrgIdentifier(customerId), owner);
     }
 
+    @JsonIgnore
+    public static String getOrderedContainedType() {
```
axthosarouris, line 44: static method version for the "getOrderedType". Static method is necessary for queries, so that we do not generate empty objects to get this value and the non-static method is necessary for implementing the interfaces.

storage-model/src/main/java/no/unit/nva/publication/storage/model/daos/ResourceDao.java
```diff
@@ -51,14 +65,25 @@ public URI getCustomerId() {
...
     @Override
     protected String getOwner() {
         return data.getOwner();
     }
 
     @Override
-    public SortableIdentifier getIdentifier() {
-        return data.getIdentifier();
+    public String getOrderedType() {
+        return getOrderedContainedType();
```
axthosarouris, line 80: read earlier comment

brinxmat (approved):

brinxmat (approved):

PR #173, opened 2021-02-01
Np 1979 dynamodb update event

axthosarouris:


storage-model/src/main/java/no/unit/nva/publication/storage/model/DoiRequest.java
```diff
@@ -154,13 +179,68 @@ public String getOwner() {
 
     @Override
     public String getStatusString() {
-        return getStatus().toString();
+        return Objects.nonNull(getStatus()) ? getStatus().toString() : null;
```
brinxmat, line 182: Just a static import of nonNull

brinxmat (changes requested):

storage-model/src/main/java/no/unit/nva/publication/storage/model/DoiRequest.java
brinxmat: Fixed in next pr

brinxmat (approved):

PR #184, opened 2021-02-09
Np 1979 doi event producer decides on event

axthosarouris:
Doi event producer informs on the kind of event.  A listener can listen to both events or any of them


publication-event-dtopublicationdoi-producer/src/test/resources/resource_update_event_new_image_only_no_doi.json
```diff
@@ -13,12 +13,6 @@
   "detail": {
     "version": "1.0",
     "timestamp": "2021-02-03T15:01:05.418Z",
-    "requestContext": {
```
axthosarouris, line 16: not important part of the json

brinxmat (approved):

PR #194, opened 2021-02-12
Np 1979 modify publication handler validates request

axthosarouris:
Until now, the handler did not check to see if the identifier in the body and the identifier in the path are the same. As a consequence one could update another publication from that specified in the path


update-doi-request/src/main/java/no/unit/nva/doi/requests/handlers/UpdateDoiRequestStatusHandler.java
```diff
@@ -97,16 +96,11 @@ private void validateUser(RequestInfo requestInfo) throws NotAuthorizedException
     private boolean userIsNotAuthorized(RequestInfo requestInfo) {
         return
             !(
-                userHasAccessRight(requestInfo, APPROVE_DOI_REQUEST)
-                && userHasAccessRight(requestInfo, REJECT_DOI_REQUEST)
+                requestInfo.userHasAccessRight(APPROVE_DOI_REQUEST.toString())
```
axthosarouris, line 99: temporary fix until row level security is in place

brinxmat (approved):

PR #197, opened 2021-02-15
Np 1979 create doi request test coverage

axthosarouris:


build.gradle
```diff
@@ -120,15 +120,15 @@ subprojects {
                 limit {
                     counter = 'METHOD'
                     value = 'COVEREDRATIO'
-                    minimum = 0.0
+                    minimum = 0.000
```
axthosarouris, line 123: This will be set to 1 when all modules have test coverage 1.

brinxmat (approved):

PR #204, opened 2021-02-17
Np 1812 insert message ensuring unique identifier

axthosarouris:


brinxmat (approved):

storage-model/src/test/java/no/unit/nva/publication/storage/model/MessageTest.java
```diff
@@ -56,10 +58,10 @@ public void simpleMessageReturnsMessageRelatedToResourceAndNotRelatedToDoiReques
     }
     
     @Test
-    public void simpleMessageReturnsMessageWithAllFieldsFieldIn() {
+    public void simpleMessageReturnsMessageWithAllFieldsFieldInExceptForIdentifier() {
```
axthosarouris, line 61: identifier is being set by the MessageService. The reason for this is that I want to be able to test what is happening when we produce duplicate identifiers.This however is unlikely to happen since our identifier is a timestamp(millis)+UUID. In practice we would have to create biliions of resources per millisecond in order to have a slight possibility of this to happen

brinxmat (approved):

PR #212, opened 2021-02-23
Np 1778 list messages for user includes uri in each message

axthosarouris:


publication-commons/src/main/java/no/unit/nva/publication/model/MessageDto.java
```diff
@@ -23,9 +29,28 @@ public static MessageDto fromMessage(Message message) {
         messageDto.setSenderIdentifier(message.getSender());
         messageDto.setText(message.getText());
         messageDto.setDate(message.getCreatedTime());
+        messageDto.setMessageId(message.getId());
```
axthosarouris, line 32: messageId will be removed from Dao in later PR

torarnet (approved):

PR #225, opened 2021-03-02
Np 2243 resource service allows publication with predefined identifier

axthosarouris:


publication-commons/src/main/java/no/unit/nva/publication/service/impl/ResourceService.java
```diff
@@ -71,22 +71,32 @@ public ResourceService(AmazonDynamoDB client, Clock clock, Supplier<SortableIden
...
-    
+
     public Publication createPublication(Publication inputData) throws TransactionFailedException {
         Resource newResource = Resource.fromPublication(inputData);
         newResource.setIdentifier(identifierSupplier.get());
         newResource.setCreatedDate(clockForTimestamps.instant());
+        return insertResource(newResource);
+    }
+
+    public Publication createPublicationWithPredefinedIdentifier(Publication publication)
```
axthosarouris, line 86: Useful when we migrate publications.

brinxmat (approved):

PR #226, opened 2021-03-04
Np 2243 write all resources to new table

axthosarouris:
This is a first approach for data migration. It gets the work done for now, but it is far from finished.

Problems (not exhaustive list):
1.  Runs only through a test
2. It is slow (not a problem now, because we do not have many records)
3. Part of the implementation is not generic enough and is focused mostly in current migration task


create-message/src/test/java/no/unit/nva/publication/messages/create/CreateMessageHandlerTest.java
```diff
@@ -78,7 +79,7 @@ public void initialize() throws TransactionFailedException {
 
     @Test
     public void handlerStoresMessageWhenCreateRequestIsReceivedByAuthenticatedUser()
-        throws IOException {
+        throws IOException, NotFoundException {
```
axthosarouris, line 82: I start thinking that all excpections should be Unchecked exceptions

data-migration/src/main/java/no/unit/nva/publication/migration/DataMigration.java
```diff
@@ -0,0 +1,88 @@
...
+        );
+    }
+
+    private List<ResourceUpdate> updateMessages() {
+        MessagesImporter messagesImporter = new MessagesImporter(s3Driver, s3DataPath, messageService);
+        List<PublicationMessagePair> allMessages = messagesImporter.geDoiRequestMessages().collect(Collectors.toList());
+
+        return messagesImporter.insertMessages(allMessages.stream());
+    }
+
```
brinxmat, line 64: Annotate with @SafeVarargs, I believe this holds in this case

brinxmat (changes requested):

brinxmat (commented):
As I understand, the update is not transactional, given that DynamoDB has a restriction that up-to ten Items may be updated in one transaction, this becomes problematic. Nevertheless, I suggest avoiding partial import of data, so perhaps check that the full import has succeeded before progressing, or remove the partials and add to report.

brinxmat (approved):

janbj (approved):

PR #234, opened 2021-03-08
Re define resources table

axthosarouris:


cloudformation/nva_resources_dynamodb.yml
```diff
@@ -6,18 +6,7 @@ Parameters:
     Description: Name to use for table persisting NVA resources.
     Type: String
     Default: nva_resources
-  ResourcesByOwnerIndexName:
```
axthosarouris, line 9: The indexes are tightly coupled with the logic of the publicaiton-service and the names need to represent the intention of each index. Any customization of the index names may obscure the purpose of the index.

janbj (approved):

PR #272, opened 2021-03-22
Fixes required for CloudFormation deploy

janbj:


axthosarouris (approved):

template.yaml
```diff
@@ -1066,6 +1064,61 @@ Resources:
...
+              - Effect: Allow
+                Action:
+                  - events:PutEvents
+                Resource: !Ref EventBusArn
+              - Effect: Allow
+                Action:
+                  - sqs:SendMessage
+                  - sqs:ReceiveMessage
+                  - sqs:DeleteMessage
+                Resource: !GetAtt DeleteDraftPublicationHandlerDLQ.Arn
```
axthosarouris, line 1104: Are the event-related rights necessary?

template.yaml
```diff
@@ -1066,6 +1064,61 @@ Resources:
...
+        - PolicyName: LambdaDefaults
+          PolicyDocument:
+            Version: '2012-10-17'
+            Statement:
+              - Effect: Allow
+                Action:
+                  - logs:CreateLogGroup
+                  - logs:CreateLogStream
+                  - logs:PutLogEvents
+                  - lambda:InvokeFunction
```
axthosarouris, line 1089: Maybe the lambda invoke function is not necessary

axthosarouris (changes requested):

template.yaml
janbj: The original SAM policy template added to the the default lambda InvokeFunction role so I think this is necessary.

template.yaml
janbj: If you take a look at the Lambda config you will see that it is necessary:

`      EventInvokeConfig:
        DestinationConfig:
          OnSuccess:
            Type: EventBridge
            Destination: !Ref EventBusArn
          OnFailure:
            Type: SQS
            Destination: !GetAtt DeleteDraftPublicationHandlerDLQ.Arn`

brinxmat (approved):

axthosarouris (approved):

PR #276, opened 2021-03-25
Add Stage Parameter

janbj:


template.yaml
```diff
@@ -66,6 +66,17 @@ Parameters:
     MinValue: 1
     MaxValue: 5
     Description: Max attempt to put events into AWS EventBridge Event Bus. Default is 1.
+  Stage:
+    Type: String
+    Default: Dev
+    AllowedValues:
```
axthosarouris, line 72: How about test?

janbj: Haven't decided yet but I think Test and Sandbox is just another Non-Feature, Non-Prod Stage like Dev.

axthosarouris (approved):

brinxmat (approved):

PR #300, opened 2021-04-12
Lint fetch publication and remove DynamoDbLocal from where not necessary

axthosarouris:
Coninuing to edit dependencies according to gradle-lint's  suggestions. In this PR I removed the package for DynamoDBLocal which although it does not affect the performance (because it is in `testImplementation`), it does affect the gradle lint report and creates noise.
To proceed further with the cleaning it is therefore necessary to remove it from all modules where it it not needed.

### DIFF
--- a/modify-publication/build.gradle
+++ b/modify-publication/build.gradle
@@ -1,5 +1,5 @@
 repositories {
-    maven { url "https://s3.eu-central-1.amazonaws.com/dynamodb-local-frankfurt/release" }
+
 }
 
 dependencies {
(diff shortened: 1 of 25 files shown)


modify-publication/build.gradle
```diff
@@ -1,5 +1,5 @@
 repositories {
-    maven { url "https://s3.eu-central-1.amazonaws.com/dynamodb-local-frankfurt/release" }
+
```
brinxmat, line 2: Delete closure

brinxmat (changes requested):

brinxmat (approved):

PR #302, opened 2021-04-14
Create common s3 ion reader

axthosarouris:
This is a Reader for files in an S3 bucket that are in Amazon's ION format.  We choose to use ION and DynamoDB json format because it is independent of the source that produced it, namely DynamoDb and it will make it easier to convert it to any format we like without too many weird mappings.


data-import-from-s3/src/test/java/no/unit/nva/dataimport/StubS3Driver.java
```diff
@@ -40,15 +41,22 @@ public String getFile(String filename) {
...
     private Stream<String> fileContent(String filename) {
-        InputStream inputStream = IoUtils.inputStreamFromResources(filename);
-        if (streamIsEmpty(inputStream)) {
-            throw new IllegalStateException("File does not exist or file is empty");
-        }
+        InputStream inputStream = attempt(() -> IoUtils.inputStreamFromResources(filename))
+                                      .orElseThrow(fail -> fileNotFoundException());
+
         GZIPInputStream gzipInputStream = attempt(() -> new GZIPInputStream(inputStream)).orElseThrow();
         BufferedReader reader = new BufferedReader(new InputStreamReader(gzipInputStream));
```
svenngjeto, line 48: Maybe force charset onto the inputstreamreader also?  
 BufferedReader reader = new BufferedReader(new InputStreamReader(gzipInputStream , StandardCharsets.UTF_8));

brinxmat (dismissed):

brinxmat (approved):

PR #303, opened 2021-04-14
Use common ion reader in data migration

axthosarouris:


data-import-from-s3/src/main/java/no/unit/nva/dataimport/S3IonReader.java
```diff
@@ -1,51 +1,77 @@
...
 import software.amazon.ion.system.IonReaderBuilder;
 import software.amazon.ion.system.IonTextWriterBuilder;
 
-class S3Reader {
+public class S3IonReader {
 
     public static final String PK0 = DatabaseConstants.PRIMARY_KEY_PARTITION_KEY_NAME;
     // Looking for Strings '<end_of_previous_object>}<possible_white_space>{"Item":{"PK0"'
-    public static final String CONSECUTIVE_JSON_OBJECTS = "(})\\s*(\\{\"Item\":\\{\"" + PK0 + "\")";
+    public static final String CONSECUTIVE_JSON_OBJECTS = "(})\\s*(\\{\"Item\":\\{\".+?\")";
```
axthosarouris, line 30: I am not a great fan of Regular expressions, but I have not found any other solution. Any suggestion welcome. 
The motivation of using the "PK0" in the regex was to avoid  the unlikely event where the content contains a string that the matches the expression. 

Example: 
  A Publication about Json objects where its abstract lists a collection of  Json objects  that are not in an array and each Json object has as first key the word "Item" 

The regex changed because we are using the Reader for any kind of format and not only for entries that follow the current Database pattern (e.g. data migration from the previous format). I consider the scenario of matching the above pattern accidentally quite unlikely as I would expect the quotes to be escaped in a String literal. Therefore I proceed with this solution until a better solution is availalable.

svenngjeto (approved):

data-import-from-s3/src/main/java/no/unit/nva/dataimport/S3IonReader.java
brinxmat: Probably (probably), I'd consider using a streaming parser and checking JsonToken.START_OBJECT, JsonToken.END_OBJECT, which may or may not work.

brinxmat (approved):

PR #309, opened 2021-04-19
Merge doi request handlers

axthosarouris:
Merging handlers that handle the same entities to reduce the fragmentation of the project.
In this PR DoiRequest handlers (Create,List,Update)  are merged into one module.

The only dependency that was not common according to the dependencies was the nva-datamodel-java.  But all three use the Publication class, so it is certainly a runtime dependency for all of them. 

This merging will make the managing and refactoring of the whole service easier.

More merges  will come


.github/workflows/cfn-linter.yaml
```diff
@@ -12,7 +12,7 @@ jobs:
       uses: actions/checkout@v2
 
     - name: cfn-lint
-      uses: scottbrenner/cfn-lint-action@master
+      uses: scottbrenner/cfn-lint-action@main
```
axthosarouris, line 15: Cfn lint fails because of an update in the main branch name

brinxmat (approved):

svenngjeto (approved):

PR #321, opened 2021-04-23
Np 2261 write to dynamo

axthosarouris:
This is a first version of converting CRISTIN records and writing them to dynamo.

The driver loads all Cristin records from a bucket converts them to Publications and writes them to Dynamo. Obviously this does not scale but all the logic is now in place (Except from the mapping that is a quite complex procedure).

The next steps are the following:
1. Deploy the following lambda to the cloud.
2. Split the functionality in two lambdas, One lambda gets all filenames from the bucket and emits an event for each bucket. The other lambda will accept the event and process each file separately. In this way we increase both scalability and parallelization.
3. Improve the mapping procedure to include more fields and replace the hard-coded values with correct ones.

### DIFF
--- a/cristin-import/build.gradle
+++ b/cristin-import/build.gradle
@@ -1,14 +1,16 @@
 apply plugin: 'io.freefair.lombok'
 
 dependencies {
+    implementation project(":publication-constants")
+    implementation project(":publication-commons")
 
-    testImplementation project(":publication-testing")
     implementation(group: 'com.github.bibsysdev', name: 'nva-datamodel-java', version: project.ext.nvaDatamodelVersion) {
         exclude group: 'com.github.bibsysdev', module: 'nva-commons'
     }
 
     implementation group: 'com.github.bibsysdev', name: 's3', version: project.ext.nvaCommonsVersion
     implementation group: 'com.github.bibsysdev', name: 'core', version: project.ext.nvaCommonsVersion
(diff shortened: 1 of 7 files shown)


cristin-import/build.gradle
```diff
@@ -29,4 +38,20 @@ dependencies {
...
+        include "*.so"
+        include "*.dll"
+    }
+    into 'build/dynamodb-local'
+}
+
+test.dependsOn copyNativeDeps
+test.doFirst {
+    systemProperty "java.library.path", 'build/dynamodb-local'
+}
```
axthosarouris, line 57: use DynamoDb local

cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinMapper.java
```diff
@@ -25,9 +27,16 @@ public Publication generatePublication() {
                    .withAdditionalIdentifiers(Set.of(extractIdentifier()))
                    .withEntityDescription(generateEntityDescription())
                    .withCreatedDate(extractEntryCreationDate())
+                   .withPublisher(extractOrganization())
+                   .withOwner(SOME_REAL_OWNER)
                    .build();
     }
 
+    private Organization extractOrganization() {
```
axthosarouris, line 35: temporary value for registering resources. Objective here is to write things to dynamo without necessarily the correct data

brinxmat (approved):

PR #326, opened 2021-04-27
Np 2520 lambda emmits event for s3 filename

axthosarouris:
Lambda function for emitting reading the contents of an S3 folder and emitting an event for each filename in the folder

### DIFF
--- a/cristin-import/build.gradle
+++ b/cristin-import/build.gradle
@@ -10,7 +10,6 @@ dependencies {
 
     implementation group: 'com.github.bibsysdev', name: 's3', version: project.ext.nvaCommonsVersion
     implementation group: 'com.github.bibsysdev', name: 'core', version: project.ext.nvaCommonsVersion
-    implementation group: 'com.github.bibsysdev', name: 'apigateway', version: project.ext.nvaCommonsVersion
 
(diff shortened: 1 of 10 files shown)


cristin-import/build.gradle
```diff
@@ -22,15 +21,22 @@ dependencies {
     implementation group: 'software.amazon.awssdk', name: 'eventbridge', version: '2.16.46'
     implementation(group: 'software.amazon.awssdk', name: 's3', version: project.ext.awsSdk2Version)
 
+    implementation group: 'org.apache.commons', name: 'commons-collections4', version: '4.4'
```
axthosarouris, line 24: Used for partitioning lists instead of writing custom code

cristin-import/src/main/java/no/unit/nva/cristin/lambda/ApplicationConstants.java
```diff
@@ -0,0 +1,10 @@
+package no.unit.nva.cristin.lambda;
+
+public final class ApplicationConstants {
+
+    public static final String EVENT_BUS_NAME = "default";
```
axthosarouris, line 5: These will be filled in later and the values are going to be set by env variables

janbj (approved):

PR #328, opened 2021-04-28
Np 2520 deploy fan out function

axthosarouris:


cristin-import/build.gradle
```diff
@@ -18,8 +18,16 @@ dependencies {
     runtimeOnly group: "com.fasterxml.jackson.datatype", name: "jackson-datatype-jdk8", version: project.ext.jacksonVersion
 
     implementation group: 'com.amazonaws', name: 'aws-java-sdk-dynamodb', version: project.ext.awsSdkVersion
+
+    implementation group: 'software.amazon.awssdk', name: 'aws-core', version: project.ext.awsSdk2Version
```
axthosarouris, line 22: gradle lint

cristin-import/src/main/java/no/unit/nva/cristin/lambda/EventEmitter.java
```diff
@@ -77,19 +89,30 @@ public void addEvents(Collection<T> eventDetails) {
...
 
         return PutEventsRequestEntry.builder()
-                   .eventBusName(ApplicationConstants.EVENT_BUS_NAME)
+                   .eventBusName(EVENT_BUS_NAME)
                    .resources(invokingFunctionArn)
                    .detailType(detailType)
                    .time(Instant.now())
                    .detail(eventDetail.toJsonString())
-                   .source(invokingFunctionArn)
+                   .source("lambda")
```
brinxmat, line 115: Constant

axthosarouris, line 115: this is fixed to be the caller class name in a subsequent PR

cristin-import/src/main/java/no/unit/nva/cristin/lambda/FilenameEventEmitter.java
```diff
@@ -43,32 +52,59 @@ public FilenameEventEmitter(S3Client s3Client, EventBridgeClient eventBridgeClie
...
+    @JacocoGenerated
+    private static S3Client defaultS3Client() {
+        return S3Client.builder()
+                   .region(ApplicationConstants.AWS_REGION)
+                   .httpClient(UrlConnectionHttpClient.create())
+                   .build();
+    }
+
+    private URI createUri(URI s3Location, String filename) {
+        if (s3Location.getPath().endsWith(PATH_SEPARATOR)) {
```
brinxmat, line 79: Ternary to give single return

cristin-import/src/main/java/no/unit/nva/cristin/lambda/FilenameEvent.java
```diff
@@ -5,6 +5,7 @@
 import com.fasterxml.jackson.annotation.JsonProperty;
 import java.net.URI;
 import java.util.Optional;
+import nva.commons.core.JacocoGenerated;
```
axthosarouris, line 8: This class will be removed and importRequest will be reused. So the process is that if an actor sends an ImportRequest to a folder this will be break into several ImportRequests (one per file in folder). This is more generic than having two almost identical classes that do the same thing.

cristin-import/src/test/java/no/unit/nva/cristin/lambda/FilenameEventEmitterTest.java
```diff
@@ -73,15 +77,18 @@ public void handlerThrowsNoExceptionWhenInputIsValid() {
         assertDoesNotThrow(action);
     }
 
-    @Test
-    public void handlerEmitsEvensForEveryFilenameInS3BucketWhenInputIsAnExistingNotEmptyS3Location()
+    @ParameterizedTest(
+        name = "handlerEmitsEventsWithFullFileUriForEveryFilenameInS3BucketWhenInputIsAnExistingNotEmptyS3Location")
+    @ValueSource(strings = {SOME_S3_LOCATION, SOME_S3_LOCATION + PATH_SEPARATOR})
```
brinxmat, line 82: What you're testing here seems to be that the s3-location is valid and has an optional trailing path separator. Consequently, the name of the test should be changed and the ValueSource should be empty string and path separator

brinxmat (changes requested):

cristin-import/src/main/java/no/unit/nva/cristin/lambda/FilenameEventEmitter.java
```diff
@@ -43,32 +52,59 @@ public FilenameEventEmitter(S3Client s3Client, EventBridgeClient eventBridgeClie
...
+
+    @JacocoGenerated
+    private static S3Client defaultS3Client() {
+        return S3Client.builder()
+                   .region(ApplicationConstants.AWS_REGION)
+                   .httpClient(UrlConnectionHttpClient.create())
+                   .build();
+    }
+
+    private URI createUri(URI s3Location, String filename) {
```
axthosarouris, line 78: The S3 client does not return URIs when one is listing files. The URI has to be recomposed. Any better way to create a URI where we do the following are welcome:

```
"https://someHost.org/partial/path" + "continuation/of/the/path" = "https://someHost.org/partial/path/continuation/of/the/path"

```

cristin-import/src/test/java/no/unit/nva/cristin/lambda/FakeEventBridgeClient.java
```diff
@@ -30,8 +35,19 @@ public void close() {
 
     }
 
-    public List<PutEventsRequest> getEventRequests() {
-        return evenRequests;
+    @Override
+    public ListEventBusesResponse listEventBuses(ListEventBusesRequest listEventBusesRequest)
```
axthosarouris, line 39: In line 76 in EventEmitter.java, we make an explicit search for the bus that we are about to write into. This creates the need to mock the respective functionality in our FakeEventBridgeClient. The mocking is returning a (singleton) list of the supplied eventbus name

cristin-import/src/test/java/no/unit/nva/cristin/lambda/FakeEventBridgeClient.java
```diff
@@ -30,8 +35,19 @@ public void close() {
...
+        throws AwsServiceException, SdkClientException {
+        EventBus eventBus = EventBus.builder().name(eventBusName).build();
+        return ListEventBusesResponse.builder().eventBuses(eventBus).build();
+    }
+
+    @Override
+    public PutEventsResponse putEvents(PutEventsRequest putEventsRequest)
```
axthosarouris, line 46: The same as previously.

torarnet (approved):

cristin-import/src/test/java/no/unit/nva/cristin/lambda/FilenameEventEmitterTest.java
```diff
@@ -79,10 +79,11 @@ public void handlerThrowsNoExceptionWhenInputIsValid() {
 
     @ParameterizedTest(
         name = "handlerEmitsEventsWithFullFileUriForEveryFilenameInS3BucketWhenInputIsAnExistingNotEmptyS3Location")
-    @ValueSource(strings = {SOME_S3_LOCATION, SOME_S3_LOCATION + PATH_SEPARATOR})
-    public void handlerEmitsEventsWithFullFileUriForEveryFilenameInS3BucketWhenInputIsAnExistingNotEmptyS3Location()
+    @ValueSource(strings = {"", PATH_SEPARATOR})
```
brinxmat, line 82: Call it EMPTY_STRING

brinxmat (changes requested):

brinxmat (approved):

PR #331, opened 2021-04-29
Np 2521 handler entries fanout handler

axthosarouris:


cristin-import/src/main/java/no/unit/nva/cristin/lambda/EventEmitter.java
```diff
@@ -116,11 +118,15 @@ private PutEventsRequestEntry createPutEventRequestEntry(T eventDetail) {
                    .resources(invokingFunctionArn)
                    .detailType(detailType)
                    .time(Instant.now())
-                   .detail(eventDetail.toJsonString())
+                   .detail(toJson(eventDetail))
```
axthosarouris, line 121: We do not longer expect the generic class to extend JsonSerializable (own interface) because we want to serialize JsonNodes

cristin-import/src/main/java/no/unit/nva/cristin/lambda/FilenameEventEmitter.java
```diff
@@ -35,7 +35,7 @@
 
     public static final String NON_EMITTED_FILENAMES_WARNING_PREFIX = "Some files failed to be emitted:";
     public static final String PATH_SEPARATOR = "/";
-    public static final String CANONICAL_NAME = FilenameEvent.class.getCanonicalName();
+    public static final String CANONICAL_NAME = FilenameEventEmitter.class.getCanonicalName();
```
axthosarouris, line 38: typo

cristin-import/src/main/java/no/unit/nva/cristin/lambda/InputEntriesEventEmitter.java
```diff
@@ -0,0 +1,130 @@
...
+                   .map(array -> toStream(array).collect(Collectors.toList()))
+                   .orElseThrow();
+    }
+
+    private Stream<JsonNode> toStream(ArrayNode root) {
+        return StreamSupport
+                   .stream(Spliterators.spliteratorUnknownSize(root.elements(), Spliterator.ORDERED), SEQUENTIAL);
+    }
+
+    private List<JsonNode> parseContentAsJsonArray(String content) throws JsonProcessingException {
```
brinxmat, line 117: Might prefer this:

```
    private List<JsonNode> parseContentAsJsonArray(String content) throws JsonProcessingException {
        return toStream(getNodeIterator(content))
    }

    private ArrayNode getNodeIterator(String content) throws JsonProcessingException {
        JsonNode jsonNode = JsonUtils.objectMapperNoEmpty.readTree(content);
        if (jsonNode.isArray()) {
            return (ArrayNode) jsonNode;
        } else {
            throw new IllegalArgumentException("Content is not array node");
        }
    }
````

brinxmat (changes requested):

brinxmat (approved):

PR #332, opened 2021-04-29
Np 2521 deploy entries fanout handler

axthosarouris:


template.yaml
```diff
@@ -1199,6 +1199,31 @@ Resources:
...
+      Environment:
+        Variables:
+          AWC_ACCOUNT_ID: !Ref AWS::AccountId
+          TABLE_NAME: !Ref NvaResourcesTable
+          EVENT_BUS: "orestis-test-bus"
+      Events:
+        EventBridgeEvent:
+          Type: EventBridgeRule
+          Properties:
+            EventBusName: "orestis-test-bus"
```
axthosarouris, line 1221: This is a temporary setting for easier monitoring of the events. This will be replaced either with the default bus or another bus that will be define in the template.

brinxmat (approved):

PR #336, opened 2021-04-30
NP-2522: save cristin entries in DynamoDb

axthosarouris:


cristin-import/build.gradle
```diff
@@ -54,6 +57,10 @@ dependencies {
 
 }
 
+test {
+    environment "MAX_SLEEP_TIME", "1"
```
axthosarouris, line 61: If writing to dynamo fails the handler will sleep for some time. But we do not want to sleep for a long time during testing.

publication-commons/src/test/java/no/unit/nva/publication/service/impl/ResourceServiceTest.java
```diff
@@ -106,23 +106,37 @@
     private ResourceService resourceService;
     private Clock clock;
     private DoiRequestService doiRequestService;
-    
+
     @BeforeEach
     public void init() {
         super.init();
-        clock = mock(Clock.class);
```
axthosarouris, line 113: moved to own method

brinxmat (approved):

PR #340, opened 2021-05-03
Np 2524 field transformation

axthosarouris:


cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinTitle.java
```diff
@@ -15,7 +15,8 @@
     public static final String MAIN_TITLE = "titteltekst";
     public static final String LANGUAGE_CODE = "sprakkode";
     public static final String ABSTRACT = "abstract";
-    public static final String VALUE_OF_ORIGINAL_TITLE_STATUS = "J"; //probably means "Ja"
+    public static final String ORIGINAL_TITLE = "J"; //probably means "Ja"
+    public static final String NOT_ORIGINAL_TITLE = "N"; //probably means "Ja"
```
brinxmat, line 19: I guess that "N" probably has a misleading comment. I guess that I might have called these something else, but I don't think it is worth changing

brinxmat (approved):

cristin-import/src/main/java/no/unit/nva/cristin/MappingConstants.java
```diff
@@ -0,0 +1,11 @@
+package no.unit.nva.cristin;
+
+public final class MappingConstants {
+
+    public static String MAIN_CATEGORY_BOOK = "BOK";
```
janbj, line 5: Do you have a source for this? Maybe in documentation for Cristin?

cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinMapper.java
```diff
@@ -52,9 +62,44 @@ private EntityDescription generateEntityDescription() {
...
+        return new Reference.Builder()
+                   .withPublicationInstance(publicationInstance)
+                   .withPublishingContext(publicationContext)
+                   .build();
+    }
+
+    private PublicationContext buildPublicationContext() throws InvalidIsbnException {
+        if (isBook()) {
+            return new Book.Builder()
+                       .withIsbnList(Collections.emptyList())
```
janbj, line 81: We don't map ISBN?

janbj (approved):

cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinMapper.java
axthosarouris: Fields are not yet available

cristin-import/src/main/java/no/unit/nva/cristin/MappingConstants.java
axthosarouris: I am in contact with Daniel to find the source that contains this information

PR #341, opened 2021-05-03
Np 2524 custom owner for resources

axthosarouris:
In this PR we assume that the owner of the imported publications is given in import time as a parameter. This is a temporary solution so that the resources can be registered and be visible as well. 

The flow of events is the following:
```
HumanUser  --[ImportRequest]-->  CristinFilenameEventEmitter  --[ImportRequest]--> CristinEntriesEventEmitter --[FileContentsEvent]--> CristinEventsEventConsumer
```

The parameter is first given in `CristinFilenameEventEmitter` (through the ImportRequest) and is passed down to each subsequent event-consumer/producer until it reached `CristinEventsEventConsumer`. There it is used for generating the NVA publication that is going to be stored in Dynamo


cristin-import/src/main/java/no/unit/nva/cristin/lambda/CristinEntriesEventEmitter.java
```diff
@@ -62,18 +62,24 @@ protected String processInput(ImportRequest input, AwsEventBridgeEvent<ImportReq
         S3Driver s3Driver = new S3Driver(s3Client, input.extractBucketFromS3Location());
         String content = fetchFileFromS3(input, s3Driver);
         List<JsonNode> contents = parseContents(content);
-        EventEmitter<JsonNode> eventEmitter = newEventEmitter(context);
-        eventEmitter.addEvents(contents);
-        List<PutEventsResult> failedEntries = eventEmitter.emitEvents();
+        Stream<FileContentsEvent<JsonNode>> eventBodies = generateEventBodies(input, contents);
```
axthosarouris, line 65: This handler does not need to know the specific structure of the the file contents.

cristin-import/src/test/java/no/unit/nva/cristin/lambda/CristinEntriesEventEmitterTest.java
```diff
@@ -32,32 +35,39 @@
...
-
     private S3Client s3Client;
     private FakeEventBridgeClient eventBridgeClient;
     private CristinEntriesEventEmitter handler;
     private ByteArrayOutputStream outputStream;
 
     @BeforeEach
     public void init() {
-        s3Client = new FakeS3Client(CONTENTS_AS_JSON_ARRAY);
+        s3Client = new FakeS3Client(filesWithContentsAsJsonArrays());
```
axthosarouris, line 54: re-usage of InputStreams requires regeneration of the stream

brinxmat (approved):

PR #345, opened 2021-05-04
merge message modules

axthosarouris:
Part of the task of reducing the number of modules for better overview and management. Notice that these two modules have exactly the same dependencies. This can be seen by the fact that the build.gradle file of list-messages was deleted and no changes were required on the build.gradle file of `create-message` that survived.


publication-commons/src/main/java/no/unit/nva/publication/service/impl/ResourceService.java
```diff
@@ -43,24 +43,24 @@
 
 @SuppressWarnings({"PMD.GodClass", "PMD.AvoidDuplicateLiterals"})
 public class ResourceService extends ServiceWithTransactions {
-    
+
```
axthosarouris, line 46: This class was only linted

publication-commons/src/main/java/no/unit/nva/publication/service/impl/ResourceService.java
```diff
@@ -97,86 +97,86 @@ public Publication insertPreexistingPublication(Publication publication)
         return insertResource(resource);
     }
 
-    private Publication insertResource(Resource newResource) throws TransactionFailedException {
-        TransactWriteItem[] transactionItems = transactionItemsForNewResourceInsertion(newResource);
-        TransactWriteItemsRequest putRequest = newTransactWriteItemsRequest(transactionItems);
-        sendTransactionWriteRequest(putRequest);
+    public Publication markPublicationForDeletion(UserInstance userInstance,
```
axthosarouris, line 100: automatic re-ordering of the methods

brinxmat (approved):

PR #351, opened 2021-05-06
refactoring: move classes to common module before generalizing them

axthosarouris:
Importing from an S3 folder consists basically of 3 steps:
 1. list all files in folder and emit one event per file
 2. read all entries in file and emit one event per entry
 3. process entry

Each step is implemented as a Lambda handler and the steps are connected through EventBridge events.  
With the assumption that the content of the files is Json (or Ion) content, then the first two steps are quite general and can be used in many scenarios (e.g. data import, data-migration etc.) 

In the current state, the handlers that implement the first 2 steps are not general enough to handle other cases, but they can be easily refactored and generalized. This is the first step of the refactoring  and it consists of simply moving the classes to be refactored to a different module. The next step will be to refactor them in place


cristin-import/build.gradle
```diff
@@ -21,27 +22,24 @@ dependencies {
...
-    implementation group: 'software.amazon.awssdk', name: 'aws-core', version: project.ext.awsSdk2Version
-    implementation group: 'software.amazon.awssdk', name: 'sdk-core', version: project.ext.awsSdk2Version
-    implementation group: 'software.amazon.awssdk', name: 'eventbridge', version: '2.16.46'
-    implementation group: 'software.amazon.awssdk', name: 's3', version: project.ext.awsSdk2Version
-    implementation group: 'software.amazon.awssdk', name: 'url-connection-client', version: project.ext.awsSdk2Version
-    implementation group: 'software.amazon.awssdk', name: 'http-client-spi', version: project.ext.awsSdk2Version
     implementation group: 'software.amazon.awssdk', name: 'regions', version: project.ext.awsSdk2Version
 
-    implementation group: 'org.apache.commons', name: 'commons-collections4', version: '4.4'
+    runtimeOnly group: 'software.amazon.awssdk', name: 'url-connection-client', version: project.ext.awsSdk2Version
```
axthosarouris, line 27: gradle lint changes

brinxmat (approved):

PR #352, opened 2021-05-06
CristinFilenameEventEmitter becomes FilenameEventEmitter

axthosarouris:
CristinFilenameEventEmitter acquires a more generic functionality of emitting files for any kind import event. It sends in the body an import event type which later can server as event detail-type. (https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-events.html)

### DIFF
--- a/s3-import-commons/src/main/java/no/unit/nva/publication/s3imports/FilenameEventEmitter.java
+++ b/s3-import-commons/src/main/java/no/unit/nva/publication/s3imports/FilenameEventEmitter.java
@@ -29,6 +29,13 @@ import software.amazon.awssdk.services.eventbridge.model.PutEventsRequest;
 import software.amazon.awssdk.services.eventbridge.model.PutEventsRequestEntry;
 import software.amazon.awssdk.services.s3.S3Client;
 
+/**
+ * {@link FilenameEventEmitter} accepts an {@link ImportRequest}, it lists all the files in the S3 location defined in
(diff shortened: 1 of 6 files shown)


s3-import-commons/src/main/java/no/unit/nva/publication/s3imports/FilenameEventEmitter.java
```diff
@@ -114,12 +122,18 @@ private void logWarningForNotEmittedFilenames(List<PutEventsResult> failedReques
                                eventBridgeClient);
 
         List<ImportRequest> filenameEvents = files.stream()
-                                                 .map(uri -> new ImportRequest(uri, publicationsOwner))
+                                                 .map(uri -> newImportRequestForSingleFile(importRequest, uri))
                                                  .collect(Collectors.toList());
         eventEmitter.addEvents(filenameEvents);
```
janbj, line 127: A FilenameEvent is an ImportRequest?

janbj (approved):

s3-import-commons/src/main/java/no/unit/nva/publication/s3imports/FilenameEventEmitter.java
axthosarouris: yes. The class FilenameEvent does not exist any longer. One can have an import request for a folder or for a file

PR #355, opened 2021-05-07
Import requests event bus

axthosarouris:
Replace manually made event bus for s3 imports with new event-bus defined in template. The new bus definition makes use of the "suffix" variable for creating  a new bus for every feature branch

### DIFF
--- a/template.yaml
+++ b/template.yaml
@@ -62,10 +62,6 @@ Parameters:
     Type: String
     Default: ''
     Description: Suffix used for naming resources for feature branches to avoid conflicts.
-    #temporary parameter
-  CristinImportEventBusName:
-    Type: String
-    Default: orestis-test-bus
 
 Conditions:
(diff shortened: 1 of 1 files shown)


brinxmat (approved):

template.yaml
```diff
@@ -1291,9 +1293,14 @@ Resources:
...
           TABLE_NAME: !Ref NvaResourcesTable
-          EVENT_BUS: !Ref CristinImportEventBusName
+          EVENT_BUS: !GetAtt ImportRequestsEventBus.Name
 
 
+  #=============================Event Buses =============================================================
+  ImportRequestsEventBus:
+    Type: AWS::Events::EventBus
+    Properties:
+      Name: !Join [ '',[ 'import-requests-event-bus',!Ref Suffix ] ]
```
axthosarouris, line 1303: The Suffix allows the creation of a separate event-bus for every feature branch when the new pipeline will be in place

PR #356, opened 2021-05-07
Delete old migration code

axthosarouris:
Delete the code for migrating from the old database structure to the new structure. The new migration will be in different direction and the existing code will have to be totally transformed. However, after the creation of the general importer handlers, most of the code is not necessary any more and a much smaller module should take its place


template.yaml
```diff
@@ -1215,7 +1215,7 @@ Resources:
         Variables:
           AWC_ACCOUNT_ID: !Ref AWS::AccountId
           TABLE_NAME: !Ref NvaResourcesTable
-          EVENT_BUS: !Ref CristinImportEventBusName
+          EVENT_BUS: !GetAtt ImportRequestsEventBus.Name
```
axthosarouris, line 1218: small bugfix

brinxmat (approved):

PR #360, opened 2021-05-11
move ImportResult to s3-import-commons and use it in data-import-from-s3

axthosarouris:


data-import-from-s3/src/main/java/no/unit/nva/dataimport/DataImportHandler.java
```diff
@@ -50,26 +51,34 @@ public DataImportHandler(S3Driver s3Driver, AmazonDynamoDB dynamoClient) {
         this.dynamoClient = dynamoClient;
     }
 
-    public ImportResult[] importAllFilesFromFolder(Map<String, String> request) {
+    public List<ImportResult<FailedDynamoEntriesReport>> importAllFilesFromFolder(Map<String, String> request) {
```
axthosarouris, line 54: This handler will change a lot in the future. I could delete it but I am keeping it only for maintaining the functionality.

torarnet (approved):

PR #372, opened 2021-05-18
Np 2575 dummy file for all resources

axthosarouris:
Create a dummy file for all imported Cristin Resources, so that they can be published and therefore indexed and looked up. This will enable the PO to give more detailed feedback on the mapping of the fieds

### DIFF
--- a/cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinMapper.java
+++ b/cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinMapper.java
@@ -1,5 +1,7 @@
 package no.unit.nva.cristin.mapper;
 
+import static no.unit.nva.cristin.lambda.constants.MappingConstants.DUMMY_UUID;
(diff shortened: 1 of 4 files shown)


cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinMapper.java
```diff
@@ -43,11 +54,30 @@ public Publication generatePublication() {
                    .withPublisher(extractOrganization())
                    .withOwner(cristinObject.getPublicationOwner())
                    .withStatus(PublicationStatus.DRAFT)
+                   .withFileSet(createFileSetPointingToTheNilFile())
+                   .build();
+    }
+
+    private static File createNonExistentFile() {
```
brinxmat, line 61: Turns out FE doesn't use URIs for llicenses either… :( should probably create a ticket for that…

brinxmat (approved):

PR #374, opened 2021-05-19
NP-2576: report unknown categories to s3

axthosarouris:
I tried to keep the PR small, but I felt that refactoring to increase the readability would make it easier.

Presently the only known category is Book-Anthology.  Any other category will be reported as an error in S3.
Because of the fact that the CRISTIN categories are enums that reflect our incomplete knowledge of the actual categories, we have to parse the contents of the event inside the handler in order to be able to report the original content in the error report file.

### DIFF
--- a/cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinSecondaryCategory.java
+++ b/cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinSecondaryCategory.java
@@ -5,14 +5,14 @@ import com.fasterxml.jackson.annotation.JsonValue;
 import java.util.Map;
 
 public enum CristinSecondaryCategory {
-    ANTHOLOGY;
+    ANTHOLOGY, TEMPORARILY_UNKNOWN, UNMAPPED;
 
-    private static final Map<String, CristinSecondaryCategory> ALIASES_MAP = createAliasesMap();
+    private static final Map<String, CristinSecondaryCategory> KNOWN_ALIASES_MAP = knownAliases();
     private static final Map<CristinSecondaryCategory, String> DEFAULT_NAMES_MAP = defaultNamesMap();
 
     @JsonCreator
     public static CristinSecondaryCategory fromString(String category) {
-        return ALIASES_MAP.get(category);
+        return KNOWN_ALIASES_MAP.getOrDefault(category, UNMAPPED);
     }
 
     @JsonValue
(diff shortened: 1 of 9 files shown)


cristin-import/src/test/java/no/unit/nva/cristin/lambda/CristinEntryEventConsumerTest.java
```diff
@@ -84,149 +113,234 @@ public void handlerReturnsAnNvaPublicationEntryWhenInputIsEventWithCristinResult
...
-    public void handlerSavesPublicationToDynamoDbWhenInputIsEventWithCristinResult() throws JsonProcessingException {
-        String input = validEventToJsonString();
-        handler.handleRequest(stringToStream(input), outputStream, CONTEXT);
-        CristinObject cristinObject = generatePublicationFromResource(input);
-        Publication expectedPublication = cristinObject.toPublication();
-        UserInstance userInstance = extractUserInstance(expectedPublication);
+    public void handlerSavesPublicationToDynamoDbWhenInputIsEventWithCristinResult() {
+        CristinObject cristinObject = cristinDataGenerator.randomObject();
+        AwsEventBridgeEvent<FileContentsEvent<JsonNode>> awsEvent = cristinDataGenerator.toAwsEvent(cristinObject);
+        InputStream input = stringToStream(awsEvent.toJsonString());
```
axthosarouris, line 119: These 3 lines are repeated throughout the code and they are not a method because we need all 3 values in most cases.

cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinSecondaryCategory.java
```diff
@@ -5,26 +5,31 @@
...
+        return UNMAPPED.equals(this);
+    }
+
+    private static Map<String, CristinSecondaryCategory> knownAliases() {
         return Map.of("ANTOLOGI", ANTHOLOGY);
     }
 
     private static Map<CristinSecondaryCategory, String> defaultNamesMap() {
-        return Map.of(ANTHOLOGY, "ANTOLOGI");
+        return Map.of(ANTHOLOGY, "ANTOLOGI",
```
brinxmat, line 32: can this be enum field 'mapping'?

brinxmat (dismissed):

brinxmat (approved):

PR #381, opened 2021-05-21
Cleanup phase 2

axthosarouris:


data-import-from-s3/src/main/java/no/unit/nva/dataimport/DataImportHandler.java
```diff
@@ -51,7 +51,7 @@ public DataImportHandler(S3Driver s3Driver, AmazonDynamoDB dynamoClient) {
         this.dynamoClient = dynamoClient;
     }
 
-    public List<ImportResult<FailedDynamoEntriesReport>> importAllFilesFromFolder(Map<String, String> request) {
+    public List<ImportResult<FailedDynamoEntriesReport>> handleRequest(Map<String, String> request) {
```
axthosarouris, line 54: change handler's name function to follow the de-facto pattern

template.yaml
```diff
@@ -1253,11 +1253,11 @@ Resources:
 
 
   #==========================Manually triggered functions=============================================
-  ImportFromS3Function:
+  ImportDynamoImageFromS3:
```
axthosarouris, line 1256: Change function names for better understanding

brinxmat (approved):

torarnet (approved):

PR #383, opened 2021-05-23
Fix mapping of roles and hard code unknown fileds (series title, series number)

axthosarouris:


cristin-import/src/test/java/no/unit/nva/cristin/mapper/CristinMapperTest.java
```diff
@@ -233,6 +243,75 @@ public void mapReturnsPublicationWithHardcodedLink() {
...
+                                                .getContributors()
+                                                .stream()
+                                                .filter(c -> c.getRole().equals(expectedNvaRole))
+                                                .findAny();
+        assertThat(contributor.isPresent(), is(true));
+        assertThat(contributor.orElseThrow().getRole(), is(equalTo(expectedNvaRole)));
+    }
+
+    @Test
+    public void mapReturnsPublicationsWhereCristinCreatorRoleIsMappedToNvaCreatorRole() {
```
torarnet, line 269: Is this name the right one? you cast an exception and assert that, you do not assert that you have a valid map of publications with their roles? or am i wrong?

torarnet (commented):
See comment in test class

cristin-import/src/test/java/no/unit/nva/cristin/mapper/CristinMapperTest.java
axthosarouris: it was the wrong test name. In general, I am taking the approach that if I do not know the mapping of a Crstin value to an NVA value, it is an error. The mapping will be finished when there are no errors left

torarnet (approved):

PR #395, opened 2021-05-31
NP-2616: refactor UpdatePublicationHandler tests

axthosarouris:
Rewrote all the tests by using a "real" service connected to a local database instead of mocking it. In this way we can test more complicated scenarios all the way down to writing to the database without extensive mocking and replicating logic

### DIFF
--- a/update-publication/src/test/java/no/unit/nva/publication/update/UpdatePublicationHandlerTest.java
+++ b/update-publication/src/test/java/no/unit/nva/publication/update/UpdatePublicationHandlerTest.java
@@ -100,7 +100,7 @@ public class UpdatePublicationHandlerTest extends ResourcesDynamoDbLocalTest {
 
         updatePublicationHandler.handleRequest(inputStream, output, context);
         GatewayResponse<PublicationResponse> gatewayResponse = GatewayResponse.fromOutputStream(output);
-        PublicationResponse body = gatewayResponse.getBodyObject(PublicationResponse.class);
+        final PublicationResponse body = gatewayResponse.getBodyObject(PublicationResponse.class);
         assertEquals(SC_OK, gatewayResponse.getStatusCode());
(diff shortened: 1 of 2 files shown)


update-publication/src/test/java/no/unit/nva/publication/update/UpdatePublicationHandlerTest.java
```diff
@@ -28,100 +27,88 @@
...
 
         output = new ByteArrayOutputStream();
         updatePublicationHandler =
             new UpdatePublicationHandler(publicationService, environment);
         publication = createPublication();
     }
-    
-    @Test
-    @DisplayName("handler returns OK Response on ReportBasic input")
-    public void handlerReturnsOkResponseOnReportBasicInput() throws IOException, ApiGatewayException {
```
axthosarouris, line 97: This is a very specific test and I do not whether it is necessary to keep it.

update-publication/src/test/java/no/unit/nva/publication/update/UpdatePublicationHandlerTest.java
```diff
@@ -28,100 +27,88 @@
...
-        InputStream inputStream = generateInputStreamWithValidBodyAndHeadersAndPathParameters(
-            modifiedPublication.getIdentifier()).build();
-        updatePublicationHandler.handleRequest(inputStream, output, context);
-        GatewayResponse<PublicationResponse> gatewayResponse = toGatewayResponse();
-        assertEquals(SC_OK, gatewayResponse.getStatusCode());
-    }
-    
+
     @Test
-    @DisplayName("handler Returns OK Response On Valid Input")
```
axthosarouris, line 110: This test has changed to handler returns OK when Input is valid and user is the resource owner

brinxmat (dismissed):

brinxmat (approved):

PR #406, opened 2021-06-03
Np 2678 remove publications owner from import request

axthosarouris:
When starting a Cristin import the user was required to supply the publications owner. This is no longer necessary and the value is now hardcoded ton arbitrary value.


cristin-import/src/test/java/no/unit/nva/cristin/lambda/CristinEntryEventConsumerTest.java
```diff
@@ -146,20 +146,6 @@ private void injectValuesThatAreCreatedWhenSavingInDynamo(AwsEventBridgeEvent<Fi
         expectedPublication.setPublishedDate(actualPublication.getPublishedDate());
     }
 
-    @Test
-    public void handlerSavesPublicationWithOwnerBeingTheSuppliedByTheEventOwner() {
```
axthosarouris, line 150: This requirement is no longer true. The publications owner is hardcoded now

cristin-import/build.gradle
```diff
@@ -42,7 +42,6 @@ dependencies {
     testImplementation(group: 'org.javers', name: 'javers-core', version: project.ext.javersVersion) {
         because "For comparing java objects and detailing the differences in readable way"
     }
-    testImplementation group: 'com.github.bibsysdev', name: 'apigateway', version: project.ext.nvaCommonsVersion
```
axthosarouris, line 45: removed according to directions from gradle-lint

torarnet (approved):

torarnet (commented):
Will approve when you fix failing tests in Pull Request Github Action

torarnet (approved):

PR #407, opened 2021-06-03
NP-2677: import event type is set as an env variable

axthosarouris:
You can get the whole context of the task here: https://unit.atlassian.net/browse/NP-2677

### DIFF
--- a/s3-import-commons/build.gradle
+++ b/s3-import-commons/build.gradle
@@ -59,6 +59,7 @@ dependencies {
 
 test {
     environment "MAX_SLEEP_TIME", "1"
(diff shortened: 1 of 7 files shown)


s3-import-commons/build.gradle
```diff
@@ -59,6 +59,7 @@ dependencies {
 
 test {
     environment "MAX_SLEEP_TIME", "1"
+    environment "IMPORT_EVENT_TYPE", "importEventTypeTestingValue"
```
axthosarouris, line 62: set env variable for testing. The env variable is read as a constant for simplicity, so we need to set it in the test environment.

torarnet (approved):

brinxmat (approved):

PR #436, opened 2021-06-22
gradle7: migration to gradle 7.

axthosarouris:
This is a first step for migrating to gradle 7. More steps need to be done regarding the jacocoReport, but the current version, although not the cleanest, is cleaner than the previous one.


build.gradle
```diff
@@ -1,192 +1,29 @@
-plugins {
```
axthosarouris, line 1: all moved to buildSrc as a plugin

build.gradle
```diff
@@ -1,192 +1,29 @@
...
-    }
-
-    test {
-        useJUnitPlatform {
-            excludeTags("RemoteTest")
-        }
-        failFast = true
-        testLogging {
-            events 'skipped', 'passed', 'failed'
+    configurations.testImplementation.canBeResolved = true
```
axthosarouris, line 11: This part is for using DynamoDB locally

buildSrc/build.gradle
```diff
@@ -0,0 +1,21 @@
...
+    gradlePluginPortal()
+}
+
+wrapper {
+    distributionType = Wrapper.DistributionType.ALL
+    gradleVersion = "7.0.2"
+}
+
+dependencies {
+    implementation group: 'nebula.lint', name: 'nebula.lint.gradle.plugin', version: '16.23.0'
```
axthosarouris, line 15: the versions of the plugins are set as normal dependencies in the building of the plugin

buildSrc/src/main/groovy/nva-publication-api.java-common-conventions.gradle
```diff
@@ -0,0 +1,156 @@
...
+    failFast = true
+    testLogging {
+        events 'skipped', 'passed', 'failed'
+    }
+
+    finalizedBy jacocoTestReport
+    //TODO: investigate the usage of a BuildConfig file with BuildConfig plugin
+    environment "API_HOST", "localhost"
+    environment "API_SCHEME", "https"
+}
```
axthosarouris, line 52: I am not sure if these settings should be defined in the plugin. If yes, then we are taking about a very specialized plugin that cannot be published.

buildSrc/src/main/groovy/nva-publication-api.java-common-conventions.gradle
```diff
@@ -0,0 +1,156 @@
...
+                }
+                scm {
+                    connection = 'scm:git:git://github.com/BIBSYSDEV/nva-publication-api.git'
+                    developerConnection = 'scm:git:ssh://github.com/BIBSYSDEV/nva-publication-api.git'
+                    url = 'https://github.com/BIBSYSDEV/nva-publication-api/tree/main'
+                }
+            }
+        }
+    }
+}
```
axthosarouris, line 156: Similarlly, not sure if this fits into a generalized plugin

gradle/libs.versions.toml
```diff
@@ -0,0 +1,76 @@
+[versions]
```
axthosarouris, line 1: Here we have all dependencies. Each module can choose the ones they need

publication-commons/build.gradle
```diff
@@ -1,77 +1,46 @@
...
-    }
-    into 'build/dynamodb-local'
-}
-
-test.dependsOn copyNativeDeps
-test.doFirst {
-    systemProperty "java.library.path", 'build/dynamodb-local'
 }
 
-// Build and package publication-commons and dependencies for Lambda Layer
```
axthosarouris, line 60: All logic used for Lambda layers is removed because gradle was complaining and we are not using layers anymore

buildSrc/src/main/groovy/nva-publication-api.java-common-conventions.gradle
brinxmat: This feels like a more global plugin — some things will be globally true, others not

brinxmat (dismissed):

brinxmat (approved):

PR #440, opened 2021-06-25
NP-2794 Create DynamoDB index GSI for cristinId

Eirik-Eide-Saur:


storage-model/src/test/java/no/unit/nva/publication/storage/model/daos/WithCristinIdentifierTest.java
```diff
@@ -0,0 +1,83 @@
...
+
+public class WithCristinIdentifierTest extends ResourcesDynamoDbLocalTest {
+
+    public void init() {
+        super.init();
+    }
+
+    @ParameterizedTest
+    @MethodSource("withPrimaryKeyInstancesProvider")
+    public void dynamoDbClientReturnsItemWithMatchingCristinId(List<ResourceDao> daos) {
```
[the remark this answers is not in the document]
axthosarouris: I am not sure why is this a parametrized test. Do you want check the resources one by one? Why isn't one resource enough?

storage-model/src/test/java/no/unit/nva/publication/storage/model/daos/WithCristinIdentifierTest.java
```diff
@@ -0,0 +1,83 @@
...
+        init();
```
axthosarouris, line 36: Why do we need to call init() again? It is called automatically before each test

storage-model/src/test/java/no/unit/nva/publication/storage/model/daos/WithCristinIdentifierTest.java
```diff
@@ -0,0 +1,83 @@
...
+                .collect(SingletonCollector.collectOrElse(null));
+    }
+
+    private Map<String, Condition> createConditionsWithCristinId(String cristinId) {
+        Condition condition = new Condition()
+                .withComparisonOperator(ComparisonOperator.EQ)
+                .withAttributeValueList(new AttributeValue(cristinId));
+        return Map.of(RESOURCES_BY_CRISTIN_ID_INDEX_PARTITION_KEY_NAME, condition);
+    }
+}
```
axthosarouris, line 83: These methods should be in the new interface. Now half of the logic for querying the data lies outside the interface. 
The idea is that we should be able to create a query object (or use an existing object) and use the object to query client.  Also since you wrote the logic, why not put it in the class so that we can use it in the future?

Example (in case we only know the Cristin Identifier)
```
Resource queryResource = Resource.builder()
                .withAdditionalIdentifiers(Set.of(new AdditionalIdentifier("Cristin",randomString())))
                .build();
        ResourceDao query = new ResourceDao(queryResource);
        client.query(query.getByCristinId())
```

storage-model/src/test/java/no/unit/nva/publication/storage/model/daos/WithCristinIdentifierTest.java
```diff
@@ -0,0 +1,83 @@
...
+
+import static no.unit.nva.publication.storage.model.DatabaseConstants.RESOURCES_BY_CRISTIN_ID_INDEX_PARTITION_KEY_NAME;
+import static no.unit.nva.publication.storage.model.DatabaseConstants.RESOURCES_TABLE_NAME;
+import static no.unit.nva.publication.storage.model.DatabaseConstants.RESOURCE_BY_CRISTIN_ID_INDEX_NAME;
+import static no.unit.nva.publication.storage.model.daos.DaoUtils.sampleResourceDao;
+import static no.unit.nva.publication.storage.model.daos.DaoUtils.toPutItemRequest;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.containsInAnyOrder;
+
+public class WithCristinIdentifierTest extends ResourcesDynamoDbLocalTest {
```
axthosarouris, line 27: We are missing a test where we insert a mixture of objects with and without CristinId and when we scan the index we expect to find only the one with the CristinId. Pay attention that a mixture of objects can contain only two objects, so the test does not need to be very complicated.

axthosarouris (changes requested):

storage-model/src/test/java/no/unit/nva/publication/storage/model/daos/WithCristinIdentifierTest.java
Eirik-Eide-Saur: It is not tagged with the @BeforeEach, so not calling it makes the test fail. Seeing that I am to add another test it would probably be advantageous to git it the BeforeEach and not calling it in either method

storage-model/src/main/java/no/unit/nva/publication/storage/model/daos/ResourceDao.java
```diff
@@ -78,6 +86,21 @@ public SortableIdentifier getIdentifier() {
...
 
+
+    @Override
+    public String getCristinIdentifier() {
+        return Optional.ofNullable(data.getAdditionalIdentifiers()).stream()
+                .flatMap(Collection::stream)
+                .filter(this::keyEqualsCristin)
+                .map(AdditionalIdentifier::getValue)
+                .collect(SingletonCollector.collectOrElse(null));
+
```
axthosarouris, line 97: better to return an Optional here to indicate the possibility of non-existence.

storage-model/src/main/java/no/unit/nva/publication/storage/model/daos/WithCristinIdentifier.java
```diff
@@ -0,0 +1,51 @@
...
+import static no.unit.nva.publication.storage.model.DatabaseConstants.RESOURCES_BY_CRISTIN_ID_INDEX_SORT_KEY_NAME;
+import static no.unit.nva.publication.storage.model.DatabaseConstants.RESOURCES_TABLE_NAME;
+import static no.unit.nva.publication.storage.model.DatabaseConstants.RESOURCE_BY_CRISTIN_ID_INDEX_NAME;
+import static no.unit.nva.publication.storage.model.DatabaseConstants.RESOURCE_INDEX_FIELD_PREFIX;
+
+public interface WithCristinIdentifier {
+
+    @JsonProperty(RESOURCES_BY_CRISTIN_ID_INDEX_PARTITION_KEY_NAME)
+    default String getResourceByCristinIdPartitionKey() {
+        return getCristinIdentifier() == null ? getCristinIdentifier()
```
axthosarouris, line 24: if getCristinIdentifier returns optional then this becomes simpler. Also ` ? null`  make it clearer what we are returning. When you use the ternary operator set the condition and each directive in separate lines.

storage-model/src/main/java/no/unit/nva/publication/storage/model/daos/WithCristinIdentifier.java
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    @JsonProperty(RESOURCES_BY_CRISTIN_ID_INDEX_SORT_KEY_NAME)
+    default String getResourceByCristinIdSortKey() {
+        return RESOURCE_INDEX_FIELD_PREFIX + KEY_FIELDS_DELIMITER + getIdentifier();
+    }
+
+    SortableIdentifier getIdentifier();
+
+    String getCristinIdentifier();
```
axthosarouris, line 35: Add @JsonIgnore in both getters. Add a test verifying that these values are not serialized.

storage-model/src/test/java/no/unit/nva/publication/storage/model/ResourceTest.java
```diff
@@ -162,8 +163,9 @@ public Publication samplePublication(Reference reference) {
     }
 
     private Set<AdditionalIdentifier> sampleAdditionalIdentifiers() {
-        AdditionalIdentifier identifier = new AdditionalIdentifier(randomString(), randomString());
-        return Set.of(identifier);
+        AdditionalIdentifier identifier1 = new AdditionalIdentifier(CRISTIN_SOURCE, randomString());
```
axthosarouris, line 166: pick better names. e.g. cirstinIdentifer, otherRandomIdentifier.

storage-model/src/test/java/no/unit/nva/publication/storage/model/daos/WithCristinIdentifierTest.java
```diff
@@ -0,0 +1,78 @@
...
+                        .withIndexName(RESOURCE_BY_CRISTIN_ID_INDEX_NAME));
+        assertThat(result.getCount(), is(equalTo(1)));
+    }
+
+    @Test
+    public void getResourceByCristinIdPartitionKeyReturnsANullValueWhenObjectHasNoCristinIdentifier()
+            throws MalformedURLException, InvalidIssnException {
+        ResourceDao daoWithoutCristinId = createResourceDaoWithoutCristinIdentifier();
+        assertThat(daoWithoutCristinId.getResourceByCristinIdPartitionKey(), is(equalTo(null)));
+    }
```
axthosarouris, line 58: This belongs to the ResourceDaoTrest

axthosarouris (changes requested):

axthosarouris:
You should update the cloud formation template as well.

storage-model/src/main/java/no/unit/nva/publication/storage/model/daos/WithCristinIdentifier.java
Eirik-Eide-Saur: No idea how to test if something i serialized or not

storage-model/src/test/java/no/unit/nva/publication/storage/model/daos/ResourceDaoTest.java
```diff
@@ -41,4 +45,12 @@ public void constructPrimaryPartitionKeyReturnsStringContainingTypePublisherAndO
                              + SAMPLE_USER;
         assertThat(primaryPartitionKey, is(equalTo(expectedKey)));
     }
+
+    @Test
+    public void getResourceByCristinIdPartitionKeyReturnsANullValueWhenObjectHasNoCristinIdentifier()
+            throws MalformedURLException, InvalidIssnException {
+        ResourceDao daoWithoutCristinId = WithCristinIdentifierTest.createResourceDaoWithoutCristinIdentifier();
+        assertThat(daoWithoutCristinId.getResourceByCristinIdentifierPartitionKey(),
+                Matchers.is(Matchers.equalTo(null)));
```
axthosarouris, line 54: Use static imports

axthosarouris (changes requested):

axthosarouris (approved):

PR #441, opened 2021-06-25
NP-2861: report unknown Cristin properties.

axthosarouris:


cristin-import/src/test/java/no/unit/nva/cristin/lambda/CristinEntryEventConsumerTest.java
```diff
@@ -301,6 +308,33 @@ public void savesFileInInputFolderErrorExceptionNameInputFileLocationInputFileWh
...
+        S3Driver s3Driver = new S3Driver(s3Client, NOT_IMPORTANT);
+        String file = s3Driver.getFile(expectedFilePath);
+
+        assertThat(file, is(not(emptyString())));
+        assertThat(file, containsString(UNKNOWN_PROPERTY_NAME_IN_RESOURCE_FILE_WITH_UNKNOWN_PROPERTY));
+    }
+
+    // This test will fail until we have mapped all properties in the resource file or we
+    // have removed (from the resource and the import) the ones that we do not want to keep.
+    @Disabled
```
axthosarouris, line 330: Test for helping spotting unknown fields. The alternative would be to run the job in AWS multiple times (once for each unknown field), or create some complex logic that lists all unmapped fields. With this test one can verify that they have mapped all fields.

brinxmat (approved):

PR #442, opened 2021-06-30
NP-2878: add fields to "ignored lists" when they are not used

axthosarouris:
Exported Cristin data, contain many fields that we currently do not use. To keep track on the fields that we are not using we require that the mapping fails when there is at least one field that we do not parse explicitly or we do not ignore explicitly. 
By "ignoring explicitly" we mean to add the field in the `@JsonIgnoreProperties` annotation in each object that we need to ignore a field. 
The ultimate target is to require that the export data do not have any fields that are not participating in the mapping to NVA entries and we do not have any fields in the `@JsonIgnoreProperties` field.

### DIFF
--- a/cristin-import/src/test/resources/cristin_entry_of_known_type_with_all_fields.json
+++ b/cristin-import/src/test/resources/cristin_entry_of_known_type_with_all_fields.json
@@ -14,8 +14,8 @@
     "fileUri": "s3://bucket/parent/child/filename.json",
     "contents": {
       "id": 417536,
-      "varbeidhovedkatkode": "TIDSSKRIFTPUBL",
-      "varbeidunderkatkode": "ARTIKKEL",
+      "varbeidhovedkatkode": "BOK",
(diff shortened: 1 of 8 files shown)


cristin-import/src/test/resources/cristin_entry_of_known_type_with_all_fields.json
```diff
@@ -14,8 +14,8 @@
     "fileUri": "s3://bucket/parent/child/filename.json",
     "contents": {
       "id": 417536,
-      "varbeidhovedkatkode": "TIDSSKRIFTPUBL",
-      "varbeidunderkatkode": "ARTIKKEL",
+      "varbeidhovedkatkode": "BOK",
```
axthosarouris, line 17: changed entry so that the valid entry has a known (to the CristinMapper) type

Eirik-Eide-Saur (approved):

PR #443, opened 2021-06-30
NP-2878: update documentation regarding vitenskapeligarbeid_lokal

axthosarouris:


terjehell (approved):

terjehell (dismissed):

terjehell (approved):

brinxmat (dismissed):

docs/cristin.ttl
```diff
@@ -72,6 +72,10 @@
     and they have contributed as "ILLUSTRATOR" affiliated with InstB for the same publication.
     """ .
 
+:VITENSKAPELIGARBEID_LOKAL
```
axthosarouris, line 75: I am not really sure about the meaning of this field.

Antares42: This field contains metadata relevant for NVI reporting, one set per contributing* institution.

So for a given publication/result, let's say there are four authors: One affiliated** with UIB, two affiliated with NTNU and one affiliated with some foreign university.

For the current NVI reporting process, that means that some authority (superuser) from both UIB and NTNU have to verify that this result is worthy of including it in the NVI report, and correctly credited.

The field VITENSKAPELIGARBEID_LOKAL contains the status and verdict for these two institutions. Who verified the process and when, whether they can confirm that some conditions are met, and whether they then regard the result as acceptable for reporting or not.

In terms of data modeling, this information has a different cardinality/multiplicity than the authors or their affiliations, so it is stored in a separate array of objects.

*) only "cristin institutions", i.e. those that participate in the NVIU process, get a record in the VITENSKAPELIGARBEID_LOKAL table
**) technically it's about the unique set of insitutions representing the different and overlapping affiliations that the different authors may have

docs/cristin.ttl
```diff
@@ -116,13 +120,15 @@
 
 :dato_opprettet
     a               :cristinfield ;
-    rdfs:comment    "The date the resource was registered in Cristin." ;
+    rdfs:comment    "When the domain is :Result, it is the date the resource was registered in Cristin. Otherwise UNKNOWN meaning." ;
```
Antares42, line 123: Ohhh, hm, are you attempting to define the field "dato_opprettet" independent of the object in which it is used?

Hm. Hm. In general, there is no guarantee that two fields with the same name but in different tables / objects, have the same meaning.

That being said, a quick survey of fields names across the current json export shows most field names being unique.

And in particular "brukernavn_opprettet", "dato_opprettet" and "eierkode_opprettet" usually mean "who and when created this row in the table / this instance of the object". But the details may differ, like when vitenskapeligarbeid_lokal only has one "eierkode" field that applies to both "opprettet" and "siste_endring" for obvious reasons. :-| 

So, global definitions, bad gut feeling.

docs/cristin.ttl
```diff
@@ -467,14 +482,91 @@
...
+    rdfs:comment    """
         The role of the respective Contributor for the respective publication affiliated for this role with the respective institution.
         This entry is immutable and does not change in time.
         """ ;
+    rdfs:range      xsd:string ;
+    rdfs:domain     :VARBEID_PERSON_STED_ROLLE_Class ;
+    :isignored      "false" ;
+    :mapstonvafield " maps to NVA Role" .
+
+:eierkode
```
Antares42, line 495: Same gut feeling about global definitions. "eierkode" occurs in VITENSKAPELIGARBEID_LOKAL, ARKIVFIL, and ARKIVPOST.

In the first, it distinguishes between the NVI reporting records of the respective involved institutions. In the second, it records which insitution the user who uploaded a file to an archive belongs to. In the third, I'm not sure, I think something about which institution's archive a reference to a file was found in...

Antares42 (commented):
(Eek, var ikke bevisst på forskjellen mellom "comment" og "review")

docs/cristin.ttl
axthosarouris: I was considering creating one instance per object but the documentation becomes very verbose because we have to create weird names to distinguish between the fields. We use the domain the object that the item belongs to and we can elaborate in the comment field for each domain.

docs/cristin.ttl
```diff
@@ -103,22 +112,23 @@
          """ .
 
 :FAGFELT_Class
-    a       :CristinClass;
-    rdfs:comment """The scientific field in which the book is published""" ;
+    a               :CristinClass ;
+    rdfs:comment    """The scientific field in which the book is published""" ;
```
Antares42, line 116: NB: Is also used for single book chapters (BOK_RAPPORT_DEL). The inheritance rules between book and chapter are somewhat arcane.

docs/cristin.ttl
```diff
@@ -103,22 +112,23 @@
...
 :TYPE_TIDSSKRIFTPUBLIKASJON_Class
-    a       :CristinClass;
-    rdfs:comment """A type publication called tidsskrift""".
+    a            :CristinClass ;
+    rdfs:comment """A type publication called tidsskrift""" .
 
 :TIDSSKRIFT_Class
-    a       :CristinClass;
+    a            :CristinClass ;
     rdfs:comment """UNKNOWN""" .
```
Antares42, line 125: Contains information about a journal where a result i published.

NB: For e.g. books that are published as part of a series, the series is also modeled as a TIDSSKRIFT (called TIDSSKRIFT_SERIE in the JSON export, but containing the same fields and stored in the same DB table). I guess this is because both have a name and an ISSN, so they were modeled together.

The JSON at least gives them two different field names, so there's that.

docs/cristin.ttl
```diff
@@ -103,22 +112,23 @@
...
 
 :VARBEID_EMNEORD_Class
-    a       :CristinClass;
+    a            :CristinClass ;
     rdfs:comment """UNKNOWN""" .
```
Antares42, line 129: An array of keywords attached to the publication/result. The keywords have an id, but the list is not actively managed (free to create new ids/keywords)

docs/cristin.ttl
```diff
@@ -163,13 +173,21 @@
 
 :dato_opprettet
     a               :cristinfield ;
-    rdfs:comment    "The date the resource was registered in Cristin." ;
+    rdfs:comment    """
+    When the domain is :Result, it is the date the resource was registered in Cristin or its predecessor.
+    When the domain is :VITENSKAPELIGARBEID_LOKAL, it is the date the resource was registered in NVI.
```
Antares42, line 178: Well, not exactly, unfortunately, because nothing is ever individually registered in NVI.

All data is compiled in the Cristin database, checked and crosschecked, and then a report is generated once a year which is sent to NVI.

The date of the individual record is when that was created in the Cristin database, whether or not it eventually was sent to NVI. It's really basically only a database timestamp.

How about:

When the domain is :VITENSKAPELIGARBEID_LOKAL, it is the date the resource's NVI processing record for the respective institution was created.

docs/cristin.ttl
```diff
@@ -163,13 +173,21 @@
...
+
+    Note: there are approx 200_000 entries with dates prior to 2004 which was the year Cristin
```
Antares42, line 180: Dangling sentence ;-)

docs/cristin.ttl
```diff
@@ -197,7 +215,7 @@
     """ .
 
 :eierkode_opprettet
-    a            rdf:Property ;
+    a            :cristinfield ;
     rdfs:comment """
         eierkode_opprettet and brukernavn are the primary key for idenitfying which user created the entry. As a general rule a physical person
```
Antares42, line 220: Since it specifically says eierkode_opprettet, it should probably also say brukernavn_opprettet

docs/cristin.ttl
```diff
@@ -208,42 +226,50 @@
 
 
 :brukernavn_opprettet
-    a            rdf:Property ;
+    a            :cristinfield ;
     rdfs:comment """
-       eierkode_opprettet and brukernavn are the primary key for idenitfying which user created the entry. As a general rule a physical person
+        When the domain is :Result, eierkode_opprettet and brukernavn are the primary key for idenitfying which user created the entry. As a general rule a physical person
```
Antares42, line 231: same here, brukernavn_opprettet

docs/cristin.ttl
```diff
@@ -208,42 +226,50 @@
...
         can have many usernames (brukernavn) but only one eierkode. This holds even if a physical person moves from institution to institution
         and/or changes roles. Duplicates that violate this assumption are possible to occur but they are not expected to be the rule.
+
+        When it is :VITENSKAPELIGARBEID_LOKAL_Class it is the user who created the NVI entry."
+        This entails that the semantics of this term are unclear, possibly even polysemous.
```
Antares42, line 236: Well, as discussed, the semantics include the containing object, not the JSON fieldname alone. "brukernavn_opprettet" is whoever was logged in to create that object, be it the result itself or some of its attached objects.

docs/cristin.ttl
```diff
@@ -260,39 +286,39 @@
...
+    a            :cristinfield ;
     rdfs:comment "Free text for super-users to add comments when it comes to NVI approvals." ;
     rdfs:range   xsd:string ;
     rdfs:domain  :Result ;
     :isignored   "true" .
 
 :merknadtekst_godkjenning
-    a            rdf:Property ;
+    a            :cristinfield ;
     rdfs:comment "Free text fore super-users to add explanations/details on why the publication was approved or rejected." ;
```
Antares42, line 304: fore -> for

would also extend to "...approved or rejected for NVI reporting."

Because this field is the final verdict, while "merknadtekst" above is the scrap of notes that super-users can write to each other while working on their decision.

docs/cristin.ttl
```diff
@@ -260,39 +286,39 @@
...
     :isignored   "true" .
 
 :kildekode
-    a            rdf:Property ;
+    a            :cristinfield ;
     rdfs:comment """
     The source (provider) the result was imported from, if any.
     Mostly superseded by the list of import references in table VARBEID_KILDE
-    It is a string containing arbitrary codes such as "scopus" "norart" etc.
+    It is a string containing arbitrary codes such as "scopus" "norart", "BIBSYS" etc.
```
Antares42, line 314: NB: This, together with kildepostid, is something we have to check more closely, because we also have a list/array VARBEID_KILDE which can include more than source, say when two records from different sources were merged into one Cristin result.

PS: Ah, I see you noted something along these lines on "kildepostid". So yeah, maybe enough to note that one should rather look at VARBEID_KILDE

docs/cristin.ttl
```diff
@@ -473,21 +501,21 @@
     :isignored   "true" .
 
 :institusjonsnavn_opprinnelig
-    a            rdf:Property ;
+    a            :cristinfield ;
     rdfs:comment "The original name of the institution at the time the Cristin entry was created. Immutable value" ;
     rdfs:range   xsd:string ;
     rdfs:domain  :VARBEID_PERSON_STED_Class ;
     :isignored   "true" .
```
Antares42, line 508: I see many of these as "ignored", and can agree wrp/ dereferencing and such, but doesn't NVA in its archive role have use for knowing that an article was published crediting Rikshospitalet, even though Rikshospitalet later became a part of Oslo University Hospital?

Not complaining, just asking. :-)

docs/cristin.ttl
```diff
@@ -514,14 +542,98 @@
...
+    :isignored   "true" .
+
+
+:status_kontrollert
+    a            :cristinfield ;
+    rdfs:range   xsd:string ;
+    rdfs:domain  :VITENSKAPELIGARBEID_LOKAL_Class ;
+    rdfs:comment """
+     It indicates if the respective post has been controlled or not. It receives values "J", "N".
+     It is unknown whether it refers to the Cristin entry or the NVI entry.
```
Antares42, line 570: It is part of the NVI entry, so it refers to the respective institution's opinion on whether the Cristin entry should be reported to NVI and count towards RBK funding.

NULL = this institution has not yet given their opinion
J = this institution believes the current result should be included in the NVI report
N = this institution believes the current result should NOT be included in the NVI report (e.g. doesn't meet academic standards, or has other issues)

In the period up to generating the yearly report to NVI, the relevant institutions should ping-pong until they agree on J or N.

docs/cristin.ttl
```diff
@@ -514,14 +542,98 @@
...
+    """ ;
+    :isignored   "true" .
+
+:brukernavn_kontrollert
+    a            :cristinfield ;
+    rdfs:range   xsd:string ;
+    rdfs:domain  :VITENSKAPELIGARBEID_LOKAL_Class ;
+    rdfs:comment """
+     The username who controlled the respective entry. Unknown whether the field refers to the NVI entry or the Cristin entry.
```
Antares42, line 579: The username who controlled / judged / validated the respective entry *representing their respective institution*, distinguished by VITENSKAPELIGARBEID_LOKAL.eierkode.

Is contained in the VITENSKAPELIGARBEID_LOKAL object, so refers to the NVI processing record.

docs/cristin.ttl
```diff
@@ -514,14 +542,98 @@
...
+     Unknown whether the user belongs to NSD or to the institution that registered the entry (e.g. NTNU)
```
Antares42, line 580: User belongs to the institution given by the "eierkode" in the same record. As always, brukernavn+eierkode serve as a unique identifier / composite foreign key to Cristin's BRUKER table.

docs/cristin.ttl
```diff
@@ -514,14 +542,98 @@
...
+
+    """ ;
+    :isignored   "true" .
+
+:dato_kontrollert
+    a            :cristinfield ;
+    rdfs:range   xsd:string ;
+    rdfs:domain  :VITENSKAPELIGARBEID_LOKAL_Class ;
+    rdfs:comment """
+     The date the respective entry was controlled. Unknown whether the field refers to the NVI entry or the Cristin entry.
```
Antares42, line 590: Same. Domain is the NVI processing record for one institution, date refers to when a superuser of said institution did their review of the Cristin entry / result / publication.

docs/cristin.ttl
```diff
@@ -514,14 +542,98 @@
...
+
+    """ ;
+    :isignored   "true" .
+
+:status_bekreftet_arkivsporsmal
+    a            :cristinfield ;
+    rdfs:range   xsd:string ;
+    rdfs:domain  :VITENSKAPELIGARBEID_LOKAL_Class ;
+    rdfs:comment """UNKNOWN meaning""" ;
```
Antares42, line 599: (get to these next time)

docs/cristin.ttl
```diff
@@ -734,11 +845,11 @@
     """ .
 
 :emneordlopenr
-    a               :cristinfield ;
-    rdfs:comment    "UNKNOWN" ;
-    rdfs:range      xsd:integer ;
-    rdfs:domain     :VARBEID_EMNEORD_Class ;
-    :isignored      "true" .
+    a            :cristinfield ;
+    rdfs:comment "UNKNOWN" ;
```
Antares42, line 849: It's the id of a given keyword record in Cristin's uncurated keyword list.

Antares42 (commented):
Fortsetter neste uke, god helg!

docs/cristin.ttl
axthosarouris: :isignored shows the sitatuation as it is now. My goal is to document what is the meaning of the each field and what we do with this at present. When the product owener, project manager, read it, they will decide whether we need this information or not.

PR #447, opened 2021-07-02
Np 2867 implement mappings from cristin to nva

Eirik-Eide-Saur:
Tests for error-reports has not been implemented and the testdata is not generic. This is a preliminary commit


cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinBookReport.java
```diff
@@ -0,0 +1,35 @@
...
+    private String isbn;
+    @JsonProperty(PUBLISHER)
+    private String publisherName;
+    @JsonProperty(NUMBER_OF_PAGES)
+    private String numberOfPages;
+
+    public CristinBookReport() {
+
+    }
+
```
axthosarouris, line 34: add a copy method that returns a builder. see CrstinObject.

Eirik-Eide-Saur: Done

cristin-import/src/test/java/no/unit/nva/cristin/CristinDataGenerator.java
```diff
@@ -191,10 +194,21 @@ private CristinObject createObjectWithCristinContributorRoleCode(Integer index,
...
                 .withPublicationYear(randomYear())
                 .withContributors(contributors)
+                .withBookReport(randomBookReport())
                 .withPublicationOwner(HardcodedValues.HARDCODED_PUBLICATIONS_OWNER)
                 .build();
     }
 
+    private List<CristinBookReport> randomBookReport() {
+        CristinBookReport bookReport = new CristinBookReport();
+        bookReport.setIsbn("9788770342827");
```
axthosarouris, line 204: use 
```
private String randomIsbn() {
        return FAKER.code().isbn10();
    }

```

axthosarouris (changes requested):

cristin-import/src/test/java/no/unit/nva/cristin/CristinDataGenerator.java
```diff
@@ -191,10 +196,25 @@ private CristinObject createObjectWithCristinContributorRoleCode(Integer index,
                 .withEntryCreationDate(LocalDate.now())
                 .withPublicationYear(randomYear())
                 .withContributors(contributors)
+                .withBookReport(randomBookReport())
                 .withPublicationOwner(HardcodedValues.HARDCODED_PUBLICATIONS_OWNER)
                 .build();
     }
 
+    private List<CristinBookReport> randomBookReport() {
+        CristinBookReport bookReport = new CristinBookReport().copy().build();
```
axthosarouris, line 205: Why are you creating an object and then you copy it. The way to use the builder is the following:
```
CristinBookReport bookReport = new CristinBookReport().copy().build();
        CristinBookReport.builder()
            .withIsbn(randomIsbn13())
            .withPublisherName(randomString())
            .withNumberOfPages(randomString());
        return List.of(bookReport);    
```

cristin-import/src/test/java/no/unit/nva/cristin/CristinDataGenerator.java
```diff
@@ -191,10 +196,25 @@ private CristinObject createObjectWithCristinContributorRoleCode(Integer index,
...
+        bookReport.setIsbn(randomIsbn13());
+        bookReport.setPublisherName(randomString());
+        bookReport.setNumberOfPages(randomString());
+        List bookReportList = new ArrayList();
```
axthosarouris, line 209: Although you can use List.of(...) here, in general try not use generic classes (e.g List<...> ) without specifying the type. Try to always specify the type because when you will be comparing two lists they will not be equal if one is List<Object> and the other List<SpecificType> even if they contain the same objects.

axthosarouris (changes requested):

cristin-import/src/test/java/no/unit/nva/cristin/lambda/CristinEntryEventConsumerTest.java
```diff
@@ -455,4 +550,13 @@ private CristinObject generatePublicationFromResource(String input) throws JsonP
         assert nonNull(cristinObject.getId()); //java assertion produces Error not exception
         return cristinObject;
     }
+
+    private boolean containsStrings(String sourceString, List<String> listOfStrings) {
```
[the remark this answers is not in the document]
Eirik-Eide-Saur: This was done because I was unable to find a hamcrest method that had this functionality

cristin-import/src/test/java/no/unit/nva/cristin/lambda/CristinEntryEventConsumerTest.java
```diff
@@ -324,13 +332,92 @@ public void handleRequestThrowsExceptionWhenInputContainsUnknownProperty() throw
         assertThat(file, containsString(UNKNOWN_PROPERTY_NAME_IN_RESOURCE_FILE_WITH_UNKNOWN_PROPERTY));
     }
 
+    @Test
+    public void handlerThrowsExceptionWhenIsbnValueIsNull() {
+        CristinObject cristinInput = cristinDataGenerator.randomObject();
```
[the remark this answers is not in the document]
Eirik-Eide-Saur: I created a separate method creating a random object of type Monograph with bookreport values. But I do run into problems if I dont include the ".withBookReport()" when building the other versions of CristinObjects. This is because it fails the check assertThat(cristinObject, doesNotHaveEmptyValues()); which is included at the end of all the methods.

cristin-import/src/test/java/no/unit/nva/cristin/lambda/CristinEntryEventConsumerTest.java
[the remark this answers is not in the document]
Eirik-Eide-Saur: It is now removed

axthosarouris (approved):

PR #457, opened 2021-07-07
Add test for missing pages in a book report

Eirik-Eide-Saur:
Added test for just the pages value, according to Kari the other 2 fields can be null event though it is a TYPE_BOK_RAPPORT

### DIFF
--- a/cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinMapper.java
+++ b/cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinMapper.java
@@ -43,6 +43,7 @@ import no.unit.nva.model.pages.MonographPages;
 import no.unit.nva.model.pages.Pages;
 import no.unit.nva.model.pages.Range;
 import nva.commons.core.SingletonCollector;
+import nva.commons.core.StringUtils;
 import nva.commons.core.attempt.Try;
 import nva.commons.core.language.LanguageMapper;
 
@@ -53,8 +54,6 @@ public class CristinMapper {
     public static final String ERROR_PARSING_MAIN_CATEGORY = "Error parsing main category";
     public static final String ERROR_PARSING_MAIN_OR_SECONDARY_CATEGORIES = "Error parsing main or secondary "
                                                                             + "categories";
-    public static final String ERROR_MISSING_NUMBER_OF_PAGES_IN_BOOK_REPORT = "Number of pages in a Cristin "
-            + "entry of type Book report can not be null";
 
     private final CristinObject cristinObject;
 
@@ -222,8 +221,8 @@ public class CristinMapper {
 
     private String extractNumberOfPages() {
         String numberOfPages = extractCristinBookReport().getNumberOfPages();
-        if (numberOfPages == null) {
-            throw new MissingFieldsException(ERROR_MISSING_NUMBER_OF_PAGES_IN_BOOK_REPORT);
+        if (StringUtils.isBlank(numberOfPages)) {
+            throw new InvalidCristinBookReportEntryException("numberOfPages", numberOfPages);
         }
         return numberOfPages;
     }
--- a/cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinMapper.java
+++ b/cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinMapper.java
@@ -53,6 +53,8 @@ public class CristinMapper {
     public static final String ERROR_PARSING_MAIN_CATEGORY = "Error parsing main category";
     public static final String ERROR_PARSING_MAIN_OR_SECONDARY_CATEGORIES = "Error parsing main or secondary "
                                                                             + "categories";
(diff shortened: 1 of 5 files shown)


cristin-import/src/test/resources/features/BookMonographRules.feature
```diff
@@ -38,5 +38,7 @@ Feature:
     Then the NVA Resource has a PublicationContext with publisher equal to "House of Publishing"
 
 
-
-
+  Scenario: Map throws exception when Cristin with type book report has no number of pages value.
```
axthosarouris, line 41: Mapping fails when a Cristin Result of type Monografi has no information about the number of pages

cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinMapper.java
```diff
@@ -219,7 +221,11 @@ private CristinBookReport extractCristinBookReport() {
     }
 
     private String extractNumberOfPages() {
-        return extractCristinBookReport().getNumberOfPages();
+        String numberOfPages = extractCristinBookReport().getNumberOfPages();
+        if (numberOfPages == null) {
+            throw new MissingFieldsException(ERROR_MISSING_NUMBER_OF_PAGES_IN_BOOK_REPORT);
```
axthosarouris, line 226: I think this is an InvalidCristinBookEntry exception . MissingFieldsException is only for fields that we should fill in but we are not filling in for some reason that we do not know. The InvalidCristinBookEntry exception does not exist yet, it should be a RuntimeException.

cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinMapper.java
```diff
@@ -219,7 +221,11 @@ private CristinBookReport extractCristinBookReport() {
     }
 
     private String extractNumberOfPages() {
-        return extractCristinBookReport().getNumberOfPages();
+        String numberOfPages = extractCristinBookReport().getNumberOfPages();
+        if (numberOfPages == null) {
```
axthosarouris, line 225: StringUtils.isBlank(numberOfPages). Also in general the style is to use the  isNull(), notNull()  methods

cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinMapper.java
```diff
@@ -53,6 +53,8 @@
     public static final String ERROR_PARSING_MAIN_CATEGORY = "Error parsing main category";
     public static final String ERROR_PARSING_MAIN_OR_SECONDARY_CATEGORIES = "Error parsing main or secondary "
                                                                             + "categories";
+    public static final String ERROR_MISSING_NUMBER_OF_PAGES_IN_BOOK_REPORT = "Number of pages in a Cristin "
```
axthosarouris, line 56: This is only for Books, not for any Cristin entry.

cristin-import/src/test/resources/features/BookMonographRules.feature
```diff
@@ -38,5 +38,7 @@ Feature:
...
+    Given that the Book Report entry has an empty numberOfPages field
```
axthosarouris, line 42: Given that the Book Report entry has an empty "numberOfPages" field

axthosarouris (changes requested):

cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinMapper.java
```diff
@@ -219,7 +220,11 @@ private CristinBookReport extractCristinBookReport() {
     }
 
     private String extractNumberOfPages() {
-        return extractCristinBookReport().getNumberOfPages();
+        String numberOfPages = extractCristinBookReport().getNumberOfPages();
+        if (StringUtils.isBlank(numberOfPages)) {
+            throw new InvalidCristinBookReportEntryException("numberOfPages", numberOfPages);
+        }
+        return numberOfPages;
```
axthosarouris, line 227: I think there is a problem here that when this error will occur the JSON fields are not actually  called BookReport and numberOfPages. I would suggest moving this check in the getter of the object. 
i.e.,
```
   public String getNumberOfPages() {
        if (Objects.nonNull(numberOfPages)) {
            return numberOfPages;
        }

        throw new InvalidCristinBookReportEntryException(NUMBER_OF_PAGES, numberOfPages);
    }
```

In this way return at least the correct name of the field and if the name of the field changes then the error message will change automatically

axthosarouris: In general if you have something like this in the code:
```
foo(){
var x= this.object.getX();
var y = this.object.getY();
var z = this.object.processXandY(x,y);
this.object.doodle(x)
this.object.booble(y)
```
This is an indication that the whole method belongs to the class of the obect

axthosarouris (changes requested):

axthosarouris (approved):

PR #458, opened 2021-07-08
Custom resource to create deployment in api gateway

janbj:


template.yaml
```diff
@@ -1297,6 +1297,59 @@ Resources:
           EVENT_BUS: !GetAtt ImportRequestsEventBus.Name
           IMPORT_EVENT_TYPE: import.cristin.entry-event
 
+  #============================ Deploy API =============================================================
```
axthosarouris, line 1300: Add a comment why is this necessary. It can be as simple as "This  solves the problem described here: http://....."

axthosarouris (changes requested):

svenngjeto (dismissed):

template.yaml
```diff
@@ -1297,6 +1297,74 @@ Resources:
...
+
+  ApiGatewayCreateDeploymentCustomResource:
+    DependsOn: NvaPublicationApi
+    Type: AWS::CloudFormation::CustomResource
+    Properties:
+      ServiceToken: !GetAtt ApiGatewayCreateDeploymentLambda.Arn
+      RestApiId: !Ref NvaPublicationApi
+      StageName: !Ref NvaPublicationApi.Stage
+      Timestamp: '${BUILD_TIMESTAMP}'
+      AuthorizerArn: !Ref CognitoAuthorizerArn
```
janbj, line 1367: This will force update when authorizer ARN changes. Do we need both?

template.yaml
```diff
@@ -1297,6 +1297,74 @@ Resources:
...
+            cfnresponse.send(event, context, responseStatus, responseData)
+
+  ApiGatewayCreateDeploymentCustomResource:
+    DependsOn: NvaPublicationApi
+    Type: AWS::CloudFormation::CustomResource
+    Properties:
+      ServiceToken: !GetAtt ApiGatewayCreateDeploymentLambda.Arn
+      RestApiId: !Ref NvaPublicationApi
+      StageName: !Ref NvaPublicationApi.Stage
+      Timestamp: '${BUILD_TIMESTAMP}'
```
janbj, line 1366: This will force update on every build.

template.yaml
axthosarouris: Keep  it for now.

axthosarouris (approved):

template.yaml
svenngjeto: It will cover the rare situation when we create a completely new Cognito without rebuilding publication-api?

svenngjeto (approved):

PR #460, opened 2021-07-08
Map cristin artikkel fag to nva

Eirik-Eide-Saur:


cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinMapper.java
```diff
@@ -138,6 +147,20 @@ private PublicationContext buildPublicationContext() throws InvalidIsbnException
                        .withOpenAccess(false)
                        .build();
         }
+        if (isJournal()) {
+            List<String> isbnList = new ArrayList<>();
+            isbnList.add(extractIsbn());
+            URL theUrl = new URL("https://www.example.com/");
+            return new Journal.Builder()
+                    .withLevel(HARDCODED_LEVEL)
```
axthosarouris, line 155: HARDCODED_JOURNAL_LEVEL

cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinMapper.java
```diff
@@ -138,6 +147,20 @@ private PublicationContext buildPublicationContext() throws InvalidIsbnException
...
+                    .withPeerReviewed(HARDCODED_PEER_REVIEWED)
```
axthosarouris, line 156: HARDCODED_JOURNAL_PEER_REVIEWED

cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinMapper.java
Eirik-Eide-Saur: Why cant they be re-used from the Book version?

cristin-import/src/main/resources/ignoredFields.txt
```diff
@@ -7,6 +7,7 @@ modifiedDate
 projects
 publishedDate
 doiRequest
+.owner
```
axthosarouris, line 10: owner cannot be empty. it is part of the primary key. inject a value when you create the test data

axthosarouris: There is a method in CirstinObject called hardcodePublicationOwner

axthosarouris (changes requested):

axthosarouris (approved):

terjehell (approved):

PR #481, opened 2021-07-26
NP-2926 Map ARTIKKEL, ARTIKKEL_POP og OVERSIKTSART to JOURNAL_ARTICLE

Eirik-Eide-Saur:


Eirik-Eide-Saur:
CHAPTERACADEMIC AND LEXICALIMPORT where both skipped as they are of type CHAPTER. CHAPTERACADEMIC is already mapped.

cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinSecondaryCategory.java
```diff
@@ -10,7 +10,7 @@
 public enum CristinSecondaryCategory {
     ANTHOLOGY("ANTOLOGI", "ANTHOLOGY"),
     MONOGRAPH("MONOGRAFI", "MONOGRAPH"),
-    JOURNAL_ARTICLE("ARTIKKEL_FAG", "JOURNAL_ARTICLE"),
+    JOURNAL_ARTICLE("ARTIKKEL_FAG", "JOURNAL_ARTICLE", "ARTIKKEL", "ARTIKKEL_POP", "OVERSIKTSART"),
```
axthosarouris, line 13: This is our internal mapping of the Cristin categories. I think that we should not compress the cristin categories and do in practice the mapping to the NVA catogory here. It is better to create more categories with names based loosely on the cristin english names and perform the mapping in the CristinMapper, i.e., extend the `isJournal()` method.

cristin-import/src/test/resources/features/JournalArticleRules.feature
```diff
@@ -7,6 +7,21 @@ Feature: Mapping of Journal Article
     When the Cristin Result is converted to an NVA Resource
     Then the NVA Resource is an instance of "JournalArticle"
 
+  Scenario: Map returns a Journal Article NVA Resource when the Cristin Result is an "ARTICLE"
```
axthosarouris, line 10: An "Academic Article" maps to a Journal Article. I think it is better if in the scenario titles we use the english name of the cristin category as it appears in the excel file (column VARBEIDUNDERKATNAVN_ENGELSK).

cristin-import/src/test/resources/features/JournalArticleRules.feature
```diff
@@ -7,6 +7,21 @@ Feature: Mapping of Journal Article
...
+    Given a valid Cristin Result with secondary category "ARTIKKEL"
```
axthosarouris, line 11: ditto

cristin-import/src/test/resources/features/JournalArticleRules.feature
```diff
@@ -7,6 +7,21 @@ Feature: Mapping of Journal Article
...
+    When the Cristin Result is converted to an NVA Resource
+    Then the NVA Resource is an instance of "JournalArticle"
+
+  Scenario: Map returns a Journal Article NVA Resource when the Cristin Result is an "POPULARARTICLE"
```
axthosarouris, line 15: ditto

cristin-import/src/test/resources/features/JournalArticleRules.feature
```diff
@@ -7,6 +7,21 @@ Feature: Mapping of Journal Article
...
+    Given a valid Cristin Result with secondary category "ARTIKKEL_POP"
+    When the Cristin Result is converted to an NVA Resource
+    Then the NVA Resource is an instance of "JournalArticle"
+
+  Scenario: Map returns a Journal Article NVA Resource when the Cristin Result is an "ACADEMICREVIEW"
```
axthosarouris, line 20: ditto

axthosarouris (changes requested):

axthosarouris (changes requested):

PR #482, opened 2021-08-04
NP-2741 Map Bokanmeldelse to JournalReview

Eirik-Eide-Saur:


cristin-import/src/test/resources/features/JournalReviewRules.feature
```diff
@@ -0,0 +1,9 @@
+Feature:
+
+  Background:
+    Given a valid Cristin Result with secondary category "BOKANMELDELSE"
+
+  Scenario: Map returns NVA Resource with Reference having a PublicationInstance of type
```
axthosarouris, line 6: Cristin Result of type Book Review maps to NVA JournalReview

axthosarouris (approved):

PR #483, opened 2021-08-04
Rewrite scenarios for BookMonograph and BookAnthology

axthosarouris:
Update scenarios for BookMonograph and BookAnthology for scenarios to include both types and also for scenarios to be more understandable for people that are not involved in the details of the mapping process (code)


cristin-import/src/test/java/cucumber/BookFeatures.java
```diff
@@ -58,10 +57,10 @@ public void theBookReportHasAEqualTo(String numberOfPages) {
...
     public void theNvaResourceHasAPublicationContextWithNumberOfPagesEqualTo(String expectedNumberOfPages) {
         PublicationInstance<?> context = scenarioContext.getNvaEntry()
-                                             .getEntityDescription()
-                                             .getReference()
-                                             .getPublicationInstance();
-        BookMonograph book = (BookMonograph) context;
+            .getEntityDescription()
+            .getReference()
+            .getPublicationInstance();
+        PeerReviewedMonograph book = (PeerReviewedMonograph) context;
```
axthosarouris, line 63: Necessary whey checking for both BookMonograph and BookAnthology

cristin-import/src/test/resources/features/BookFeatures.feature
```diff
@@ -0,0 +1,102 @@
+Feature: Book conversion features
```
axthosarouris, line 1: All rules from the BookAnthology and the BookMonograph were moved here. Rules that applied to both types were adapted in order for both types to be checked

Eirik-Eide-Saur: Should this file be called BookRules to match the rest of the feature files?

cristin-import/src/test/resources/features/BookFeatures.feature
```diff
@@ -0,0 +1,102 @@
...
+    When the Cristin Result is converted to an NVA Resource
+    Then the NVA Resource has a PublicationContext with publisher equal to "House of Publishing"
+    Examples:
+      | secondaryCategory |
+      | MONOGRAFI         |
+      | ANTOLOGI          |
+
+
+  Scenario Outline: NPI subject heading is copied from Cristin Result as is.
+  the Cristin Entry's Book report subjectField "subjectFieldCode" entry
```
Eirik-Eide-Saur, line 69: Should this be something along the lines of "The value is stored in the Cristin Entry's Book report subjectField "subjectFieldCode" entry"?

axthosarouris: I think I will remove the title completely.

Eirik-Eide-Saur (approved):

PR #491, opened 2021-08-09
NP-2962 Map Doktorgradavhandling to DegreePhd

Eirik-Eide-Saur:


cristin-import/src/test/resources/features/DegreePhdRules.feature
```diff
@@ -0,0 +1,9 @@
+Feature:
+  Mapping rules for Degree_phds
+
+  Background:
+    Given a valid Cristin Result with secondary category "DRGRADAVH"
+
+  Scenario: Cristin Result with type "Drgradavh" is converted to an NVA entry with PublicationInstance of type "DegreePhd".
```
axthosarouris, line 7: Use the file in Confluence to file the english name so that it is readable for Terje. 
E.g. Cristin Result  "Ph.D thesis" is converted to an NVA entry with type "DegreePhd"

docs/cristin.ttl
```diff
@@ -193,7 +193,10 @@
          Anthology -> BookAnthology
          Monograph -> BookMonograph
          Rapport -> ReportResearch
-         Kapittel -> ChapterArticle
+         Kapittel, Popvit_Kapittel, Faglig_Kapittel-> ChapterArticle
```
axthosarouris, line 196: break it down to 3 lines

axthosarouris (changes requested):

cristin-import/src/test/resources/features/DegreePhdRules.feature
```diff
@@ -0,0 +1,9 @@
+Feature:
+  Mapping rules for DegreePhds
+
+  Background:
+    Given a valid Cristin Result with secondary category "DRGRADAVH"
+
+  Scenario: Cristin Result "Ph.D thesis" is converted to an NVA entry with type "DegreePhd".
```
axthosarouris, line 7: Use the description found in the respective file in Confluence. "Ph.D thesis" was an example.   We want to use the file so that there is no ambiguity in our communication with Terje. Although, one can say that we Doctoral thesis or Dissertation or Ph.D. thesis are synonymous, just habits create ambiguity in the code and simple things like searching for conversion rules of a specific type become difficult.

axthosarouris (changes requested):

axthosarouris (approved):

PR #504, opened 2021-08-16
NP-2993 Add test and mapping for journalLetter.

Eirik-Eide-Saur:


cristin-import/src/test/java/no/unit/nva/cristin/CristinDataGenerator.java
```diff
@@ -124,6 +124,8 @@ public static CristinObject randomObject(String secondaryCategory) {
                 return randomBookAnthology();
             case FEATURE_ARTICLE:
                 return randomFeatureArticle();
+            case JOURNAL_LETTER:
+                return randomJournalLetter();
             case JOURNAL_ARTICLE:
```
axthosarouris, line 129: Not a comment for this PR, but we have 4 secondary categories where the user is asking for a specific category and we return JOURNAL_ARTICLE. I think this can be problematic in the future. 
I am referring to the part 
```
 case JOURNAL_ARTICLE:
            case ARTICLE:
            case POPULAR_ARTICLE:
            case ACADEMIC_REVIEW:
```

axthosarouris (approved):

PR #508, opened 2021-08-16
NP-2996 Add mapping and test for JournalCorrigendum.

Eirik-Eide-Saur:


cristin-import/src/main/java/no/unit/nva/cristin/lambda/constants/HardcodedValues.java
```diff
@@ -21,8 +21,6 @@
     //JOURNAL
     public static final Level HARDCODED_JOURNAL_LEVEL = Level.LEVEL_0;
     public static final URI HARDCODED_JOURNAL_URI = URI.create("https://www.journal-example.com/");
-    public static final String HARDCODED_JOURNAL_PAGE = "1";
```
axthosarouris, line 24: Why are we removing these fields? Why is this change injected into a seemingly unrelated PR?

Eirik-Eide-Saur: Terje asked to have them removed, and it seemed somewhat fitting to remove them here as is has to do with the mapping of Journals

axthosarouris (approved):

PR #511, opened 2021-08-20
NP-3004 MAP and Test Content types for Journal articles

Eirik-Eide-Saur:


cristin-import/src/main/java/no/unit/nva/cristin/mapper/ContentType.java
```diff
@@ -0,0 +1,37 @@
+package no.unit.nva.cristin.mapper;
+
+import nva.commons.core.SingletonCollector;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.Objects;
+
+public enum ContentType {
```
axthosarouris, line 9: This is a misuse of an enum. You are using an enum instead for a HashMap
The mistake here is the following.
You have the enum SecondaryCategory and you copy paste the aliases to the ContentType enum to create an implicit link between them. This is problematic for many reasons:
1. You are copying pasting string literals and you apply implicit functionality to them, so they are now more than string literals, they are some kind of variables.
2. The enum aliases represent some kind of mapping and this is not what the enums are supposed to do.
3. The logic of the mapping is now dependent on the string literals and if one string literal changes it is very difficult to trace the changes. Basically a compilation error has been turned into a runtime error.
4. The mapping logic now lies in two places. Implicitly in the enum and partially in the lines 
```
  ContentType contentTypeEnum = ContentType.fromString(secondaryCategory.getValue());
        JournalArticleContentType contentType = JournalArticleContentType
                .lookup(contentTypeEnum.retrieveContentTypeValue());
```
My suggestion is the following code in the CristinSecondaryCategory class
```
public static final Map<CristinSecondaryCategory, JournalArticleContentType> mapToJournalContentType =
        createMapToJournalContentType();

    public JournalArticleContentType toJournalArticleContentType() {
        if (mapToJournalContentType.containsKey(this)) {
            return mapToJournalContentType.get(this);
        } else {
            throw new IllegalStateException(conversionError(this, JournalArticleContentType.class));
        }
    }

    private static Map<CristinSecondaryCategory, JournalArticleContentType> createMapToJournalContentType() {
        return Map.of(JOURNAL_ARTICLE, JournalArticleContentType.PROFESSIONAL_ARTICLE,
                      POPULAR_ARTICLE, JournalArticleContentType.POPULAR_SCIENCE_ARTICLE,
                      ARTICLE, JournalArticleContentType.RESEARCH_ARTICLE,
                      ACADEMIC_REVIEW, JournalArticleContentType.REVIEW_ARTICLE);
    }

```

then code in the CristinMapper becomes:
```
 return new JournalArticle.Builder()
                   .withContent(secondaryCategory.toJournalArticleContentType())
                   .withPages(numberOfPages)
                   .withPeerReviewed(HARDCODED_JOURNAL_PEER_REVIEWED)
                   .withVolume(extractVolume())
                   .build();
```

axthosarouris (changes requested):

cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinMapper.java
```diff
@@ -253,7 +254,7 @@ private Chapter buildChapterForPublicationContext() {
         } else if (isJournal(cristinObject) && isJournalCorrigendum(cristinObject)) {
             return createJournalCorrigendum();
         } else if (isJournal(cristinObject) && isJournalArticle(cristinObject)) {
-            return createJournalArticle();
+            return createJournalArticle(cristinObject.getSecondaryCategory());
```
axthosarouris, line 257: cristinObject is a field in the class. You do not need to pass it as a parameter

axthosarouris (changes requested):

axthosarouris (approved):

PR #515, opened 2021-08-24
NP-3036 Separate the PublicationInstance building into a class

Eirik-Eide-Saur:


cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinMapper.java
```diff
@@ -385,16 +229,12 @@ private String extractMainTitle() {
                    .flatMap(Collection::stream);
     }
 
-    private CristinBookOrReportMetadata extractCristinBookReport() {
+    public CristinBookOrReportMetadata extractCristinBookReport() {
```
Eirik-Eide-Saur, line 232: This and the method extractCristinJournalPublication method are made public as they need to be reach from the new class PublicationInstanceBuilder

cristin-import/src/main/java/no/unit/nva/cristin/mapper/PublicationInstanceBuilder.java
```diff
@@ -0,0 +1,198 @@
...
+    public static final String ERROR_PARSING_MAIN_CATEGORY = "Error parsing main category";
+    public static final String ERROR_PARSING_MAIN_OR_SECONDARY_CATEGORIES = "Error parsing main or secondary "
+            + "categories";
+
+    private final CristinObject cristinObject;
+    private final CristinMapper mapper;
+
+    public PublicationInstanceBuilder(CristinObject cristinObject) {
+        this.cristinObject = cristinObject;
+        this.mapper = new CristinMapper(cristinObject);
```
Eirik-Eide-Saur, line 52: Not a big fan of this solution. I do believe all the "extractData" methods should be moved into its own class, but for now this gives access to the methods needed both in this and the CristinMapper class.

cristin-import/src/main/java/no/unit/nva/cristin/mapper/PublicationInstanceBuilder.java
```diff
@@ -0,0 +1,199 @@
...
+                .withPages(createMonographPages())
+                .build();
+    }
+
+    private PublicationInstance<? extends Pages> createChapterArticle() {
+        return new ChapterArticle.Builder().build();
+    }
+
+    private String extractPagesBegin() {
+        return mapper.extractCristinJournalPublication().getPagesBegin();
```
axthosarouris, line 184: A Mapper has a PublicationInstanceBuilder which has a Mapper for just getting a field from the CristinObect?

axthosarouris: The only reason you are using this method is to handle the case where the cristinObject is null. Require that in the constructor and simplify your code

axthosarouris (changes requested):

cristin-import/src/main/java/no/unit/nva/cristin/mapper/PublicationInstanceBuilder.java
```diff
@@ -0,0 +1,200 @@
...
+    public static final String ERROR_CRISTIN_OBJECT_IS_NULL = "CristinObject can not be null";
+    public static final String ERROR_PARSING_SECONDARY_CATEGORY = "Error parsing secondary category";
+    public static final String ERROR_PARSING_MAIN_CATEGORY = "Error parsing main category";
+    public static final String ERROR_PARSING_MAIN_OR_SECONDARY_CATEGORIES = "Error parsing main or secondary "
+            + "categories";
+
+    private final CristinObject cristinObject;
+
+    public PublicationInstanceBuilder(CristinObject cristinObject) {
+        Validate.notNull(cristinObject, ERROR_CRISTIN_OBJECT_IS_NULL);
```
axthosarouris, line 52: better use Objects.requireNonNull that is native Java. Also the class you are using is not supposed to be used by end-users of the library.

cristin-import/src/main/java/no/unit/nva/cristin/mapper/PublicationInstanceBuilder.java
```diff
@@ -0,0 +1,200 @@
...
+        this.cristinObject = cristinObject;
+    }
+
+    @SuppressWarnings("PMD.CognitiveComplexity")
```
axthosarouris, line 56: If quality check fail, it is an indication that things should be refactored even more, I have created a PR with a suggestion on breaking up this class even more.  Here is the link: https://github.com/BIBSYSDEV/nva-publication-api/pull/516 .

The suggestion is not finished, it has some replication in the exception handling but shows the general direction. If you like my suggestion you can merge my branch to yours and polish it where necessary.

cristin-import/src/test/java/no/unit/nva/cristin/lambda/CristinEntryEventConsumerTest.java
```diff
@@ -261,7 +262,7 @@ public void handlerThrowsExceptionWhenSecondaryCategoryTypeIsNotKnown() throws J
 
         Throwable cause = exception.getCause();
         assertThat(cause, is(instanceOf(UnsupportedOperationException.class)));
-        assertThat(cause.getMessage(), is(containsString(CristinMapper.ERROR_PARSING_SECONDARY_CATEGORY)));
+        assertThat(cause.getMessage(), is(containsString(PublicationInstanceBuilder.ERROR_PARSING_SECONDARY_CATEGORY)));
```
axthosarouris, line 265: "is" is not necessary here.

axthosarouris (changes requested):
based on the comments above

axthosarouris (approved):

PR #522, opened 2021-09-02
Np 3085 accept null isbn

Eirik-Eide-Saur:


cristin-import/src/test/resources/features/BookRules.feature
```diff
@@ -107,3 +107,9 @@ Feature: Book conversion rules
       | MONOGRAFI         |
       | ANTOLOGI          |
 
+  Scenario: Mapping does not fail when a Cristin Result that is a "Book" has a null value for isbn.
+    Given a valid Cristin Result with secondary category "MONOGRAFI"
+    And the cristin result's isbn is set to null
```
axthosarouris, line 112: And the Cristin Result does not have an ISBN

cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinMapper.java
```diff
@@ -168,7 +169,9 @@ private PublicationContext buildPublicationContext()
 
     private Book buildBookForPublicationContext() throws MalformedURLException, InvalidIsbnException {
         List<String> isbnList = new ArrayList<>();
-        isbnList.add(extractIsbn());
+        if (nonNull(extractIsbn())) {
```
axthosarouris, line 172: Suggestion:
```
 private Optional<String> extractIsbn() {
        return Optional.ofNullable(extractCristinBookReport().getIsbn());
    }

    private Book buildBookForPublicationContext() throws MalformedURLException, InvalidIsbnException {
        List<String> isbnList =extractIsbn().stream().collect(Collectors.toList());
```
Optional indicates that a value may or may not be there. I think it reflects more the decision we are taking to accept empty ISBN values. we should start using more this pattern that checking for null here and there.

axthosarouris (changes requested):

axthosarouris (approved):

axthosarouris (approved):

PR #525, opened 2021-09-07
Upgraded datamodel to version 0.13.0

janbj:


axthosarouris (dismissed):

brinxmat (approved):

cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinMapper.java
```diff
@@ -37,16 +13,43 @@
...
+import no.unit.nva.model.contexttypes.UnconfirmedPublisher;
 import no.unit.nva.model.exceptions.InvalidIsbnException;
 import no.unit.nva.model.exceptions.InvalidIssnException;
+import no.unit.nva.model.exceptions.InvalidUnconfirmedSeriesException;
 import no.unit.nva.model.instancetypes.PublicationInstance;
 import no.unit.nva.model.pages.Pages;
 import nva.commons.core.attempt.Try;
 import nva.commons.core.language.LanguageMapper;
 
+import java.net.URI;
```
axthosarouris, line 28: fix the intellij idea rules for the static imports to be on top of the normal imports. Or in general we have to agree to a rule about this. It would make the PRs much smaller

cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinMapper.java
```diff
@@ -144,12 +147,12 @@ private Reference buildReference() {
...
 
     private PublicationContext buildPublicationContext()
-        throws InvalidIsbnException, MalformedURLException, InvalidIssnException {
+            throws InvalidIsbnException, InvalidIssnException, InvalidUnconfirmedSeriesException {
         if (isBook(cristinObject)) {
             return buildBookForPublicationContext();
         }
         if (isJournal(cristinObject)) {
-            return buildJournalForPublicationContext();
+            return buildUnconfirmedJournalForPublicationContext();
```
axthosarouris, line 155: maybe `buildUnconfirmedJournalPublicationContext` , i.e., without "for"

cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinMapper.java
```diff
@@ -160,33 +163,30 @@ private PublicationContext buildPublicationContext()
...
     }
 
-    private Book buildBookForPublicationContext() throws InvalidIsbnException {
+    private Book buildBookForPublicationContext() throws InvalidIsbnException, InvalidUnconfirmedSeriesException {
         List<String> isbnList = extractIsbn().stream().collect(Collectors.toList());
-        return new Book.Builder()
-                .withIsbnList(isbnList)
-                .withPublisher(extractPublisherName())
-                .build();
+        return new Book(null, null, null, buildUnconfirmedPublisher(), isbnList);
```
axthosarouris, line 168: replace nulls with constants,EMPTY_SERIES,EMPTY_SERIES_TITLE,EMPTY_SERIES_NUMBER.
 A simpler constructor would be easier to handle

cristin-import/src/test/java/cucumber/ReportFeatures.java
```diff
@@ -17,14 +19,15 @@ public ReportFeatures(ScenarioContext scenarioContext) {
     }
 
     @Then("the NVA Resource Report has a PublicationContext with publisher equal to {string}")
-    public void theNvaResourceReportHasAPublicationContextWithPublisherEqualTo(String expectedPublisher) {
+    public void theNvaResourceReportHasAPublicationContextWithPublisherEqualTo(String expectedPublisherName) {
```
axthosarouris, line 22: Maybe we should change the scenario line to be 
```
the NVA Resource Report has a PublicationContext with a publisher with name equal to {string}
```
and add a scenario line  `that the Publisher cannot be verified througt a URI.`

axthosarouris (changes requested):

cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinMapper.java
janbj: That breaks the pattern used in this class

cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinMapper.java
brinxmat: Bump the model to 0.13.1 and use the simple constructor (the one without unconfirmedSeriesTitle)

brinxmat (changes requested):

cristin-import/src/test/java/cucumber/ReportFeatures.java
janbj: if we do that I suggest a new task for it.

brinxmat (dismissed):

axthosarouris (dismissed):

axthosarouris (approved):

brinxmat (approved):

PR #526, opened 2021-09-08
Np-3087 Test and map that books and reports dont throw errors when th…

Eirik-Eide-Saur:
…ey dont have a subjectfield unless they are of the secondarycategory MONOGRAFI.

One of the existing test are changed to have the opposite effect of what it had. This is due to an earlier change in functionality but the test kept running green, due to it failing for missing subjectField


cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinMapper.java
```diff
@@ -243,11 +243,14 @@ private AdditionalIdentifier extractIdentifier() {
     }
 
     private String extractNpiSubjectHeading() {
-        if (!isBook(cristinObject)) {
+        if (!isBook(cristinObject) && !isReport(cristinObject)) {
```
axthosarouris, line 246: the code is not easily readable. Code should be readable as normal text as much as possible.
```
  private String extractNpiSubjectHeading() {
        CristinSubjectField subjectField = extractSubjectField();
        if (isNull(subjectField)) {
            return EMPTY_STRING;
        } else {
            return extractSubjectFieldCode(subjectField);
        }
    }

    private String extractSubjectFieldCode(CristinSubjectField subjectField) {
        return Optional.ofNullable(subjectField.getSubjectFieldCode())
            .map(String::valueOf)
            .orElseThrow(() -> new MissingFieldsException(CristinSubjectField.MISSING_SUBJECT_FIELD_CODE));
    }

    private boolean resourceShouldAlwaysHaveAnNpiSubjectHeading() {
        return cristinObject.getSecondaryCategory().equals(CristinSecondaryCategory.MONOGRAPH);
    }

    private boolean resourceTypeIsNotExpectedToHaveAnNpiSubjectHeading() {
        return !(isBook(cristinObject) || isReport(cristinObject));
    }

    private CristinSubjectField extractSubjectField() {
        if (resourceTypeIsNotExpectedToHaveAnNpiSubjectHeading()) {
            return null;
        }
        CristinSubjectField subjectField = extractCristinBookReport().getSubjectField();
        if (resourceShouldAlwaysHaveAnNpiSubjectHeading() && isNull(subjectField)) {
            throw new MissingFieldsException(CristinBookOrReportMetadata.SUBJECT_FIELD_IS_A_REQUIRED_FIELD);
        }
        return subjectField;
    }
```

axthosarouris (changes requested):

axthosarouris (approved):

PR #527, opened 2021-09-08
Np-3086 Test and map hovedoppgave og forskerlinjeoppgave to master_th…

Eirik-Eide-Saur:
…esis

### DIFF
--- a/cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinSecondaryCategory.java
+++ b/cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinSecondaryCategory.java
@@ -30,7 +30,7 @@ public enum CristinSecondaryCategory {
     RESEARCH_REPORT("RAPPORT", "RESEARCH_REPORT"),
     DEGREE_PHD("DRGRADAVH", "DEGREE_PHD"),
     DEGREE_MASTER("MASTERGRADSOPPG", "DEGREE_MASTER"),
-    SECOND_THESIS("HOVEDFAGSOPPGAVE", "SECOND_THESIS"),
+    SECOND_DEGREE_THESIS("HOVEDFAGSOPPGAVE", "SECOND_DEGREE_THESIS"),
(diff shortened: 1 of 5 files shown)


cristin-import/src/test/resources/features/DegreeMasterRules.feature
```diff
@@ -3,9 +3,15 @@ Feature: Mapping rules for Master thesis
   Background:
     Given a valid Cristin Result with secondary category "MASTERGRADSOPPG"
 
-  Scenario: Cristin Result "Master thesis" is converted to an NVA entry with type "DegreeMaster".
+  Scenario Outline: Cristin Result "Master thesis", "Second thesis " or "Medical thesis" is converted to an NVA entry with type "DegreeMaster".
```
axthosarouris, line 6: Can we change "Second thesis" to "Second Degree Thesis" or something like this?

cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinSecondaryCategory.java
```diff
@@ -30,6 +30,8 @@
     RESEARCH_REPORT("RAPPORT", "RESEARCH_REPORT"),
     DEGREE_PHD("DRGRADAVH", "DEGREE_PHD"),
     DEGREE_MASTER("MASTERGRADSOPPG", "DEGREE_MASTER"),
+    SECOND_THESIS("HOVEDFAGSOPPGAVE", "SECOND_THESIS"),
```
axthosarouris, line 33: Can you add a description of the second thesis in the cristin ontology file in the field "varbeidunderkatkode"?

axthosarouris (changes requested):

cristin-import/src/test/resources/features/DegreeMasterRules.feature
axthosarouris: remove the trailing space in "second thesis"

axthosarouris (changes requested):

axthosarouris (approved):

PR #539, opened 2021-09-16
Np 3108 map dates

Eirik-Eide-Saur:


cristin-import/src/test/java/cucumber/GeneralMappingRules.java
```diff
@@ -132,6 +137,18 @@ public void theNvaResourceHasACreationDateEqualTo(String expectedIsoInstant) {
...
+    @Then("the NVA Resource has a Published Date equal to {string}")
+    public void theNvaResourceHasAPublishedDateEqualTo(String expectedIsoInstant) {
+        Instant expectedInstant = Instant.parse(expectedIsoInstant);
+        assertThat(scenarioContext.getNvaEntry().getPublishedDate(), is(equalTo(expectedInstant)));
+    }
+
+    @Then("the NVA Resource has a Modified Date equal to {string}")
+    public void theNvaResourceHasAModifiedDateEqualTo(String expectedIsoInstant) {
+        Instant expectedInstant = Instant.parse(expectedIsoInstant);
+        assertThat(scenarioContext.getNvaEntry().getModifiedDate(), is(equalTo(expectedInstant)));
```
axthosarouris, line 149: Does this mean that we will create an entry in the database where the creation time will be after the modified date?

axthosarouris: Also, I think that modified date is controlled by the nva-publication-service. it does not matter what we entrer

Eirik-Eide-Saur: I do believe the Cristin entry comes with a seperate value for when it was created, so we use that one and not the one created by dynamoDB. I was told to map the Cristin value last_modified_date to the Modified date in NVA

publication-commons/src/test/java/no/unit/nva/publication/service/impl/ResourceServiceTest.java
```diff
@@ -135,6 +135,30 @@ public void createResourceWithPredefinedCreationDateStoresResourceWithCreationDa
...
+            throws TransactionFailedException, NotFoundException {
+        Publication inputPublication = PublicationGenerator.publicationWithoutIdentifier();
+        Instant predefinedPublishTime = Instant.now();
+        Instant predefinedCreatedTime = RESOURCE_CREATION_TIME;
+        Instant predefinedModifiedTime = RESOURCE_MODIFICATION_TIME;
+
+        inputPublication.setPublishedDate(predefinedPublishTime);
+        inputPublication.setCreatedDate(predefinedCreatedTime);
+        inputPublication.setModifiedDate(predefinedModifiedTime);
+
```
axthosarouris, line 149: check also that PublicationStatus is PUBLISHED

cristin-import/src/main/resources/possiblyEmptyFields.txt
```diff
@@ -18,4 +18,5 @@ entityDescription.abstract
 entityDescription.tags
 entityDescription.npiSubjectHeading
 entityDescription.reference.doi
-projects
\ No newline at end of file
+projects
+modifiedDate
```
axthosarouris, line 22: modifiedDate cannot be null. it should be equal to createdDate if not there.

cristin-import/src/test/resources/features/GeneralMappingRules.feature
```diff
@@ -69,6 +69,16 @@ Feature: Mappings that hold for all types of Cristin Results
     When the Cristin Result is converted to an NVA Resource
     Then the NVA Resource has a Creation Date equal to "2011-12-03T00:00:00Z"
 
+  Scenario: The NVA Resource Published Date is set to be the Cristin entry's creation date
+    Given that Cristin Result has created date equal to the local date "2011-12-03"
+    When the Cristin Result is converted to an NVA Resource
+    Then the NVA Resource has a Published Date equal to "2011-12-03T00:00:00Z"
+
+  Scenario: The NVA Resource Modified Date is set to be the Cristin entry's last modified date
```
axthosarouris, line 77: When the modified date is null the modified date should be equal to created date

axthosarouris (changes requested):

cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinMapper.java
```diff
@@ -133,6 +135,12 @@ private Instant extractEntryCreationDate() {
             .orElse(null);
     }
 
+    private Instant extractEntryLastModifiedDate() {
+        return Optional.ofNullable(cristinObject.getEntryLastModifiedDate())
+                .map(ld -> ld.atStartOfDay().toInstant(zoneOffset()))
+                .orElse(extractEntryCreationDate());
```
axthosarouris, line 141: use orElseGet(()->) the difference between the two is that orElse(actionHere()) will always execute even the Optional is succesfull. 
This is happenning because when I do foo(boo()) boo is always called before foo and the result of boo's execution is passed to foo. So when you call orElse(boo()), you first calculate the boo() value and then you may or may not use it. 
Instead you can use orElseGet(()->) to pass as a parameter a function that may or  may not be executed

axthosarouris (changes requested):

publication-commons/src/main/java/no/unit/nva/publication/service/impl/ResourceService.java
```diff
@@ -96,6 +99,16 @@ public Publication createPublicationWithPredefinedCreationDate(Publication input
         return insertResource(newResource);
     }
 
+    public Publication createPublicationWhilePersistingEntryFromLegacySystems(Publication inputData)
+            throws TransactionFailedException {
+        Resource newResource = Resource.fromPublication(inputData);
```
axthosarouris, line 104: We have created a weird situation where we allow a publication to be DRAFT, i.e., NOT PUBLISHED but still to have a published date. Except for the fact that this does not stand, this date will disappear when the Publication is actually Published. Since all Publications from legacy systems are already published, I suggest we just set that status of all Publications to PUBSLISHED when this method is  called. That is to say it is the Service and this specific method that will set the publication status to PUBLISHED so that a situtaion where a Publication is not published but still has a publication date is not possible.

axthosarouris (changes requested):
Resource service should set the publication status of all Publications imported from legacy systems to PUBLISHED

axthosarouris (approved):

PR #543, opened 2021-09-20
NP-3061: Cristin publisher

axthosarouris:
The PR is longer than it should be, but this is because of the Cucumber tests. Please read the cucumber tests first.

Fixes NP-3061.  Maps Cristin publishers to NVA publishers. In Cristin, publishers are stored in 2 places, in a field with the name "utgivernanv" and in a field with the name "forlag"  (Example below).  The implemented logic is the following:

Case 1: if the field `forlag` contains a non empty NSD code, then Cristing publisher is mapped to an NVA `Publisher` with ID that is the URI produced by the URL location of our NSD proxy, the NSD code and the publication year. 
Example: for the entry given below the NVA publisher would have the ID https://api.dev.nva.aws.unit.no/publication-channels/18424/<publicationYear>

Case 2: if the field "forlag" does not contain an NSD code, then the Cristin publisher is mapped to an NVA `UnconfirmedPublisher` and the publisher name will be equal to the value of the field `forlag.forlagsnavn`. 

Case 3: if the field `forlag.forlagsnavn` is empty and `utgivernanv` is not empty, then we create an `UnconfirmedPublisher` with name equal to the value of the field `utgivernavn`

Case 4: If there is no information about the Publication's publisher we throw an Exception to indicate that the mapping cannot continue.

### DIFF
--- a/cristin-import/src/main/java/no/unit/nva/cristin/mapper/nva/NvaBookLikeBuilder.java
+++ b/cristin-import/src/main/java/no/unit/nva/cristin/mapper/nva/NvaBookLikeBuilder.java
@@ -1,28 +1,45 @@
 package no.unit.nva.cristin.mapper.nva;
 
+import java.net.URI;
 import java.util.List;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import no.unit.nva.cristin.mapper.CristinBookOrReportMetadata;
 import no.unit.nva.cristin.mapper.CristinObject;
+import no.unit.nva.cristin.mapper.CristinPublisher;
+import no.unit.nva.cristin.mapper.Nsd;
+import no.unit.nva.cristin.mapper.nva.exceptions.NoPublisherException;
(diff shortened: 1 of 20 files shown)


cristin-import/src/main/java/no/unit/nva/cristin/mapper/nva/NvaBookLikeBuilder.java
```diff
@@ -32,4 +49,58 @@ protected BookSeries buildSeries() {
...
+            .map(this::createConfirmedPublisher);
+    }
+
+    private Optional<Integer> extractPublishersNsdCode() {
+        return extractBookOrReportMetadata()
+            .map(CristinBookOrReportMetadata::getCristinPublisher)
+            .map(CristinPublisher::getNsdCode);
+    }
+
+    private PublishingHouse createUnconfirmedPublisher() {
```
axthosarouris, line 76: necessary methods in order to be able to return the same type (PublishingHouse)

brinxmat (approved):

PR #551, opened 2021-09-30
NP-3184 Map secondary categories for Event.

Eirik-Eide-Saur:


cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinMapper.java
```diff
@@ -239,6 +244,10 @@ private Chapter buildChapterForPublicationContext() {
         return new Chapter.Builder().build();
     }
 
+    private PublicationContext buildEventForPublicationContext() {
+        return new Event.Builder().build();
```
axthosarouris, line 248: empty event?

axthosarouris (approved):

PR #564, opened 2021-10-06
NP-3238 Include timestamps in events and use it when creating error-f…

Eirik-Eide-Saur:
…olders.

### DIFF
--- a/s3-import-commons/src/main/java/no/unit/nva/publication/s3imports/FilenameEventEmitter.java
+++ b/s3-import-commons/src/main/java/no/unit/nva/publication/s3imports/FilenameEventEmitter.java
@@ -114,7 +114,7 @@ public class FilenameEventEmitter implements RequestStreamHandler {
         UriWrapper bucketUri = inputFolderUri.getHost();
         return bucketUri
                    .addChild(ERRORS_FOLDER)
-                   .addChild(timestampToString(clock.instant()))
+                   .addChild(timestampToString(timestamp))
                    .addChild(inputFolderUri.getPath())
                    .addChild(ERROR_REPORT_FILENAME);
     }


cristin-import/src/main/java/no/unit/nva/cristin/lambda/CristinEntryEventConsumer.java
```diff
@@ -2,13 +2,15 @@
 
 import static no.unit.nva.cristin.lambda.constants.HardcodedValues.HARDCODED_PUBLICATIONS_OWNER;
 import static no.unit.nva.publication.s3imports.ApplicationConstants.MAX_SLEEP_TIME;
+import static no.unit.nva.publication.s3imports.FileEntriesEventEmitter.timestampToString;
```
axthosarouris, line 5: Utility methods should be gathered together in independent classes. If we ever get rid of the FileEntriesEventEmitter we would have to change all classes that are importing this method but have otherwise nothing to do with the class.

cristin-import/src/test/java/no/unit/nva/cristin/CristinDataGenerator.java
```diff
@@ -199,7 +202,7 @@ public static CristinObject randomObject(String secondaryCategory) {
     public static <T> AwsEventBridgeEvent<FileContentsEvent<JsonNode>> toAwsEvent(T inputData) {
         AwsEventBridgeEvent<FileContentsEvent<JsonNode>> event = new AwsEventBridgeEvent<>();
         JsonNode cristinData = convertToJsonNode(inputData);
-        FileContentsEvent<JsonNode> eventDetail = new FileContentsEvent<>(randomUri(), cristinData);
+        FileContentsEvent<JsonNode> eventDetail = new FileContentsEvent<>(randomUri(), TIMESTAMP, cristinData);
```
axthosarouris, line 205: why hardcoded value here. Also the name of the parameter gives no information. This is an ARBITRARY_TIMESTAMP or shorter SOME_TIMESTAMP. But if you are not using the value anywhere else, why not being a random value?

s3-import-commons/src/main/java/no/unit/nva/publication/s3imports/FileEntriesEventEmitter.java
```diff
@@ -141,18 +142,24 @@ private UriWrapper generateErrorReportUri(ImportRequest input, Try<List<PutEvent
...
         return bucket
                    .addChild(ERRORS_FOLDER)
+                   .addChild(timestampToString(input.getTimestamp()))
                    .addChild(errorType)
                    .addChild(inputUri.getParent().map(UriWrapper::getPath).orElse(UnixPath.EMPTY_PATH))
                    .addChild(makeFileExtensionError(inputUri.getFilename()));
     }
 
+    public static String timestampToString(Instant timestamp) {
+        return timestamp.toString().replaceAll(" ", "_");
```
axthosarouris, line 152: Standard solutions are always better: `return DateTimeFormatter.ISO_INSTANT.format(timestamp);`

s3-import-commons/src/main/java/no/unit/nva/publication/s3imports/FilenameEventEmitter.java
```diff
@@ -2,6 +2,7 @@
 
 import static java.util.Objects.isNull;
 import static no.unit.nva.publication.s3imports.ApplicationConstants.ERRORS_FOLDER;
+import static no.unit.nva.publication.s3imports.ApplicationConstants.defaultClock;
```
axthosarouris, line 5: FilenameEventEmitter produces errors as well and its error paths should be updated too.

s3-import-commons/src/test/java/no/unit/nva/publication/s3imports/FileEntriesEventEmitterTest.java
```diff
@@ -123,6 +126,16 @@ public void handlerEmitsEventWithResourceWhenFileUriExistsAndContainsDataAsJsonO
         assertThat(emittedResourceObjects, containsInAnyOrder(FILE_01_CONTENTS));
     }
 
+    @Test
+    public void handlerEmitsEventWithTimestampInformation() {
```
axthosarouris, line 130: not accurate enough. this description can mean anything. The test name is 
`handlerEmitsEventWithTimestampAcquiredByImportRequest`.
If you do not specify the exact functionality in the title, the next person will have no idea what this test does without looking through the code. It is essential that one understands the implemented functionality by just reading the test names. That way one can refactor the code being sure that they do not break functionality.

Our job is basically to define functionality (currently through tests) and unfortunately the technology is not that advanced and so we have to implement the functionality as well. But the core of our job is to define the functionality well.

s3-import-commons/src/test/java/no/unit/nva/publication/s3imports/FilenameEventEmitterTest.java
```diff
@@ -61,19 +65,22 @@
     private static final Context CONTEXT = mock(Context.class);
     public static final String LIST_ALL_FILES = ".";
     public static final URI EMPTY_URI = null;
+    public static final Instant FIXED_INSTANT = Instant.parse("2017-02-03T11:25:30.00Z");
```
axthosarouris, line 68: It is a bit worrying that you use the same string everywhere. It could be that we are reading the value from somewhere else. You can make the fixedInstant into a field that gets updated (Instant.now()) before each test to ensure some randomness

axthosarouris (changes requested):
Place-holder comment for the comments above

s3-import-commons/src/main/java/no/unit/nva/publication/s3imports/FilenameEventEmitter.java
```diff
@@ -107,6 +114,7 @@ private UriWrapper createErrorReportUri(ImportRequest request) {
         UriWrapper bucketUri = inputFolderUri.getHost();
         return bucketUri
                    .addChild(ERRORS_FOLDER)
+                   .addChild(timestampToString(clock.instant()))
```
axthosarouris, line 117: why another instant here?

PR #567, opened 2021-10-07
merge api handlers in one package

axthosarouris:
Most handlers have the same dependencies and it seems natural to put them all together in one package, so that they can share one JsonMapper.


create-publication/src/test/java/no/unit/nva/publication/fetch/PublicationsByOwnerHandlerTest.java
```diff
@@ -85,11 +85,9 @@ public void handlerReturnsOkResponseOnValidInput() throws IOException {
     @Test
     @DisplayName("handler Returns BadRequest Response On Empty Input")
     public void handlerReturnsBadRequestResponseOnEmptyInput() throws IOException {
-        InputStream input = new HandlerUtils(objectMapper)
-                                .requestObjectToApiGatewayRequestInputSteam(null, null);
+        InputStream input = new HandlerRequestBuilder<Void>(objectMapper).build();
```
axthosarouris, line 88: Use newest HandlerRequestBuilder than the old version.

settings.gradle
```diff
@@ -10,12 +10,9 @@ include 'create-publication'
 include 's3-import-commons'
 include 'publication-events'
 include 'delete-draft-publication-handler'
-include 'delete-publication'
```
axthosarouris, line 13: all handlers were moved to create-publication that will be renamed later

brinxmat (approved):

PR #570, opened 2021-10-08
Np 3239 split higher level exceptions in s3

Eirik-Eide-Saur:


cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinMapper.java
```diff
@@ -208,14 +210,28 @@ private Reference buildReference() {
             = new PublicationInstanceBuilderImpl(cristinObject);
         PublicationInstance<? extends Pages> publicationInstance
             = publicationInstanceBuilderImpl.build();
-        PublicationContext publicationContext = attempt(this::buildPublicationContext).orElseThrow();
+        PublicationContext publicationContext = attempt(this::buildPublicationContext)
```
axthosarouris, line 213: I was thinking, that the whole functionality of building a Reference could be moved to a new class ReferenceBuilder, but this is out of scope of this PR

cristin-import/src/main/java/no/unit/nva/cristin/mapper/PublicationInstanceBuilderImpl.java
```diff
@@ -36,9 +39,9 @@ public PublicationInstanceBuilderImpl(CristinObject cristinObject) {
         } else if (isEvent(cristinObject)) {
             return new EventBuilder(cristinObject).build();
         } else if (cristinObject.getMainCategory().isUnknownCategory()) {
-            throw new UnsupportedOperationException(ERROR_PARSING_MAIN_CATEGORY);
+            throw new UnsupportedMainCategoryRuntimeException(ERROR_PARSING_MAIN_CATEGORY);
```
axthosarouris, line 42: remove "Runtime"

axthosarouris (changes requested):

PR #575, opened 2021-10-12
NP-3284 Make invalid ISSN in journals throw a InvalidIssnRuntimeExcep…

Eirik-Eide-Saur:
…tion.

### DIFF
--- a/cristin-import/src/main/java/no/unit/nva/cristin/mapper/nva/exceptions/ExceptionHandling.java
+++ b/cristin-import/src/main/java/no/unit/nva/cristin/mapper/nva/exceptions/ExceptionHandling.java
@@ -1,0 +1,61 @@
+package no.unit.nva.cristin.mapper.nva.exceptions;
+
+import no.unit.nva.model.exceptions.InvalidIsbnException;
+import no.unit.nva.model.exceptions.InvalidIssnException;
+import nva.commons.core.JacocoGenerated;
+
+public final class ExceptionHandling {
+
+    @JacocoGenerated
+    public ExceptionHandling() {
+
+    }
+
+    @SuppressWarnings("PMD.NPathComplexity")
+    public static RuntimeException castToCorrectRuntimeException(Exception exception) {
+        if (exception instanceof InvalidIssnRuntimeException) {
+            return (InvalidIssnRuntimeException) exception;
+        }
+        if (exception instanceof InvalidIsbnRuntimeException) {
+            return (InvalidIsbnRuntimeException) exception;
+        }
+        if (exception instanceof InvalidIsbnException) {
+            return new InvalidIsbnRuntimeException(exception);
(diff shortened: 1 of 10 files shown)


cristin-import/src/test/java/no/unit/nva/cristin/lambda/CristinEntryEventConsumerTest.java
```diff
@@ -287,8 +287,22 @@ public void handlerThrowsInvalidIsbnRuntimeExceptionWhenTheIsbnIsInvalid() throw
...
+        JsonNode cristinObjectWithInvalidIssn = CristinDataGenerator.bookObjectWithInvalidIssn();
+        AwsEventBridgeEvent<FileContentsEvent<JsonNode>> awsEvent =
+                CristinDataGenerator.toAwsEvent(cristinObjectWithInvalidIssn);
+        InputStream inputStream = IoUtils.stringToStream(awsEvent.toJsonString());
+
+        Executable action = () -> handler.handleRequest(inputStream, outputStream, CONTEXT);
+
+        RuntimeException exception = assertThrows(RuntimeException.class, action);
+        Throwable cause = exception.getCause();
+        assertThat(cause, is(instanceOf(InvalidIssnRuntimeException.class)));
```
axthosarouris, line 300: The InvalidIssnRutimeException is a RuntimeException and should be thrown directly, not being the cause.

cristin-import/src/test/java/no/unit/nva/cristin/lambda/CristinEntryEventConsumerTest.java
```diff
@@ -287,8 +287,22 @@ public void handlerThrowsInvalidIsbnRuntimeExceptionWhenTheIsbnIsInvalid() throw
...
+    }
+
+    @Test
+    public void handlerThrowsInvalidIssnRuntimeExceptionWhenTheJournalIssnIsInvalid() throws JsonProcessingException {
```
axthosarouris, line 304: same for this test

axthosarouris (changes requested):

cristin-import/src/main/java/no/unit/nva/cristin/mapper/nva/exceptions/ExceptionHandling.java
```diff
@@ -0,0 +1,61 @@
...
+import nva.commons.core.JacocoGenerated;
+
+public final class ExceptionHandling {
+
+    @JacocoGenerated
+    public ExceptionHandling() {
+
+    }
+
+    @SuppressWarnings("PMD.NPathComplexity")
```
axthosarouris, line 14: When things like this happen, it is an indication that we do something wrong.

axthosarouris (approved):

PR #590, opened 2021-10-21
NP-3319 Fix documentation for top level Cristin values.

Eirik-Eide-Saur:


docs/cristin.ttl
```diff
@@ -319,8 +319,9 @@
     """ ;
     rdfs:range         xsd:date ;
     rdfs:domain        :Result, :VITENSKAPELIGARBEID_LOKAL_Class ;
-    :mapstonvafield    "createdDate when the domain is :Result" ;
-    :currently_ignored "true" .
+    :mapstonvafield    """  createdDate when the domain is :Result
```
axthosarouris, line 322: Probably we should remove the mapping descriptions since now we have the cucumber scenarios.

docs/cristin.ttl
```diff
@@ -434,10 +436,11 @@
 
 :arstall_rapportert
     a                  :cristinfield ;
-    rdfs:comment       "Year that the resource was included in an NVI report." ;
+    rdfs:comment       """Year that the resource was included in an NVI report. For Results of type Book it is used to generate a confirmedPublisher.
```
axthosarouris, line 439: Do not add anything to do with conversion to NVA. Better to use the cucumber scenarios for this. The reason I want to avoid describing the NVA conversion here, is that it is difficult to keep it in sync since this is not testable.

axthosarouris (changes requested):

PR #591, opened 2021-10-21
NP-3320 Fix documentation for varbeid_sprak.

Eirik-Eide-Saur:


docs/cristin.ttl
```diff
@@ -502,7 +503,7 @@ See varbeid.kildekode. Mostly superseded by VARBEID_KILDE. It contains only the
     rdfs:range      :VARBEID_SPRAK_Class ;
     rdfs:domain     :Result ;
     :mapstonvafield """
-        entityDescription.mainTitle, entityDescription.language
+        entityDescription.mainTitle, entityDescription.language, entityDescription.mainLanguageAbstract;
```
axthosarouris, line 506: remove completely the mapstonvafield. It was a bad idea after all.

docs/cristin.ttl
```diff
@@ -541,7 +542,7 @@ See varbeid.kildekode. Mostly superseded by VARBEID_KILDE. It contains only the
 :status_original
     a                  :cristinfield ;
     rdfs:comment       """
-    Indicator of whether the title in the respective VARBEID_SPRAK is the original title or not.
+    Indicator of whether the title and the abstract in the respective VARBEID_SPRAK is in the original language and if so the original title and abstract.
```
axthosarouris, line 545: and if so, then they are the original title and abtract

PR #618, opened 2021-11-02
ExpandedResourcePersistenceHandler tests for sent messages

axthosarouris:


expansion/src/main/java/no/unit/nva/expansion/model/ExpandedResource.java
```diff
@@ -114,7 +117,8 @@ public boolean equals(Object o) {
             return false;
         }
         ExpandedResource that = (ExpandedResource) o;
-        return Objects.equals(this.allFields, that.allFields);
+        //Comparison can only happen when comparing them as json nodes.
+        return Objects.equals(this.asJsonNode(), that.asJsonNode());
```
axthosarouris, line 121: Custom equality check for ExpandedResource because the field "type" is not inlcluded `allFields` Map.

brinxmat (approved):

PR #623, opened 2021-11-08
Post Deployment Initialization function

axthosarouris:
This is the very first version of a stack initialization function. The function belongs to the stack because it will be easier to use the stack resources. 

### Background:
In order for the initialization function to be executed after a successful pipeline, there are two alternatives:
1. The pipeline calls the function directly
     This method requires that we know from before the name (and therefore the ARN) of both the stack and the init function before the stack is created. In other words, each stack should have deterministic name and each init function should have a deterministic name as well.  
2. The function listens to pipeline events and is activated when a pipeline execution is successful.
   This approach requires that the pipeline name is passed as a parameter to the stack. Probably this is an easier way to go. 

To achieve the second approach we use CodePipeline EventBridge events as they are described here: https://docs.aws.amazon.com/codepipeline/latest/userguide/detect-state-changes-cloudwatch-events.html

### This PR.

This PR contains only the logic of a very simple handler, that logs the event. The connection to the pipeline is not yet implemented but it will follow.

### DIFF
--- a/publication-event-handlers/src/main/java/no/unit/nva/publication/events/handlers/initialization/InitializationHandler.java
+++ b/publication-event-handlers/src/main/java/no/unit/nva/publication/events/handlers/initialization/InitializationHandler.java
@@ -1,0 +1,22 @@
+package no.unit.nva.publication.events.handlers.initialization;
+
+import com.amazonaws.services.lambda.runtime.Context;
+import no.unit.nva.events.handlers.EventHandler;
+import no.unit.nva.events.models.AwsEventBridgeEvent;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class InitializationHandler extends EventHandler<PipelineEvent, Void> {
+
+    private static final Logger logger = LoggerFactory.getLogger(InitializationHandler.class);
+
+    protected InitializationHandler() {
+        super(PipelineEvent.class);
+    }
+
+    @Override
+    protected Void processInput(PipelineEvent input, AwsEventBridgeEvent<PipelineEvent> event, Context context) {
+        logger.info(event.toJsonString());
+        return null;
+    }
+}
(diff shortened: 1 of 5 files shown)


publication-event-handlers/src/main/java/no/unit/nva/publication/events/handlers/initialization/InitializationHandler.java
```diff
@@ -0,0 +1,22 @@
+package no.unit.nva.publication.events.handlers.initialization;
+
+import com.amazonaws.services.lambda.runtime.Context;
+import no.unit.nva.events.handlers.EventHandler;
+import no.unit.nva.events.models.AwsEventBridgeEvent;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class InitializationHandler extends EventHandler<PipelineEvent, Void> {
```
axthosarouris, line 9: Probably we can use the event logic we already have

publication-event-handlers/src/main/java/no/unit/nva/publication/events/handlers/initialization/PipelineEvent.java
```diff
@@ -0,0 +1,23 @@
+package no.unit.nva.publication.events.handlers.initialization;
+
+public class PipelineEvent {
+
+    private String pipeline;
```
axthosarouris, line 5: Useful information of the pipeline event.

publication-event-handlers/src/test/resources/initialization/pipeline_succeeded_event.json
```diff
@@ -0,0 +1,18 @@
+{
```
axthosarouris, line 1: Sample event copied from https://docs.aws.amazon.com/codepipeline/latest/userguide/detect-state-changes-cloudwatch-events.html

template.yaml
```diff
@@ -399,11 +399,11 @@ Resources:
       Handler: no.unit.nva.publication.fetch.FetchPublicationHandler::handleRequest
       Runtime: java11
       MemorySize: 1408
-#      AutoPublishAlias: live
-#      DeploymentPreference:
-#        Type: AllAtOnce
-#      ProvisionedConcurrencyConfig:
-#        ProvisionedConcurrentExecutions: 1
+      AutoPublishAlias: live
```
axthosarouris, line 402: Part of another PR that is not yet merged

brinxmat (approved):

PR #630, opened 2021-11-09
Add row-version to Dynamo entries

axthosarouris:


storage-model/src/test/java/no/unit/nva/publication/storage/model/ResourceTest.java
```diff
@@ -50,7 +50,7 @@
 import org.junit.jupiter.api.Test;
 
 public class ResourceTest {
-    
+
```
axthosarouris, line 53: only space changes

brinxmat (approved):

PR #664, opened 2021-12-09
Resource service uses affiliation service

axthosarouris:
Changes:
1. Added AffiliationService to ResourceService.
2. Changed all constructors to include an httpClient that will be used by AffiliationService and possibly other modules  that talk to external services.
3. Added a Fake client for testing
4. Added a Fake client (almost a copy of the testing one) for Cristin import because we need the Affiliation tag is not very important for imported resources for now and we also do not know what tag to put. The Fake client will prevent the Cristin import to call the Person service 2 million times.


cristin-import/src/main/java/no/unit/nva/cristin/lambda/CristinEntryEventConsumer.java
```diff
@@ -41,7 +42,7 @@
 /**
  * Accepts events generated by {@link FileEntriesEventEmitter} with topic equal to {@link
  * FileEntriesEventEmitter#FILE_CONTENTS_EMISSION_EVENT_TOPIC} and subtopic equal to {@link
- * CristinEntryEventConsumer#EVENT_SUBTOPIC}
+ * CristinEntryEventConsumer#EVENT_SUBTOPIC}.
```
axthosarouris, line 45: Checkstyle complained.

cristin-import/src/main/java/no/unit/nva/cristin/lambda/CristinEntryEventConsumer.java
```diff
@@ -57,6 +58,10 @@
     public static final String ERRORS_FOLDER = "errors";
 
     private static final Logger logger = LoggerFactory.getLogger(CristinEntryEventConsumer.class);
+    private static final Clock CLOCK = Clock.systemDefaultZone();
+    //imported cristin entries have for now no owner. Calling the Person proxy will only increase the cost of import.
+    private static final HttpClient CLIENT_AVOIDING_REQUESTS_FOR_RESOURCES_WITHOUT_OWNERS = new FakeHttpClient();
```
axthosarouris, line 63: Using a Fake Http Client when we are imporing data from Cristin because these entries have a placeholder for owner. There is no point invoking 2 million times the Person proxy for an imaginary user. In addition, we want all resources to be visible to us (Unit) until the final import is done.

cristin-import/src/main/java/no/unit/nva/cristin/lambda/FakeHttpClient.java
```diff
@@ -0,0 +1,100 @@
+package no.unit.nva.cristin.lambda;
```
axthosarouris, line 1: This class exists twice. Once in the testing library, and once here only for the cristin service. They give a different response, but the main reason for the replication is that I do not want to include all testing libraries in the main code.

cristin-import/src/main/resources/fake_person_api_response.json
```diff
@@ -0,0 +1,7 @@
+[
```
axthosarouris, line 1: A fake Person API response for the Cristin import.

cristin-import/src/test/java/no/unit/nva/cristin/lambda/CristinEntryEventConsumerTest.java
```diff
@@ -86,19 +85,18 @@
...
     private ByteArrayOutputStream outputStream;
     private ResourceService resourceService;
     private FakeS3Client s3Client;
+    private HttpClient httpClient;
 
     @BeforeEach
     public void init() {
         super.init();
-        resourceService = new ResourceService(super.client, Clock.systemDefaultZone());
+        httpClient = new FakeHttpClient();
```
axthosarouris, line 98: The only real change is here. The rest is improvements made the IDE

doi-requests/src/main/java/no/unit/nva/doirequest/list/ListDoiRequestsHandler.java
```diff
@@ -83,7 +84,7 @@ private static MessageService defaultMessageService(AmazonDynamoDB client, Clock
 
     @JacocoGenerated
     private static DoiRequestService defaultDoiRequestService(AmazonDynamoDB client, Clock clock) {
-        return new DoiRequestService(client, clock);
+        return new DoiRequestService(client, EXTERNAL_SERVICES_HTTP_CLIENT, clock);
```
axthosarouris, line 87: EXTERNAL_SERVICES_HTTP_CLIENT is the default HttpClient used for communicating with the Person proxy. For now it is used only the Person proxy but it could be used for other services as well.

doi-requests/src/test/java/no/unit/nva/doirequest/create/CreateDoiRequestHandlerTest.java
```diff
@@ -68,8 +70,9 @@
     public void initialize() {
         init();
         setupClock();
-        resourceService = new ResourceService(client, mockClock);
-        doiRequestService = new DoiRequestService(client, mockClock);
+        var httpClient = new FakeHttpClient();
```
axthosarouris, line 73: The FakeHttpClient returns a hardcoded value,

doi-requests/src/main/java/no/unit/nva/doirequest/create/CreateDoiRequestHandler.java
```diff
@@ -33,20 +35,22 @@
...
+
+    private static final Clock CLOCK = Clock.systemDefaultZone();
     private final DoiRequestService doiRequestService;
     private final MessageService messageService;
     private final ResourceService resourceService;
 
     @JacocoGenerated
     public CreateDoiRequestHandler() {
-        this(AmazonDynamoDBClientBuilder.defaultClient(), Clock.systemDefaultZone());
+        this(AmazonDynamoDBClientBuilder.defaultClient(), EXTERNAL_SERVICES_HTTP_CLIENT, CLOCK);
```
axthosarouris, line 46: EXTERNAL_SERVICES_HTTP_CLIENT is the HttpClient used for Bare. Since, it is likely to be used for connecting to other services as well, I want to avoid having to create one HttpClient for each service.

publication-commons/src/main/java/no/unit/nva/publication/service/impl/ResourceService.java
```diff
@@ -85,10 +89,14 @@ public ResourceService(AmazonDynamoDB client, Clock clock, Supplier<SortableIden
         this.readResourceService = new ReadResourceService(client, RESOURCES_TABLE_NAME);
         this.updateResourceService =
             new UpdateResourceService(client, RESOURCES_TABLE_NAME, clockForTimestamps, readResourceService);
+        var affiliationService = AffiliationSelectionService.create(externalServicesHttpClient);
+        attempt(() -> affiliationService.fetchAffiliation("make PMD happy")).orElseThrow();
```
axthosarouris, line 93: This line will be removed on the next PR. Added only to make it possible to include the affiliationService without actually using it and making PMD happy.

janbj (approved):

PR #668, opened 2021-12-13
Use latest NVA-datamodel

axthosarouris:


buildSrc/src/main/groovy/nva-publication-api.java-common-conventions.gradle
```diff
@@ -36,11 +36,14 @@ test {
     useJUnitPlatform {
         excludeTags("RemoteTest")
     }
-    failFast = true
+    failFast = false
     testLogging {
         events 'skipped', 'passed', 'failed'
+        showCauses true
+        exceptionFormat "full"
```
axthosarouris, line 43: effort to show failure reason in console

PR #678, opened 2022-01-05
Add resourceId to expanded messages and doi requests

janbj:


expansion/src/main/java/no/unit/nva/expansion/model/ExpandedMessage.java
```diff
@@ -27,6 +27,7 @@
     private MessageStatus status;
     private String sender;
     private SortableIdentifier resourceIdentifier;
+    private URI resourceId;
```
brinxmat, line 30: Should these not all be "id"?

expansion/src/main/java/no/unit/nva/expansion/ResourceExpansionService.java
```diff
@@ -12,4 +13,6 @@
     ExpandedDataEntry expandEntry(DataEntry dataEntry) throws JsonProcessingException, NotFoundException;
 
     Set<URI> getOrganizationIds(DataEntry dataEntry) throws NotFoundException;
+
+    URI getResourceId(SortableIdentifier identifier);
```
axthosarouris, line 17: I do not think that this method is necessary. Then we create some weird dependencies where the data entry needs the service and the service needs the data entries.
Suggestion:
```
public static ExpandedDoiRequest create(DoiRequest doiRequest,
                                            ResourceExpansionService resourceExpansionService)
        throws NotFoundException {
        ExpandedDoiRequest expandedDoiRequest = ExpandedDoiRequest.fromDoiRequest(doiRequest);
        Set<URI> ids = resourceExpansionService.getOrganizationIds(doiRequest);
        expandedDoiRequest.setOrganizationIds(ids);
        URI resourceId =  new UriWrapper(ExpansionConstants.ID_NAMESPACE)
            .addChild(expandedDoiRequest.getIdentifier().toString()).getUri();
        expandedDoiRequest.setResourceId(resourceId);
        return expandedDoiRequest;
    }
```

expansion/src/main/java/no/unit/nva/expansion/ResourceExpansionServiceImpl.java
```diff
@@ -60,4 +62,9 @@ public ExpandedDataEntry expandEntry(DataEntry dataEntry) throws JsonProcessingE
         }
         return Collections.emptySet();
     }
+
+    @Override
+    public URI getResourceId(SortableIdentifier identifier) {
+        return new UriWrapper(ExpansionConstants.ID_NAMESPACE).addChild(identifier.toString()).getUri();
```
axthosarouris, line 68: you can put this code directly into the creation of the objects as shown above

expansion/src/main/java/no/unit/nva/expansion/model/ExpandedMessage.java
axthosarouris: I think that the misunderstanding is actually the term Resource. We use it in two different ways. One way is as a better name to Publication, but we also view everything in the database as a Resource. e.g. a message is a Resource and a DoiRequest is a Resource. We can agree that all things are Resources, but we should find a name for the Publication that is unique.

axthosarouris (changes requested):

expansion/src/main/java/no/unit/nva/expansion/model/ExpandedMessage.java
janbj: Skal vi ta det i https://unit.atlassian.net/browse/NP-3524 ?

brinxmat (approved):

axthosarouris (approved):

PR #680, opened 2022-01-05
Remove dependency from identity service

axthosarouris:
Fixing an old hack, where we have to check for an AccessRight before performing an Action. The class was copy-pasted from the Identity Service


publication-rest/src/main/java/no/unit/nva/publication/AccessRight.java
```diff
@@ -0,0 +1,50 @@
+package no.unit.nva.publication;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonValue;
+import java.util.Arrays;
+import java.util.Locale;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+public enum AccessRight {
```
axthosarouris, line 10: The class was copy-pasted from the Identity Service

brinxmat (approved):

PR #683, opened 2022-01-10
fix: ExpandedResourceConversation does not have own identifier

axthosarouris:
A ResourceConversation is an aggregation(collection) of messages associated with a specific Publication. It is not a resource on its own and therefore it cannot have a permanent identifier. It is connected though to a publicationIdentifier.

This fix renames the field and fixes the tests so that they do not have the expectation that  the ResourceConversation has a field called "identifier"

### DIFF
--- a/expansion/src/test/java/no/unit/nva/expansion/model/ExpandedDatabaseEntryTest.java
+++ b/expansion/src/test/java/no/unit/nva/expansion/model/ExpandedDatabaseEntryTest.java
@@ -11,12 +11,12 @@ import static org.hamcrest.core.IsEqual.equalTo;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
-import java.net.URI;
 import java.time.Clock;
 import java.util.stream.Stream;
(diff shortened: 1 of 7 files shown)


expansion/src/test/java/no/unit/nva/expansion/model/ExpandedDatabaseEntryTest.java
```diff
@@ -31,54 +31,101 @@
 
     private static final ResourceExpansionService resourceExpansionService = new FakeResourceExpansionService();
 
-    public static Stream<ExpandedDataEntry> entryProvider() throws JsonProcessingException, NotFoundException {
-        return Stream.of(randomResource(), randomDoiRequest(), randomMessage(), randomResourceConversation());
+    public static Stream<ExpandedDataEntryWithAssociatedPublication> entryProvider()
+        throws JsonProcessingException, NotFoundException {
+        return Stream.of(ExpandedDataEntryWithAssociatedPublication.create(ExpandedResource.class),
+                         ExpandedDataEntryWithAssociatedPublication.create(ExpandedDoiRequest.class),
+                         ExpandedDataEntryWithAssociatedPublication.create(ExpandedMessage.class),
```
janbj, line 38: Should we preserve ExpandedMessage or remove it?

janbj (approved):

PR #695, opened 2022-01-31
NP-3574: Internal API and echo handler for creating a published publi…

axthosarouris:
…cation


publication-rest/src/test/java/no/unit/nva/publication/create/CreatePublishedPublicationHandlerTest.java
```diff
@@ -0,0 +1,51 @@
...
+
+    @BeforeEach
+    public void init() {
+        this.outputStream = new ByteArrayOutputStream();
+    }
+
+    @Test
+    void shouldReturnAPublicationResponseWhenRequestIsReceived()
+        throws IOException {
+        Publication publication = new Publication();
```
brinxmat, line 37: Var

brinxmat (approved):

PR #698, opened 2022-02-02
update commons

axthosarouris:


gradle/libs.versions.toml
```diff
@@ -1,7 +1,7 @@
 [versions]
 
-nva = { strictly = '1.18.4' }
-nvaDatamodel = { strictly = '0.14.23' }
+nva = { strictly = '1.20.7' }
+nvaDatamodel = { strictly = '0.15.1' }
```
axthosarouris, line 4: upgrade to latest version of datamodel will happen in later PR. This is the oldest and therefore the most compatible version of the datamodel with the latest commons

brinxmat (approved):

PR #705, opened 2022-02-03
Np 3574 create default constructor for handler

axthosarouris:


publication-commons/src/main/java/no/unit/nva/publication/PublicationServiceConfig.java
```diff
@@ -13,13 +19,31 @@
...
     public static final HttpClient EXTERNAL_SERVICES_HTTP_CLIENT = HttpClient.newBuilder().build();
+    public static final AmazonDynamoDB DEFAULT_DYNAMODB_CLIENT = defaultDynamoDbClient();
 
     public static final ObjectMapper dtoObjectMapper = JsonUtils.dtoObjectMapper;
 
     private PublicationServiceConfig() {
 
     }
+
+    @JacocoGenerated
```
axthosarouris, line 35: moved constructors of clients and services in one class

doi-requests/build.gradle
```diff
@@ -37,5 +37,9 @@ dependencies {
 
 }
 
+test{
+    environment "AWS_REGION", "eu-west-1"
```
axthosarouris, line 41: Moving the default constructors of clients from the handlers, down to the service, requires more modules to be aware of this env variable.

brinxmat (approved):

PR #716, opened 2022-02-16
Np 4260 general support conversations exclude doi request messages

axthosarouris:


expansion/src/test/java/no/unit/nva/expansion/ResourceExpansionServiceTest.java
```diff
@@ -188,6 +214,63 @@ private static UserInstance sampleSender() {
...
 
+    private void assertThatExpandedDoiRequestContainsOnlyDoiRequestMessagesAndNotAnyOfTheSupportMessages(
+        PublicationWithDoiRequestAndAllTypesOfMessages samplePublication,
+        ExpandedDoiRequest expandedDoiRequestConversation) {
+        var messagesIdentifiersInExpandedDoiRequest =
+            extractMessageIdenfiersFromExpandedDoiRequest(expandedDoiRequestConversation);
+
+        assertThat(messagesIdentifiersInExpandedDoiRequest,
+                   contains(samplePublication.getDoiRequestMessageIdentifiers()));
+        assertThat(Arrays.asList(samplePublication.getSupportMessageIdentifiers()),
```
axthosarouris, line 225: Unfortunately, hamcrest does not have an easy way to assert that the actual collection does not have any of the items of the non-expected collection. So we have to invert the test, assert that none of the non-desired items are in the actual collection.

expansion/src/test/java/no/unit/nva/expansion/FakeResourceExpansionService.java
```diff
@@ -8,7 +8,7 @@
 public class FakeResourceExpansionService extends ResourceExpansionServiceImpl {
 
     public FakeResourceExpansionService() {
-        super(null, null, null);
+        super(null, null, null,null);
```
brinxmat, line 11: lint

brinxmat (changes requested):

brinxmat (approved):

PR #722, opened 2022-04-22
remove dependency from conventions file

axthosarouris:


publication-rest/build.gradle
```diff
@@ -15,18 +15,13 @@ dependencies {
         exclude group: 'com.github.bibsysdev', module: 'nva-commons'
     }
 
-    implementation(libs.nva.datamodel.core) {
```
axthosarouris, line 18: duplicate dependency

PR #727, opened 2022-05-16
Dynamo db stream to event bridge handler stores data in s3

axthosarouris:


publication-event-handlers/src/main/java/no/unit/nva/publication/events/handlers/dynamodbstream/DynamodbStreamToEventBridgeHandler.java
```diff
@@ -19,25 +28,35 @@
...
-    protected DynamodbStreamToEventBridgeHandler(EventPublisher eventPublisher) {
+    protected DynamodbStreamToEventBridgeHandler(S3Client s3Client, EventPublisher eventPublisher) {
         this.eventPublisher = eventPublisher;
+        this.s3Client = s3Client;
     }
 
     @Override
-    public Void handleRequest(DynamodbEvent event, Context context) {
+    public URI handleRequest(DynamodbEvent event, Context context) {
+        var savedFile = storeEventInS3(event);
```
axthosarouris, line 52: try to save the file but do not fail if cannot save the file. (Preserve current functionality)

template.yaml
```diff
@@ -41,7 +41,7 @@ Parameters:
   NewNvaEventBusName:
     Type: 'AWS::SSM::Parameter::Value<String>'
     Default: '/NVA/Events/EventsBusName'
-  NewNvaEventsBucketsName:
+  NvaEventsBucketsName:
```
axthosarouris, line 44: There is no "new" nva event bucket name

PR #809, opened 2022-09-05
Add general support

axthosarouris:

PR #810, opened 2022-09-05
Add general support-restore doi request module

axthosarouris:
Depends on PR #809


publication-commons/src/main/java/no/unit/nva/publication/service/impl/TicketService.java
```diff
@@ -132,6 +132,7 @@ protected TicketEntry completeTicket(TicketEntry ticketEntry) throws ApiGatewayE
         var publication = resourceService.getPublicationByIdentifier(ticketEntry.getResourceIdentifier());
         var existingTicket =
             attempt(() -> fetchTicketByIdentifier(ticketEntry.getIdentifier()))
+                .or(() -> fetchByResourceIdentifierForLegacyDoiRequestsAndPublishingRequests(ticketEntry))
```
axthosarouris, line 135: Current doi request handlers do not support ticket identifiers and work based on the fact that doi requests are unique for each publication.

brinxmat (approved):

PR #816, opened 2022-09-05
Add general support restore ticket api

axthosarouris:


publishing-request/src/main/java/no/unit/nva/publication/publishingrequest/GeneralSupportRequestDto.java
```diff
@@ -0,0 +1,147 @@
...
+                                    @JsonProperty(MESSAGES_FIELD) List<MessageDto> messages) {
+        super();
+        this.status = status;
+        this.createdDate = createdDate;
+        this.modifiedDate = modifiedDate;
+        this.version = version;
+        this.identifier = identifier;
+        this.publicationId = publicationId;
+        this.id = id;
+        this.messages = messages;
```
brinxmat, line 66: Honestly, it isn't important right now, but I feel the organisation of the signature here doesn't reflect what is important in the creation of these objects. Is there are reason to expose the version and identifier here?

brinxmat (approved):

publishing-request/src/main/java/no/unit/nva/publication/publishingrequest/GeneralSupportRequestDto.java
axthosarouris: version yes, identifier no

PR #859, opened 2022-10-14
only linting

axthosarouris:


doi-requests/src/main/java/no/unit/nva/doirequest/create/CreateDoiRequestHandler.java
```diff
@@ -84,6 +83,12 @@ protected Integer getSuccessStatusCode(CreateDoiRequest input, Void output) {
         return HttpURLConnection.HTTP_CREATED;
     }
     
+    private static boolean doiRequestSatisfiesBasicRuleProtectingUsFromCreatingLargeAmountsOfHangingDraftDoiRequests(
+        DoiRequest ticket) {
+        //TODO replace with Duration.ofDays(1)
```
ketilaa, line 88: Did you plan to fix this now?

PR #861, opened 2022-10-17
Bump action versions

brinxmat:
- bump the action checkout v2 -> v3


anetteOlli (approved):
Didn't know there was different versions of this.

brinxmat:
> 

Your builds start failing with a cryptic-ish warning about Node JS, which is why I'm being a bit demonstrative with the old everyone-has-to-review thing here :D

torarnet (approved):

PR #896, opened 2022-10-26
Add event based conversion to schema representation

brinxmat:
Adds schema.org mapping capability to publication api


axthosarouris (approved):

brinxmat:
Superceded since we needed to adjust things in data model

PR #898, opened 2022-10-27
Refactor resource service

axthosarouris:


publication-commons/src/main/java/no/unit/nva/publication/model/business/Resource.java
```diff
@@ -122,6 +127,10 @@ public static ResourceBuilder builder() {
         return new ResourceBuilder();
     }
     
+    public Publication persistNew(ResourceService resourceService, UserInstance userInstance) {
+        return resourceService.resourceCallsCreatePublication(userInstance, this.toPublication());
```
axthosarouris, line 131: The idea is that all actions should start from the Resource (publication-api's version of publication) in the following way:

```
var newResource = Resource.fromPublication(publicaton).persistNew();
var fetchPersistedResource = Resource.queryObject(resourceIdentifier).fetch();
...
```

brinxmat (approved):

PR #902, opened 2022-10-28
rename publication creation method

axthosarouris:


publication-commons/src/main/java/no/unit/nva/publication/service/impl/ResourceService.java
```diff
@@ -115,13 +114,6 @@ public Publication resourceCallsCreatePublication(UserInstance userInstance, Pub
         return insertResource(newResource);
     }
     
-    public Publication createPublicationWithPredefinedCreationDate(Publication inputData) {
```
axthosarouris, line 118: this method is not used in reality

brinxmat (approved):

ketilaa:
@brinxmat Hva gjør vi med denne?

PR #906, opened 2022-11-03
Observe accept header

brinxmat:
- Adds schema.org conversion for *ALL* publication types

Can be tested with: `curl -H "Accept: application/vnd.schemaorg.ld+json" https://api.sandbox.nva.aws.unit.no/publication/018418918ebe-c1819fe8-29d7-4ded-83fa-ec8fa2898357`


cristin-import/src/test/java/no/unit/nva/cristin/CristinDataGenerator.java
```diff
@@ -1,7 +1,6 @@
 package no.unit.nva.cristin;
```
torarnet, line 1: Why are you deleting all this code? Is it unused?

torarnet (approved):
Approved, but I do not have much knowledge of some of the schema types.

ketilaa (approved):
Some of the same observations as @torarnet. I assume you have removed unused code, and I trust your knowledge on the schema.org and json-ld stuff :)

ketilaa (approved):

PR #912, opened 2022-11-21
Np 17940 setup brage migration pipeline

anetteOlli:


brage-import/build.gradle
```diff
@@ -0,0 +1,27 @@
+apply plugin: 'io.freefair.lombok'
```
brinxmat, line 1: Unused

brage-import/build.gradle
```diff
@@ -0,0 +1,27 @@
...
+
+dependencies {
+
+    testImplementation project(":publication-testing")
+    implementation project(":s3-import-commons")
```
brinxmat, line 6: Unused

brage-import/build.gradle
```diff
@@ -0,0 +1,27 @@
...
+
+
+    testImplementation libs.nva.testutils
+
+    implementation libs.aws.sdk2.eventbridge
```
brinxmat, line 11: Unused

brage-import/build.gradle
```diff
@@ -0,0 +1,27 @@
...
+    implementation libs.aws.sdk2.s3
+    implementation libs.aws.lambda.java.core
+    implementation libs.aws.lambda.events
+
+    implementation libs.nva.eventhandlers
```
brinxmat, line 18: Unused

brage-import/src/main/java/no/sikt/nva/brage/migration/BrageImportConfig.java
```diff
@@ -0,0 +1,8 @@
+package no.sikt.nva.brage.migration;
+
+import nva.commons.core.JacocoGenerated;
+
+@JacocoGenerated
```
brinxmat, line 5: The class is unused, which likely means it doesn't need to be in the PR. Generally, I'm against annotating classes like this since I generally forget to remove them when I do have coverage for them, which is a bit misleading

template.yaml
```diff
@@ -622,6 +626,27 @@ Resources:
...
+      Handler: no.sikt.nva.brage.migration.lambda.BrageEntryEventConsumer::handleRequest
+      Runtime: java11
+      ReservedConcurrentExecutions: 1
+      Role: !GetAtt LambdaRole.Arn
+      Events:
+        BrageMigrationEvent:
+          Type: S3
+          Properties:
+            Bucket: !Ref BrageMigrationInputBucket
+            Events: 's3:ObjectCreated:*'
```
brinxmat, line 642: Maybe we only want new objects? Not sure what the intention is

brinxmat (changes requested):

brage-import/build.gradle
anetteOlli: fixed

brage-import/build.gradle
anetteOlli: fixed

brage-import/build.gradle
anetteOlli: fixed

brage-import/build.gradle
anetteOlli: fixed

template.yaml
anetteOlli: fixed

PR #917, opened 2022-11-25
Np 19713 use is publishable from model

brinxmat:
- Removes Publication validation code from service, delegating it to a method on the model
- Removes fine-grained exception since this is difficult to achieve, once the full validation is in place, we can return to a more fine-grained report

### DIFF
--- a/publication-commons/src/main/java/no/unit/nva/publication/model/business/PublishingRequestCase.java
+++ b/publication-commons/src/main/java/no/unit/nva/publication/model/business/PublishingRequestCase.java
@@ -1,6 +1,5 @@
 package no.unit.nva.publication.model.business;
 
-import static java.util.Objects.isNull;
 import static no.unit.nva.publication.model.business.TicketEntry.Constants.CREATED_DATE_FIELD;
 import static no.unit.nva.publication.model.business.TicketEntry.Constants.CUSTOMER_ID_FIELD;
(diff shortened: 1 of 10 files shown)


publication-commons/src/main/java/no/unit/nva/publication/model/business/PublishingRequestCase.java
```diff
@@ -34,16 +29,15 @@
 @SuppressWarnings("PMD.GodClass")
 public class PublishingRequestCase extends TicketEntry {
     
-    public static final String RESOURCE_WITHOUT_MAIN_TITLE_ERROR = "Resource is missing main title: ";
+    public static final String RESOURCE_LACKS_REQUIRED_DATA = "Resource does not have required data to be "
```
torarnet, line 32: Can you use this in tests instead of writing the same string var everywhere?

torarnet (approved):

publication-commons/src/main/java/no/unit/nva/publication/model/business/PublishingRequestCase.java
brinxmat: This was deliberate, I find it a bit better to state the concern before I update the class — it decouples the class from the test, also.

ketilaa:
Will it affect the frontend error handling when you remove the fine-grained error messages temporarily?

PR #951, opened 2022-12-16
expand affiliations

fredrikpf:


expansion/src/test/java/no/unit/nva/publication/indexing/AffiliationGenerator.java
```diff
@@ -0,0 +1,68 @@
...
+        String parentSample = stringFromResources(Path.of(TOP_LEVEL_SAMPLE_JSON_FILENAME));
+        JsonNode parentJson = attempt(() -> objectMapper.readTree(parentSample)).orElseThrow();
+
+        int i = 0;
+        for (URI childURI : childUris) {
+            JsonNode content = updateFieldsForChilAffiliation(parentURI, childURI);
+
+            uriToContent.put(childURI, content.toString());
+
+            setChildrenForParentAffiliation(parentJson, i, childURI);
```
sondrev, line 34: Tip: You can increase "i" inline while accessing it.
`setChildrenForParentAffiliation(parentJson, i,++ childURI);`
and then you dont have to increase it in the next line

sondrev (changes requested):
There is alot of whitesapce changes. I suspect you might have different settings for newline-character than the others working on this project. Please check which is correct

expansion/build.gradle
```diff
@@ -10,6 +10,7 @@ dependencies {
     implementation libs.jsonld
     implementation libs.guava
     implementation libs.aws.sdk.dynamodb
+    implementation libs.bundles.jena
```
sondrev, line 13: Is this needed? I removed it and get no errors

expansion/src/main/java/no/unit/nva/expansion/model/IndexDocumentWrapperLinkedData.java
```diff
@@ -10,44 +11,73 @@
...
         inputStreams.addAll(fetchAll(extractPublicationContextUris(indexDocument)));
+        inputStreams.addAll(fetchAll(extractAffiliationUris(indexDocument)));
         return inputStreams;
     }
-    
+
     private Collection<? extends InputStream> fetchAll(List<URI> publicationContextUris) {
-        List<Optional<String>> uriContent =
-            publicationContextUris.stream().map(this::fetch).collect(Collectors.toList());
+        List<Optional<String>> uriContent = new ArrayList<>(Collections.emptyList());
```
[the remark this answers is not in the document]
ketilaa: I agree. And if you actually needed an empty list, use `List<String> emptyList = Collections.emptyList();`

expansion/src/main/java/no/unit/nva/expansion/model/IndexDocumentWrapperLinkedData.java
```diff
@@ -10,44 +11,73 @@
...
+
+        try {
+            json = JsonUtils.dtoObjectMapper.readTree(content);
+        } catch (Exception e) {
+            return;
+        }
+        final String organization = "Organization";
+
+
+        if (json.has("type") && organization.equals(json.at("/type").asText())) {
```
ketilaa, line 73: Same here. Help the garbage collector by creating static final fields on the class.

expansion/src/test/java/no/unit/nva/expansion/model/ExpandedDataEntryTest.java
```diff
@@ -66,15 +74,22 @@ public void setup() {
...
     }
-    
+
     @Test
     void shouldReturnExpandedResourceWithoutLossOfInformation() throws JsonProcessingException {
         var publication = createPublicationWithoutDoi();
+
+        publication.getEntityDescription().getContributors().forEach(
+            cont ->
+                cont.getAffiliations().remove(0)
```
ketilaa, line 84: What is going on here? What if getAffiliations() has no elements? Probably not a problem in this test, but is it a reason why you always delete the first affiliation?

ketilaa (changes requested):

expansion/src/test/java/no/unit/nva/publication/indexing/AffiliationGenerator.java
```diff
@@ -0,0 +1,68 @@
...
+    public static final String FIELD_HAS_PART = "/hasPart";
+    public static final String FIELD_FIRST_PART_OF = "/partOf/0";
+
+    public static Map<URI, String> getAffiliationsWithCommonParent(List<URI> childUris, URI parentURI) {
+        Map<URI, String> uriToContent = new java.util.HashMap<>(Collections.emptyMap());
+
+        String parentSample = stringFromResources(Path.of(TOP_LEVEL_SAMPLE_JSON_FILENAME));
+        JsonNode parentJson = attempt(() -> objectMapper.readTree(parentSample)).orElseThrow();
+
+        int i = 0;
```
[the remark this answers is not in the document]
fredrikpf: The alternative is having two loops iterating over the same values. It's not a heavy computational cost but it feels unnecessary...

expansion/src/test/java/no/unit/nva/expansion/model/ExpandedDataEntryTest.java
fredrikpf: Had some trouble when it came to mocking. Found a better solution. Please see reviewed version. 


In case you want insight on what the problem was:
This test assumes we're not expanding any URI's by not providing any URI's in fields that should be expanded. Now that we're expanding affiliations this test fails (especially since the mockdata points to example.org/somestring which returns html, so we get a parsing error). 

The approach i tried here removes the affiliation-field from the testdata entirely, making the test pass. This is not optimal since we're not testing for changes in affiliations. I.E. if the framing for affiliations is wrong the test will still pass.

My new solution will make sure we return an empty JSON (meaning we're still not expanding) instead. This way the test will still test everything it's supposed to test.

expansion/src/test/java/no/unit/nva/expansion/model/ExpandedDataEntryTest.java
```diff
@@ -38,52 +44,61 @@
...
+        when(uriRetriever.getRawContent(any(), any())).thenReturn(Optional.empty());
+
         this.resourceExpansionService =
             new ResourceExpansionServiceImpl(resourceService, ticketService);
+
     }
-    
+
     @Test
+    @Disabled("Fails sometimes dependend on what random data is created by createPublicationWithoutDoi()")
```
fredrikpf, line 83: Lager en jiratask på dette. Så at samme problem ligger i main

ketilaa: Vi har flere bugs rapportert i Jira på at serialiseringstester feiler av og til.

expansion/src/test/java/no/unit/nva/publication/indexing/ExpandedResourceTest.java
```diff
@@ -41,166 +49,210 @@
...
         final JsonNode framedResultNode = indexDocument.asJsonNode();
-        
+
         assertEquals(publisherUri.toString(), framedResultNode.at(PUBLISHER_ID_JSON_PTR).textValue());
         assertEquals(publisherName, framedResultNode.at(PUBLISHER_NAME_JSON_PTR).textValue());
         assertEquals(seriesUri.toString(), framedResultNode.at(SERIES_ID_JSON_PTR).textValue());
         assertEquals(seriesName, framedResultNode.at(SERIES_NAME_JSON_PTR).textValue());
+
+        var hasPart = framedResultNode.findValue("https://example.org/vocab#hasPart");
+        assertNotNull(hasPart);
```
ketilaa, line 97: assertThat(hasPart, CoreMatchers.not(IsEmptyString.emptyOrNullString()));
or with static imports:
assertThat(hasPart, not(emptyOrNullString()));

This one-liner does the same trick. Just for information.

ketilaa (approved):

expansion/src/main/java/no/unit/nva/expansion/model/IndexDocumentWrapperLinkedData.java
```diff
@@ -13,41 +15,67 @@
...
-                   .flatMap(Optional::stream)
-                   .map(IoUtils::stringToStream)
-                   .collect(Collectors.toList());
+            .flatMap(Optional::stream)
+            .map(IoUtils::stringToStream)
+            .collect(Collectors.toList());
     }
-    
+
+    private void fetchChildUrisForOrganizations(List<Optional<String>> uriContent, String content) {
```
sondrev, line 68: Why not let the function take in List<String>. It should be the callers responsibility to remove non-valid fields

sondrev (approved):

PR #955, opened 2022-12-29
support for fundings in Resource*

ketilaa:


tkardu (approved):
The change of NVA model version doesn´t have any other dependencies?

brinxmat (approved):

PR #964, opened 2023-01-04
UnitStyle

anetteOlli:


anetteOlli:
Får hele tiden formaterings endringer i koden i cristin-import modulen. Bruker UnitStyle.

truhacevkir (approved):
I do not get any formatting changes, so it looks good.

PR #967, opened 2023-01-05
Np 27816 cristin import interview media contribution

anetteOlli:


truhacevkir (approved):
ser bra ut

cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinMediumTypeCode.java
```diff
@@ -0,0 +1,34 @@
...
+        this.value = value;
+    }
+
+    public static CristinMediumTypeCode fromValue(String value) {
+        for (CristinMediumTypeCode type : CristinMediumTypeCode.values()) {
+            if (type.getValue().equalsIgnoreCase(value)) {
+                return type;
+            }
+        }
+        throw new IllegalArgumentException();
```
brinxmat, line 26: Generally, I'd use a stream of values() and SingletonCollector, but I don't think it's important

cristin-import/src/main/java/no/unit/nva/cristin/mapper/PublicationInstanceBuilderImpl.java
```diff
@@ -36,7 +37,10 @@ public PublicationInstance<? extends Pages> build() {
             return new ChapterArticleBuilder(cristinObject).build();
         } else if (isEvent(cristinObject)) {
             return new EventBuilder(cristinObject).build();
-        } else if (cristinObject.getMainCategory().isUnknownCategory()) {
+        }else if (isMediaContribution(cristinObject)) {
```
brinxmat, line 40: Formatting

brinxmat (changes requested):

cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinMediumTypeCode.java
anetteOlli: Fikset

cristin-import/src/main/java/no/unit/nva/cristin/mapper/PublicationInstanceBuilderImpl.java
anetteOlli: fikset

brinxmat (approved):

PR #971, opened 2023-01-09
Bump limit on query, adding status tests

brinxmat:
- byOwner results:
  - Bumps limit from DDB default (50) to 100
  - Test that statuses are observed

### DIFF
--- a/publication-commons/src/main/java/no/unit/nva/publication/service/impl/ReadResourceService.java
+++ b/publication-commons/src/main/java/no/unit/nva/publication/service/impl/ReadResourceService.java
@@ -149,7 +149,8 @@ public class ReadResourceService {
         return new QueryRequest()
                    .withTableName(tableName)
                    .withIndexName(BY_CUSTOMER_RESOURCE_INDEX_NAME)
-                   .withKeyConditions(keyConditions);
+                   .withKeyConditions(keyConditions)
+                   .withLimit(100);
     }
     
     private List<Dao> parseResultSetToDaos(QueryResult queryResult) {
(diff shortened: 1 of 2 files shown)


publication-commons/src/main/java/no/unit/nva/publication/service/impl/ReadResourceService.java
```diff
@@ -149,7 +149,8 @@ private QueryRequest queryByResourceIndex(ResourceDao queryObject) {
         return new QueryRequest()
                    .withTableName(tableName)
                    .withIndexName(BY_CUSTOMER_RESOURCE_INDEX_NAME)
-                   .withKeyConditions(keyConditions);
+                   .withKeyConditions(keyConditions)
+                   .withLimit(100);
```
brinxmat, line 153: The default is 50, which is too low right now.

A new PR will split the drafts/published into different queries and allow pagination of results

publication-commons/src/test/java/no/unit/nva/publication/service/impl/ResourceServiceTest.java
```diff
@@ -401,7 +401,17 @@ void getResourcePropagatesExceptionWithWhenGettingResourceFailsForUnknownReason(
         RuntimeException exception = assertThrows(RuntimeException.class, action);
         assertThat(exception.getMessage(), is(equalTo(expectedMessage)));
     }
-    
+
+    @ParameterizedTest(name = "Should return publication by owner when status is {0}")
```
brinxmat, line 405: Added test to ensure that we are doing what we think we are

ketilaa (approved):

PR #995, opened 2023-01-24
add topLevelAffiliation to PublicationAffiliations

fredrikpf:
The models are used may not be ideal, suspect a lot of work is done twice. Work in progress


fredrikpf:
Fails test, but does add topLevelAffiliations

ketilaa (commented):
PR is too large for my liking. No major comments, but I have to rely on @brinxmat to verify the JSON-LD stuff.

brinxmat (approved):

PR #1014, opened 2023-02-01
NP-31714-move-create-scopus, repo moved over

truhacevkir:
- Removed MetadataService that fetches id's from Kanalregister. 
- Commented out some methods in PublicationContextCreator.class that are temporary unused because of deletion of MetadataService. Those methods should be used when new service for search in Kanalregister will be implemented


build.gradle
```diff
@@ -32,3 +32,11 @@ subprojects {
 build {
     dependsOn jacocoRootReport
 }
+
+dependencies {
+    implementation 'com.github.bibsysdev:logutils:1.27.2'
```
anetteOlli, line 37: Denne skal i toml fila

anetteOlli (approved):

PR #1034, opened 2023-02-09
NP-34017-cristin-import-cristin-uris-to-nva-uris, implemented

anetteOlli:


anetteOlli:
Keep in mind that the test-data always comes from cristin-prod. So I was unsure what domain I am supposed to use. I choose to use the same domain as the environment, but that means that the uri's might be faulty.

cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinContributor.java
```diff
@@ -108,7 +110,12 @@ private List<Organization> extractAffiliations() {
 
     private URI constructId() {
         return SHOULD_CREATE_CONTRIBUTOR_ID
-                   ? UriWrapper.fromUri(CRISTIN_PERSONS_URI).addChild(identifier.toString()).getUri()
+                   ?
+                   UriWrapper.fromUri(NVA_API_DOMAIN)
```
brinxmat, line 114: Yep, this _is_ correct :)

brinxmat (commented):
You may have to update to be in sync with main

truhacevkir (approved):

PR #1046, opened 2023-02-17
NP-35375-cristin-import-detect-production-environment, implemented

anetteOlli:


anetteOlli:
According to https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-conditions.html, this is how you write equals in cloudformation yml. Also this is one of the reasons why I hate yaml. 

Reason to why I tagget you Rurik is because I was wondering if I should instead set this as a parameter in the master template upon creation of the stack instead.

brinxmat (approved):

PR #1055, opened 2023-02-22
add @explicit to frame so partOf/hasPart doesn't create infite graph

fredrikpf:


ketilaa (commented):
Looks ok. I leave it up to @brinxmat to verify the expected effect of the change :)

brinxmat (approved):

PR #1064, opened 2023-02-28
now returns 201 rather than 303 in create ticket endpoint to avoid re…

ketilaa:
…direct with missing authorization header in frontend


elkebab:
Kan virke som at dette ikke vil løse problemet helt heller, ref denne: https://stackoverflow.com/a/37931084

publication-rest/src/main/java/no/unit/nva/publication/create/CreatePublicationRequest.java
```diff
@@ -119,25 +122,33 @@ public void setFundings(List<Funding> fundings) {
         this.fundings = fundings;
     }
 
+    public void setRightsHolder(String rightsHolder) {
```
ketilaa, line 125: @brinxmat Should `rightsHolder` be part of `WithMetadata` interface?

ketilaa:
Look at that. All checks have passed ;)

brinxmat (approved):

PR #1071, opened 2023-03-06
NP-37500-DOI-event-validation-before-sending-event

anetteOlli:


anetteOlli:
Dette må rullles ut koordinert desverre. 
Dette vil jo da hindre draftDoiHandleren i å bli trigget. Slik at Kir's lambda må være på plass først.

publication-event-handlers/src/main/java/no/unit/nva/publication/events/handlers/tickets/DoiResourceRequirements.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    private static boolean hasCorrectPublishedStatus(Resource newEntry) {
+        return PublicationStatus.PUBLISHED.equals(newEntry.getStatus())
+               || PublicationStatus.PUBLISHED_METADATA.equals(newEntry.getStatus());
+    }
+
+    private static boolean mandatoryFieldsAreNotNull(Resource newEntry) {
+        return nonNull(newEntry.getIdentifier())
+               && nonNull(newEntry.getPublisher())
```
ketilaa, line 38: Is notEmpty() better that just nonNull? Are empty strings ok?

publication-event-handlers/src/main/java/no/unit/nva/publication/events/handlers/tickets/DoiRequestEventProducer.java
```diff
@@ -75,11 +75,16 @@ protected DoiMetadataUpdateEvent processInputPayload(
                    ? propagateEvent(eventBody)
                    : EMPTY_EVENT;
     }
-    
+
+    private static String readDomainName() {
```
ketilaa, line 79: Personally, I don't like these kind of lookups in the environment. It is hard to see what environment variables a given handler depends on. I would have looked up the domain name in the environment in the default constructor of the handler. But I know the publication api is full of these kind of lookups in the environment, so it is ok to leave it as is too.

ketilaa (approved):

PR #1082, opened 2023-03-23
NP-41153-disallow-doi-request-when-non-published, work in progress

truhacevkir:
This disallows creation of DoiRequest when publication is published, in other case user should use draft doi flow. 
Since PublishingRequest and GeneralSupportRequest can be created for publications with different statuses, I had to change ticketTypeProvider to publishingAndGeneralSupportTicketTypeProvider for almost all the tests that have ticket logic involved.

### DIFF
--- a/expansion/src/test/java/no/unit/nva/expansion/ResourceExpansionServiceTest.java
+++ b/expansion/src/test/java/no/unit/nva/expansion/ResourceExpansionServiceTest.java
@@ -25,16 +25,17 @@ import java.util.Set;
 import java.util.stream.Stream;
 import no.unit.nva.expansion.model.ExpandedResource;
 import no.unit.nva.expansion.model.ExpandedTicket;
-import no.unit.nva.publication.external.services.UriRetriever;
 import no.unit.nva.identifiers.SortableIdentifier;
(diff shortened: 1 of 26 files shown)


expansion/src/test/java/no/unit/nva/expansion/ResourceExpansionServiceTest.java
```diff
@@ -219,7 +231,7 @@ private static List<Class<?>> listPublicationInstanceTypes() {
 
     @SuppressWarnings("unchecked")
     private static Class<? extends TicketEntry> someOtherTicketType(Class<? extends TicketEntry> ticketType) {
-        return (Class<? extends TicketEntry>) ticketTypeProvider().filter(type -> !ticketType.equals(type))
+        return (Class<? extends TicketEntry>) ticketTypeProvider().filter(type -> !ticketType.equals(type) && !type.equals(DoiRequest.class))
```
ketilaa, line 234: Method name does not reflect what the method do. For some reason you are avoiding DoiRequest. Maybe name method someOtherTicketTypeBesidesDoiRequest() ?

PR #1086, opened 2023-03-24
Disallowe doi request

truhacevkir:


tickets-test/src/main/java/publication/test/TicketTestUtils.java
```diff
@@ -0,0 +1,88 @@
+package publication.test;
```
ketilaa, line 1: Terrible choice of package name. no.unit.nva.publication.ticket.test

truhacevkir: Endrer til no.unit.nva.publication.ticket.test

ketilaa (approved):

PR #1090, opened 2023-03-27
added delete publication for an external client

sondrev:


publication-rest/src/main/java/no/unit/nva/publication/delete/DeletePublicationHandler.java
```diff
@@ -33,22 +37,45 @@ public DeletePublicationHandler() {
...
+        if (requestInfo.clientIsThirdParty()) {
+            return createUserInstanceForExternalClientUser(requestInfo);
+        } else {
+            var owner = requestInfo.getNvaUsername();
+            var customerId = requestInfo.getCurrentCustomer();
+            return UserInstance.create(owner, customerId);
+        }
+    }
+
+    private UserInstance createUserInstanceForExternalClientUser(RequestInfo requestInfo) throws UnauthorizedException {
```
ketilaa, line 67: Is this duplicated from CreatePublicationHandler? If so, have you considered a shared util class or even nva-commons?

ketilaa (approved):

PR #1131, opened 2023-04-19
removing PendingPublishingRequestHandler

truhacevkir:
Have added some more tests to satisfy Jacoco


anetteOlli (approved):

publication-commons/src/main/java/no/unit/nva/publication/service/impl/ResourceService.java
```diff
@@ -164,12 +164,6 @@ public PublishPublicationStatusResponse publishPublication(UserInstance userInst
         throws ApiGatewayException {
         return updateResourceService.publishPublication(userInstance, resourceIdentifier);
     }
-
-    public PublishPublicationStatusResponse publishPublicationMetadata(UserInstance userInstance,
```
StigNorland, line 168: This was the task at hand, but where is it now?
If you include link to task in Jira (in the commit message), it is easier to find.

StigNorland (approved):

publication-commons/src/main/java/no/unit/nva/publication/service/impl/ResourceService.java
truhacevkir: All the task are already implemented, this one removes unused handler and function

truhacevkir: https://unit.atlassian.net/jira/software/c/projects/NP/boards/111?modal=detail&selectedIssue=NP-44577&assignee=62a07ba3954f50006fcc2c2a

PR #1140, opened 2023-04-24
https://unit.atlassian.net/browse/NP-44610

StigNorland:


docs/openapi.yaml
```diff
@@ -510,19 +510,146 @@ components:
   schemas:
     Publication:
       type: object
+      description: Complete Publication entity
```
sondrev, line 513: This description needs to be same as in nva-search-api. Currently its "description: Structure returned from successful seach". I suggest also creating a PR for nva-search-api

sondrev (approved):

PR #1142, opened 2023-04-25
Np 44599 publication context handler

monaullah:
https://unit.atlassian.net/browse/NP-44599


ketilaa (commented):
How do we make sure we actually expose a valid context? Only test on response body is string equality with what we actually return in the response. We should probably have a test named something like `shouldExposeValidJsonLdContext` where you actually validate the response somehow. I guess @brinxmat can guide you on that.

publication-rest/src/main/java/no/unit/nva/publication/fetch/FetchPublicationContextHandler.java
```diff
@@ -0,0 +1,49 @@
...
+    private static final String ENV_CUSTOM_DOMAIN_BASE_PATH = "CUSTOM_DOMAIN_BASE_PATH";
+
+    public FetchPublicationContextHandler() {
+        super(Void.class, new Environment());
+    }
+
+    @Override
+    protected String processInput(Void input, RequestInfo requestInfo, Context context) throws ApiGatewayException {
+        var host = environment.readEnv(ENV_API_HOST);
+        var path = environment.readEnv(ENV_CUSTOM_DOMAIN_BASE_PATH);
```
brinxmat, line 29: Use 

```
   import static no.unit.nva.publication.PublicationServiceConfig.ENVIRONMENT;

   …
    public static final String HOST = ENVIRONMENT.readEnv("API_HOST");
    public static final String PATH = ENVIRONMENT.readEnv("CUSTOM_DOMAIN_BASE_PATH");
    …
    var baseUri = UriWrapper.fromHost(HOST)
        .addChild(PATH)
        .getUri();
````

brinxmat (changes requested):

publication-rest/src/main/java/no/unit/nva/publication/fetch/FetchPublicationContextHandler.java
monaullah: Shouldn't the constants be private? See new commit

brinxmat (approved):

PR #1178, opened 2023-05-09
report identifiers of entries failing to be migrated

axthosarouris:


publication-commons/src/main/java/no/unit/nva/publication/service/impl/ResourceService.java
```diff
@@ -192,9 +192,17 @@ public ListingResult<Entity> scanResources(int pageSize, Map<String, AttributeVa
     public void refreshResources(List<Entity> dataEntries) {
         final var refreshedEntries = refreshAndMigrate(dataEntries);
         var writeRequests = createWriteRequestsForBatchJob(refreshedEntries);
-        writeToDynamoInBatches(writeRequests);
+        attempt(()->writeToDynamoInBatches(writeRequests))
```
axthosarouris, line 195: The code needs formatting but I do not dare to do autoformat

publication-commons/src/test/java/no/unit/nva/publication/service/impl/FailingDynamoClient.java
```diff
@@ -0,0 +1,571 @@
+package no.unit.nva.publication.service.impl;
```
axthosarouris, line 1: This class does nothing different except for calling the wrapped client for all methods except for the batchWriteItem

PR #1199, opened 2023-05-16
Np 44679 ticketentry remove publication details

monaullah:
(Sorry for the big PR, I didn't have a good understanding of the scope when I started) Maybe we should go through this IRL together, @brinxmat?
(Btw, using UnitStyle code style, but still a lot of formatting diff)

Publication information on tickets in only necessary in the ExpandedTicket for search (search-api). There is no longer need for persisted (denormalized) PublicationDetails in the database.

- Deleted PublicationDetails from model.
- Removed publicationDetails from TicketEntry
- Removed publicationSummary from TicketDtos, replaced with publicationIdentifier
- Replaced all usages of TicketEntry.extractPublicationIdentifier() with TicketEntry.getResourceIdentifier()


monaullah:
Live code review with @brinxmat, approved

PR #1204, opened 2023-05-22
Np 44682 publiction summary - remove created date and modified date

monaullah:


publication-commons/src/main/java/no/unit/nva/publication/model/PublicationSummary.java
```diff
@@ -39,10 +39,6 @@ public class PublicationSummary {
     @JsonProperty
     private User owner;
     @JsonProperty
-    private Instant createdDate;
-    @JsonProperty
-    private Instant modifiedDate;
-    @JsonProperty
```
anetteOlli, line 45: Side-note: 
Removing modifiedDate MIGHT cause updating datacite xml for findable DOI not happening. Leave it like this, but it's nice if more than just me that knows this. 

(Updating datacite xml for findable DOI, is done really really hacky. If this causes the "updating findable DOI" feature to break, we might have an excuse to implement the feature correctly instead.)

anetteOlli (approved):

publication-commons/src/main/java/no/unit/nva/publication/model/PublicationSummary.java
monaullah: Thanks for info, will follow up with team lead

PR #1212, opened 2023-05-26
openapi spec for GET importcandidate endpoint created

truhacevkir:


docs/openapi.yaml
```diff
@@ -488,6 +488,49 @@ paths:
...
+            application/ld+json:
+              schema:
+                $ref: '#/components/schemas/Publication'
+            application/vnd.datacite.datacite+xml:
+              schema:
+                $ref: '#/components/schemas/PublicationDataCiteXml'
+            application/vnd.schemaorg.ld+json:
+              schema:
+                $ref: '#/components/schemas/PublicationSchemaOrg'
+        '410':
```
anetteOlli, line 527: Not found?

docs/openapi.yaml
```diff
@@ -488,6 +488,49 @@ paths:
...
+          description: Permanently gone
```
anetteOlli, line 528: I don't think we will logically delete stuff, only physically delete candidates.

Resource service uses this response if a publication has status = deleted.

An import candidate doesn't have that status field.

docs/openapi.yaml
```diff
@@ -612,6 +655,134 @@ components:
             type: string
         title:
           type: string
+    ImportCandidate:
```
anetteOlli, line 658: Is it possible to extend ImportCandidate from publication like you do in typescript in open-api? Just wondering

docs/openapi.yaml
```diff
@@ -488,6 +488,49 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Publication'
+            application/ld+json:
+              schema:
+                $ref: '#/components/schemas/Publication'
+            application/vnd.datacite.datacite+xml:
+              schema:
+                $ref: '#/components/schemas/PublicationDataCiteXml'
```
anetteOlli, line 523: We are not going to return DataCiteXml from this endpoint.

docs/openapi.yaml
```diff
@@ -488,6 +488,49 @@ paths:
...
+            application/vnd.schemaorg.ld+json:
+              schema:
+                $ref: '#/components/schemas/PublicationSchemaOrg'
```
anetteOlli, line 526: Ask Rurik if we need to return this content-type

anetteOlli (changes requested):

docs/openapi.yaml
truhacevkir: i will test "allOf" property in openapi scheme, changing it back if it does not work

docs/openapi.yaml
truhacevkir: trenger ikke nei

docs/openapi.yaml
```diff
@@ -488,6 +488,46 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/Problem"
+
+  /import-candidate/{identifier}:
+    get:
+      summary: Fetch Import candidate by Identifier
+      tags:
+        - external
```
sondrev, line 496: You dont need this tag unless you want the call to be documented for externals. 
I also see you  dont have the "secuirty" property. Do you want this to be publicly accessible?

PR #1217, opened 2023-05-31
Np 44777 post import candidate to resource table

truhacevkir:


publication-commons/src/main/java/no/unit/nva/publication/service/impl/ResourceService.java
```diff
@@ -182,6 +198,19 @@ public PublishPublicationStatusResponse publishPublication(UserInstance userInst
         return updateResourceService.publishPublication(userInstance, resourceIdentifier);
     }
 
+    public Publication autoImportPublication(Publication inputData) {
```
truhacevkir, line 201: Not sure if it is better to take in ImportCandidate as variable, so none in future uses this method to create publication.

publication-commons/src/main/java/no/unit/nva/publication/model/business/ImportCandidate.java
```diff
@@ -26,7 +26,8 @@
 public class ImportCandidate extends Publication implements JsonSerializable {
 
     public static final String TYPE = "ImportCandidate";
-    @JsonProperty("importStatus")
+    public static final String IMPORT_STATUS = "importStatus";
```
StigNorland, line 29: when IMPORT_STATUS == "importStatus" == fieldName
You are just adding text without any purpose, i.e. your code gets longer without any benefits.

It is a pattern that is repeated all over, so I will not ask you to change it, but maybe you should ask why you are doing this?

publication-commons/src/main/java/no/unit/nva/publication/service/impl/UpdateResourceService.java
```diff
@@ -137,6 +139,17 @@ DeletePublicationStatusResponse updatePublishedStatusToDeleted(SortableIdentifie
         }
     }
 
+    ImportCandidate updateStatus(SortableIdentifier identifier, ImportStatus status)
+        throws NotFoundException {
+        var importCandidate = readResourceService.getResourceByIdentifier(identifier).toImportCandidate();
```
StigNorland, line 144: is `readResourceService` a readonly service?

"resource" is a very generic name, does it hold anything?

publication-commons/src/test/java/no/unit/nva/publication/model/business/DoiRequestTest.java
```diff
@@ -12,8 +12,10 @@
 import java.time.Clock;
 import java.time.Instant;
 import java.time.ZoneId;
+import no.unit.nva.model.role.Role;
```
StigNorland, line 15: Is this in use?

publication-commons/src/test/java/no/unit/nva/publication/model/business/DoiRequestTest.java
```diff
@@ -12,8 +12,10 @@
...
 import no.unit.nva.model.testing.PublicationGenerator;
 import no.unit.nva.publication.storage.model.exceptions.IllegalDoiRequestUpdate;
+import nva.commons.apigateway.AccessRight;
```
StigNorland, line 18: Is this in use?

publication-commons/src/main/java/no/unit/nva/publication/service/impl/TicketService.java
```diff
@@ -49,6 +51,7 @@ protected TicketService(AmazonDynamoDB client,
                             Supplier<SortableIdentifier> identifierProvider) {
         super(client);
         this.identifierProvider = identifierProvider;
+        tableName = RESOURCES_TABLE_NAME;
```
StigNorland, line 54: Why not reference constant directly?

publication-rest/build.gradle
```diff
@@ -58,5 +58,7 @@ test {
     environment "EXTERNAL_USER_POOL_URI", "https://external-user-pool.example.com"
     environment "API_HOST", "https://api-host.example.com"
     environment "CUSTOM_DOMAIN_BASE_PATH", "publication"
+    environment "IMPORT_CANDIDATES_TABLE", "import-candidates-table"
```
StigNorland, line 61: Should table names be configurable from environment?
What is the use case?

publication-rest/src/main/java/no/unit/nva/publication/create/CreatePublicationFromImportCandidateHandler.java
```diff
@@ -0,0 +1,79 @@
...
+                                                       ResourceService publicationService) {
+        super(ImportCandidate.class);
+        this.importCandidatesService = importCandidateService;
+        this.publicationService = publicationService;
+    }
+
+    @Override
+    protected PublicationResponse processInput(ImportCandidate input, RequestInfo requestInfo, Context context)
+        throws ApiGatewayException {
+        validateAccessRight(requestInfo);
```
StigNorland, line 45: Good!

publication-rest/src/main/java/no/unit/nva/publication/create/CreatePublicationFromImportCandidateHandler.java
```diff
@@ -0,0 +1,79 @@
...
+import nva.commons.core.JacocoGenerated;
+
+public class CreatePublicationFromImportCandidateHandler extends ApiGatewayHandler<ImportCandidate,
+                                                                                      PublicationResponse> {
+
+    public static final String IMPORT_CANDIDATES_TABLE =
+        new Environment().readEnv("IMPORT_CANDIDATES_TABLE");
+    public static final String ROLLBACK_WENT_WRONG_MESSAGE = "Rollback went wrong";
+    public static final String IMPORT_PROCESS_WENT_WRONG = "Import process went wrong";
+    private final ResourceService importCandidatesService;
```
StigNorland, line 27: Instead of having two instances of `ResourceService `, maybe it is better to just have all the work in the service it self?

publication-rest/src/test/java/no/unit/nva/publication/create/CreatePublicationFromImportCandidateHandlerTest.java
```diff
@@ -0,0 +1,245 @@
...
+import nva.commons.apigateway.exceptions.NotFoundException;
+import nva.commons.core.Environment;
+import org.apache.hc.core5.http.ContentType;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.zalando.problem.Problem;
+
+public class CreatePublicationFromImportCandidateHandlerTest extends ResourcesLocalTest {
+
+    public static final String PUBLICATIONS_TABLE = new Environment().readEnv("TABLE_NAME");
```
StigNorland, line 65: Are there any scenarios where two instances of this test, will have different table names?

publication-rest/src/test/java/no/unit/nva/publication/create/CreatePublicationFromImportCandidateHandlerTest.java
```diff
@@ -0,0 +1,245 @@
...
+    private ResourceService publicationService;
+    private CreatePublicationFromImportCandidateHandler handler;
+
+    @BeforeEach
+    public void setUp() {
+        super.init(IMPORT_CANDIDATES_TABLE, PUBLICATIONS_TABLE);
+        Environment environment = mock(Environment.class);
+        when(environment.readEnv(ALLOWED_ORIGIN_ENV)).thenReturn("*");
+        importCandidateService = new ResourceService(client, IMPORT_CANDIDATES_TABLE);
+        publicationService = new ResourceService(client, Clock.systemDefaultZone(), SortableIdentifier::next);
```
StigNorland, line 78: `ResourceService` is "generic" and you can supply a table name, which make me think that ResourceService is a generic interface with create, update, fetch & delete. 
however  getPublicationByIdentifier & getImportCandidateByIdentifier are very specific, and I guess are doing very specific work?

StigNorland (changes requested):

docs/openapi.yaml
```diff
@@ -525,6 +525,47 @@ paths:
...
+            schema:
+              $ref: '#/components/schemas/ImportCandidate'
+      x-amazon-apigateway-integration:
+        uri:
+          Fn::Sub: arn:aws:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${CreatePublicationFromImportCandidateFunction.Arn}:live/invocations
+        responses: { }
+        httpMethod: POST
+        type: AWS_PROXY
+      responses:
+        '200':
```
anetteOlli, line 551: IMO you are missing some error codes:

If you try to POST a publication that is not exist as an import-candidates. ex. the post has a scopus-identifier, but the publication identifier does not match in the database. Bad Request.

The publication is missing a scopus identifier. Bad Request.

You are trying to import something that already has been imported: Error code or 200 code?

Also the appropriate status code for for  a POST method is 201

docs/openapi.yaml
```diff
@@ -652,10 +693,126 @@ components:
           type: string
     ImportCandidate:
       type: object
-      description: Complete ImportCandidate entity
+      description: Complete Import candidate entity
       properties:
```
anetteOlli, line 697: allof: does not work?

expansion/src/main/java/no/unit/nva/expansion/model/ExpandedImportCandidate.java
```diff
@@ -313,6 +314,7 @@ private static List<Organization> extractOrganizations(ImportCandidate importCan
         return importCandidate.getEntityDescription().getContributors().stream()
                    .map(Contributor::getAffiliations)
                    .flatMap(List::stream)
+                   .filter(organization -> nonNull(organization.getId()))
```
anetteOlli, line 317: Not sure if all organizations in the import list has an identifier. Ask Terje.

anetteOlli: Ex. People working in foreign universities.

publication-commons/src/main/java/no/unit/nva/publication/model/storage/DynamoEntryByIdentifier.java
```diff
@@ -17,54 +16,54 @@
...
-    
+
     SortableIdentifier getIdentifier();
-    
+
     Entity getData();
-    
-    default Dao fetchByIdentifier(AmazonDynamoDB client) throws NotFoundException {
+
+    default Dao fetchByIdentifier(AmazonDynamoDB client, String tableName) throws NotFoundException {
```
anetteOlli, line 34: I prefer if we pass the table name in the constructor if that is possible. It is only the import-candidates lambdas that operates with several tables, and I think it's okay if you have two resourceServices you operate with. Like you would if you are reading and writing to two different s3 buckets you would need two s3Drivers.

publication-commons/src/main/java/no/unit/nva/publication/service/impl/UpdateResourceService.java
truhacevkir: it can only read, yes

publication-commons/src/main/java/no/unit/nva/publication/service/impl/ResourceService.java
anetteOlli: Yes, do that.

publication-rest/src/main/java/no/unit/nva/publication/create/CreatePublicationFromImportCandidateHandler.java
truhacevkir: Each resourceService is connected to each table, publicationService operates on resource-table and importCandidateService operates on import-candidate-service.

publication-rest/src/test/java/no/unit/nva/publication/create/CreatePublicationFromImportCandidateHandlerTest.java
truhacevkir: You are right, we "reuse" resourceService to make operation on different table, I see no reason to implement new ResourceService that will operate on import-candidate table only.

publication-rest/src/test/java/no/unit/nva/publication/create/CreatePublicationFromImportCandidateHandlerTest.java
truhacevkir: No, we are reading two different table names from environment, one for resources and one for import-candidates

publication-rest/build.gradle
truhacevkir: We read table name from environment to make ResourceService operate on specific table.

anetteOlli: How else would you pass in the table names?

Ex. "nva-import-candidates-master-pipelines-NvaPublicationApiPipeline-5W08RY2DNRIO-nva-publication-api" and  "nva-resources-master-pipelines-NvaPublicationApiPipeline-5W08RY2DNRIO-nva-publication-api" for 4 different aws accounts?

5W08RY2DNRIO is the stack name

publication-rest/src/main/java/no/unit/nva/publication/create/CreatePublicationFromImportCandidateHandler.java
anetteOlli: I think that sounds like a bad idea. This lambda is working with two different tables. You don't want to accidentally push Import-candidates to NVA resource table or NVA publications into Import-candidate table.

publication-rest/src/main/java/no/unit/nva/publication/create/CreatePublicationFromImportCandidateHandler.java
```diff
@@ -0,0 +1,79 @@
...
+    @Override
+    protected PublicationResponse processInput(ImportCandidate input, RequestInfo requestInfo, Context context)
+        throws ApiGatewayException {
+        validateAccessRight(requestInfo);
+        var identifier = input.getIdentifier();
+        return attempt(() -> importCandidatesService.updateImportStatus(identifier, ImportStatus.IMPORTED))
+                   .map(ImportCandidate::toPublication)
+                   .map(publicationService::autoImportPublication)
+                   .map(PublicationResponse::fromPublication)
+                   .orElseThrow(failure -> rollbackAndThrowException(input));
```
anetteOlli, line 51: nice

publication-rest/src/test/java/no/unit/nva/publication/create/CreatePublicationFromImportCandidateHandlerTest.java
anetteOlli: Yeah, I am wondering this myself. Is it because you loose the Import-status attribute from the object so that a method ImportPublication fromResource(Resource resource) method does not work?

template.yaml
```diff
@@ -828,6 +828,29 @@ Resources:
...
+      Handler: no.unit.nva.publication.create.CreatePublicationFromImportCandidateHandler::handleRequest
+      Runtime: java11
+      AutoPublishAlias: live
+      DeploymentPreference:
+        Type: AllAtOnce
+      Environment:
+        Variables:
+          ALLOWED_ORIGIN: '*'
+          IMPORT_CANDIDATES_TABLE: !Ref NvaImportCandidatesTable
+          TABLE_NAME: !Ref NvaResourcesTable
```
anetteOlli, line 844: Could TABLE_NAME be called NVA_RESOURCE_TABLE_NAME=

anetteOlli (changes requested):

docs/openapi.yaml
truhacevkir: i discussed it with Kjetil, allOf make swagger object look weird

template.yaml
truhacevkir: When we instantiating ResourceService without passing in table name, it creates "default" ResourceService that operates in resource table. But i can rename it to make it clear yes, and pass this environment variable in to ResourceService constructor.

publication-rest/src/main/java/no/unit/nva/publication/create/CreatePublicationFromImportCandidateHandler.java
StigNorland: Since you have two instances of ResourceService each to a different table, and they are "generic"
maybe `publicationService `and `candidateService` are better names? 
Unless the table you are accessing is called import-candidate, but then I would expect there to be a candidate table as well.

publication-rest/src/test/java/no/unit/nva/publication/create/CreatePublicationFromImportCandidateHandlerTest.java
StigNorland: Hmm, I am asking why you are reading this from the environment,

publication-rest/src/test/java/no/unit/nva/publication/create/CreatePublicationFromImportCandidateHandlerTest.java
StigNorland: Then why have specific calls to specific tables inside the "generic" service?
A generic service should then just be doing simple IO from and to storage  (`create`, `update`, `fetch `& `delete`.)

docs/openapi.yaml
```diff
@@ -525,6 +525,47 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Publication'
+        '401':
+          description: Unauthorized
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Publication'
```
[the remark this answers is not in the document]
truhacevkir: Right, thank you

publication-commons/src/test/java/no/unit/nva/publication/model/business/DoiRequestTest.java
truhacevkir: nice, thank you

publication-rest/src/main/java/no/unit/nva/publication/create/CreatePublicationFromImportCandidateHandler.java
truhacevkir: What about resourceService and importCandidateService, since one operates on resource-table and second one on import-candidate-service? At the same type Resource is an object we store in database, and both publication and importCandidate are resources at some point.

publication-commons/src/main/java/no/unit/nva/publication/model/storage/DynamoEntryByIdentifier.java
truhacevkir: This ones is passed through resource service constructor, but this method is used by daos.

expansion/src/main/java/no/unit/nva/expansion/model/ExpandedImportCandidate.java
truhacevkir: As I can see sentralImport is listing only Norwegian institutions, if we list all the institution we are going to end up with very long list.

publication-commons/src/main/java/no/unit/nva/publication/service/impl/TicketService.java
truhacevkir: To send not default table name to this method: fetchByIdentifier
 which is used in all database services to access resources from database

publication-rest/src/test/java/no/unit/nva/publication/create/CreatePublicationFromImportCandidateHandlerTest.java
anetteOlli: Me and Kir are in a discussion of removing everything related to import-candidates to a different repository. 
For now we just have to live with this. 

If we see loads more of "we need to make special implementation in resourceService", I think it's time to pack "import-candidates bags", and move it to a seperate github repo. (we need to make PO prioritize a non-functional task).

@StigNorland (TLDR at the end) The reason for why import-candidates are in publication-api repo.
It was because previously the workflow was just two lambdas, sitting in it's own repo, sending thousands of events to publication-api with requests for publication-api to push publications into the resource table. It worked exactly as great as the event flow does everwhere else in NVA. A couple of months ago we refactored to make import-candidates push to it's own table, as NVA don't need to know about them before they get status imported.
TLDR: Historical stupid reasons.

anetteOlli (approved):

PR #1243, opened 2023-06-15
license -> uri

truhacevkir:


brage-import/src/main/java/no/sikt/nva/brage/migration/record/license/NvaLicenseUri.java
```diff
@@ -1,29 +0,0 @@
-package no.sikt.nva.brage.migration.record.license;
-
-import com.fasterxml.jackson.annotation.JsonValue;
-import java.net.URI;
-import nva.commons.core.JacocoGenerated;
-
-public enum NvaLicenseUri {
-    CC_BY(URI.create("https://creativecommons.org/licenses/by/4.0")),
```
anetteOlli, line 8: Replace these enums with some validation of the url? domain name, path, and let version field be whatever?

anetteOlli (changes requested):

anetteOlli (approved):

PR #1251, opened 2023-06-23
NP-44911-resource-expansion-should-use-new-cristin-api

anetteOlli:


expansion/src/test/java/no/unit/nva/expansion/ResourceExpansionServiceTest.java
```diff
@@ -592,7 +592,7 @@ private void initializeServices() {
         resourceService = new ResourceService(client, CLOCK);
         messageService = new MessageService(client);
         ticketService = new TicketService(client);
-        uriRetriever = new UriRetriever();
```
anetteOlli, line 595: It can't be right that we should use a default http-client and do calls to the internet?

expansion/src/main/java/no/unit/nva/expansion/model/IndexDocumentWrapperLinkedData.java
```diff
@@ -113,4 +114,8 @@ private Stream<String> fetchContentRecursively(URI uri) {
     private Optional<String> fetch(URI externalReference) {
         return uriRetriever.getRawContent(externalReference, APPLICATION_JSON_LD.toString());
     }
+
+    private Optional<String> fetchOrganizations(URI externalReference) {
+        return uriRetriever.getRawContent(externalReference, APPLICATION_JSON_LD.toString() + CRISTIN_VERSION);
+    }
```
anetteOlli, line 120: Not sure what we are expecting here, as I can't find a single mock response that tells me what is expected from this.

Maybe I am just not finding the mock responses?

expansion/src/main/java/no/unit/nva/expansion/model/IndexDocumentWrapperLinkedData.java
```diff
@@ -113,4 +114,8 @@ private Stream<String> fetchContentRecursively(URI uri) {
     private Optional<String> fetch(URI externalReference) {
         return uriRetriever.getRawContent(externalReference, APPLICATION_JSON_LD.toString());
     }
+
+    private Optional<String> fetchOrganizations(URI externalReference) {
+        return uriRetriever.getRawContent(externalReference, APPLICATION_JSON_LD.toString() + CRISTIN_VERSION);
```
torarnet, line 119: do you need a `; `between json ld and version? can you write a test asserting accept header is correct? Why is the caller method called a name containing recursively? does it do recursion here or is it just a name?

torarnet (commented):
See comments

expansion/src/main/java/no/unit/nva/expansion/model/IndexDocumentWrapperLinkedData.java
anetteOlli: "Why is the caller method called a name containing recursively? does it do recursion here or is it just a name?" No idea, but if you look closely it is recursive.

anetteOlli: "do you need a ; between json ld and version? can you write a test asserting accept header is correct?" Will try to write test for this.

torarnet: just wondering

torarnet (approved):

monaullah (approved):

expansion/src/test/java/no/unit/nva/expansion/ResourceExpansionServiceTest.java
brinxmat: It isn't right :) so a good catch. Typically, we use URI in with example.org, which returns HTML and forces you to think…this annoys some people, but it is good practice since the fact you get HTML breaks your test even when you've messed up like this.

expansion/src/test/java/no/unit/nva/publication/indexing/ExpandedResourceTest.java
```diff
@@ -65,26 +69,111 @@
...
+    }
+
+    private static void addSeriesToMockUriRetriever(UriRetriever uriRetriever, URI seriesId) throws IOException {
+
+        var publicationChannelSampleSeries = getPublicationChannelSampleSeries(seriesId, randomString());
+        mockGetRawContentResponse(uriRetriever, seriesId, publicationChannelSampleSeries);
+    }
+
+    private static void mockGetRawContentResponse(UriRetriever uriRetriever, URI uri, String response) {
+        when(uriRetriever.getRawContent(eq(uri), any()))
```
brinxmat, line 118: `when(uriRetriever.getRawContent(eq(uri), anyString()))`

expansion/src/test/java/no/unit/nva/publication/indexing/ExpandedResourceTest.java
```diff
@@ -65,26 +69,111 @@
...
+
+    private static void addPublicationChannelPublisherToMockUriRetriever(UriRetriever mockUriRetriever,
+                                                                         URI journalId,
+                                                                         String journalName,
+                                                                         URI publisherId,
+                                                                         String publisherName)
+            throws IOException {
+        String publicationChannelSampleJournal = getPublicationChannelSampleJournal(journalId, journalName);
+        mockGetRawContentResponse(mockUriRetriever, journalId, publicationChannelSampleJournal);
+        String publicationChannelSamplePublisher = getPublicationChannelSamplePublisher(publisherId, publisherName);
```
brinxmat, line 163: `var`

brinxmat (changes requested):

expansion/src/test/java/no/unit/nva/publication/indexing/ExpandedResourceTest.java
anetteOlli: Fikset

expansion/src/test/java/no/unit/nva/publication/indexing/ExpandedResourceTest.java
anetteOlli: Fikset

brinxmat (approved):

PR #1263, opened 2023-07-05
NP-44727-autogenerated-cristin-classes

anetteOlli:


cristin-datamodel-generated/src/main/resources/schema/cristinResult.json
```diff
@@ -0,0 +1,469 @@
+{
```
brinxmat, line 1: Do we want this file, or .gitignore it?

brinxmat (changes requested):

ketilaa:
I guess we abandoned this due to obvious weaknesses in the cristin json schema?

PR #1274, opened 2023-08-04
Np 21184 secure fetch publication endpoint

truhacevkir:
Task description from Jira:

```
In order to prevent unauthorized access to publication data (status Published_metadata (unauthorized user sees AssociatedArtifacts) and Draft (unauthorized user sees all data), the FetchPublication handler should require authorization to differentiate between logged in and anonymous users. 

Draft:
Owner/Curator should see everything
Other users should get 404

Published_metadata: 
Anonymous/Non-owner users should be able to view metadata about the publication, but not associated artifacts.
Owner/Curator should see everything

Published
Everyone should see everything
```

PS: As i can see we do not longer use PUBLISHED_METADATA status, but instead updating status to PUBLISHED but keep associated files UNPUBLISHED (CreateTicketHandler contains that logic)

I think we want to return unpublished files in response, so frontend can show how many files that are waiting for approval, or not?

Or do we want to keep only type of file, .i.e UnpublishedFile so frontend can use this field to show number of unpublished files and do not expose all other UnpublishedFile fields?

### DIFF
--- a/publication-rest/src/main/java/no/unit/nva/publication/fetch/FetchPublicationHandler.java
+++ b/publication-rest/src/main/java/no/unit/nva/publication/fetch/FetchPublicationHandler.java
@@ -33,10 +33,12 @@ import no.unit.nva.publication.external.services.RawContentRetriever;
 import no.unit.nva.publication.service.impl.ResourceService;
 import no.unit.nva.schemaorg.SchemaOrgDocument;
 import no.unit.nva.transformer.Transformer;
+import nva.commons.apigateway.AccessRight;
 import nva.commons.apigateway.ApiGatewayHandler;
 import nva.commons.apigateway.RequestInfo;
 import nva.commons.apigateway.exceptions.ApiGatewayException;
 import nva.commons.apigateway.exceptions.GoneException;
+import nva.commons.apigateway.exceptions.NotFoundException;
 import nva.commons.apigateway.exceptions.UnsupportedAcceptHeaderException;
 import nva.commons.core.Environment;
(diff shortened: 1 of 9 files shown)


publication-rest/src/main/java/no/unit/nva/publication/fetch/FetchPublicationHandler.java
```diff
@@ -83,53 +82,55 @@ public FetchPublicationHandler(ResourceService resourceService,
...
     @JacocoGenerated
     private static ResourceService defaultResourceService(AmazonDynamoDB client) {
         return new ResourceService(client, CLOCK);
     }
 
+    private String throwNotFoundException() throws NotFoundException {
+        throw new NotFoundException("Publication is not found");
+    }
+
+    private boolean userIsCuratorOrOwner(RequestInfo requestInfo, Publication publication) {
```
monaullah, line 118: You could extract for readability, but not necessary:
```
private boolean userIsCuratorOrOwner(RequestInfo requestInfo, Publication publication) {
        return userIsCurator(requestInfo) || userIsOwner(requestInfo, publication);
}

private static boolean userIsOwner(RequestInfo requestInfo, Publication publication) {
    return attempt(() -> publication.getResourceOwner().getOwner().toString().equals(requestInfo.getUserName()))
               .orElse(failure -> false);
}

private static boolean userIsCurator(RequestInfo requestInfo) {
    return requestInfo.userIsAuthorized(AccessRight.APPROVE_DOI_REQUEST.toString());
}
```

monaullah (approved):

PR #1284, opened 2023-08-23
Use .jsonld in context path

brinxmat:
Looking at the JSON-LD docs, it seems that the standard usage is to include ".jsonld" for contexts (FWIW, the current context doesn't work in playground, which is why I am attempting this, though the problem may lie deeper)


truhacevkir (approved):
Random tests still fail from time to time.

monaullah (approved):

PR #1288, opened 2023-08-25
adding Cooperation property

truhacevkir:
I need this property to create filter for sampublikasjoner in search-api

### DIFF
--- a/expansion/src/main/java/no/unit/nva/expansion/model/CollaborationType.java
+++ b/expansion/src/main/java/no/unit/nva/expansion/model/CollaborationType.java
@@ -5,7 +5,7 @@ import lombok.Getter;
 @Getter
 public enum CollaborationType {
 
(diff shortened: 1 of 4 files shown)


ketilaa (approved):

expansion/src/main/java/no/unit/nva/expansion/model/CollaborationType.java
```diff
@@ -0,0 +1,15 @@
+package no.unit.nva.expansion.model;
+
+import lombok.Getter;
+
+@Getter
+public enum CollaborationType {
+
+    COLLABORATIVE("Collaborative"), NON_COLLABORATIVE("NonCollaborative");
```
anetteOlli, line 8: Tror det trengs noe default constructor for at Jackson ikke skal begynne å klage.

anetteOlli (approved):

PR #1291, opened 2023-08-28
injecting organization and owner to publication from importCandidate

truhacevkir:


publication-commons/src/main/java/no/unit/nva/publication/model/business/importcandidate/ImportCandidate.java
```diff
@@ -34,6 +35,29 @@ public ImportCandidate() {
         super();
     }
 
+    public Builder copyImportCandidate() {
```
truhacevkir, line 38: cause copy() method from inherited class returns Publication

anetteOlli (approved):

PR #1305, opened 2023-09-15
Np 45138 cristin import map musical performance

anetteOlli:


cristin-import/src/main/java/no/unit/nva/cristin/mapper/artisticproduction/CristinArtisticProduction.java
```diff
@@ -130,4 +171,129 @@ private String[] descriptionFields() {
...
+    private boolean isConcert() {
+        return Optional.ofNullable(performance).map(Performance::isConcert).orElse(false);
+    }
+
+    private String extractExtent() {
+        var shouldExtractExtent = Optional.ofNullable(artisticProductionTimeUnit)
+            .map(ArtisticProductionTimeUnit::timeUnitIsInMinutes)
+            .orElse(false);
+        return shouldExtractExtent ? duration : null;
+    }
```
truhacevkir, line 273: What about wrapping whole method in Optional:
```
private String extractExtent() {
        return Optional.ofNullable(artisticProductionTimeUnit)
                   .map(ArtisticProductionTimeUnit::timeUnitIsInMinutes)
                   .map(timeIsInMinutes -> timeIsInMinutes ? duration : null)
                   .orElse(null);
    }
```

cristin-import/src/main/java/no/unit/nva/cristin/mapper/artisticproduction/CristinArtisticProduction.java
```diff
@@ -130,4 +171,129 @@ private String[] descriptionFields() {
...
+            constructIsrc())).orElseThrow();
+    }
+
+    private Isrc constructIsrc() {
+        return Optional.of(isrc).map(this::extractIsrc).orElse(null);
+    }
+
+    private Isrc extractIsrc(String isrc) {
+        return attempt(() -> new Isrc(isrc)).orElseThrow(r -> new InvalidIsrcException(r.getException()));
+    }
```
truhacevkir, line 233: fail/failure instead of r

cristin-import/src/test/java/cucumber/ArtisticFeatures.java
```diff
@@ -110,4 +104,221 @@ public void theCristinResultLackTheDurationInBothMetadataFields() {
...
+    }
+
+    private CristinProduct readProductFromResource() {
+        var cristinProductString = readResourceFile(Path.of("type_produkt.json"));
+        return attempt(
+            () -> JsonUtils
+                .dtoObjectMapper
+                .readValue(cristinProductString, CristinProduct.class))
+            .orElseThrow();
+    }
```
truhacevkir, line 306: Why does code style wrap it like this

cristin-import/src/main/java/no/unit/nva/cristin/mapper/artisticproduction/CristinArtisticProduction.java
```diff
@@ -130,4 +171,129 @@ private String[] descriptionFields() {
...
+    private Optional<AudioVisualPublication> extractAudioVisualPublication() {
+        if (StringUtils.isNotEmpty(isrc) || hasAudioVisialMedium()) {
+            return Optional.of(createAudioVisualPublication());
+        }
+        return Optional.empty();
+    }
+
+    private boolean hasAudioVisialMedium() {
+        return nonNull(medium) && CRISTIN_MEDIUM_TO_NVA_MEDIUM.containsKey(medium);
+    }
```
truhacevkir, line 217: *Visual

cristin-import/src/main/java/no/unit/nva/cristin/mapper/artisticproduction/CristinArtisticProduction.java
```diff
@@ -9,15 +9,41 @@
...
 
-/**
+import static java.util.Objects.nonNull;
+import static no.unit.nva.model.instancetypes.artistic.music.MusicMediaType.COMPACT_DISC;
+import static no.unit.nva.model.instancetypes.artistic.music.MusicMediaType.DIGITAL_FILE;
+import static no.unit.nva.model.instancetypes.artistic.music.MusicMediaType.OTHER;
+import static no.unit.nva.model.instancetypes.artistic.music.MusicMediaType.VINYL;
+import static nva.commons.core.attempt.Try.attempt;
+
+/*
```
truhacevkir, line 46: Is it a javadoc or comment?

truhacevkir (approved):

cristin-import/src/main/java/no/unit/nva/cristin/mapper/artisticproduction/CristinArtisticProduction.java
anetteOlli: it just a comment for me (or others) to keep track of what secondary category has this object attached to it.

cristin-import/src/test/java/cucumber/ArtisticFeatures.java
anetteOlli: Looking better now?

To answer you question:
Because code-style is broken on my computer. Reloading the UnitStyle does not work.

truhacevkir (approved):

cristin-import/src/test/java/cucumber/ArtisticFeatures.java
```diff
@@ -110,4 +104,232 @@ public void theCristinResultLackTheDurationInBothMetadataFields() {
...
+            .stream()
+            .filter(manifestation -> manifestation instanceof AudioVisualPublication)
+            .findFirst();
+        assertThat(audioVisualManifestationOptional.isPresent(), is(equalTo(true)));
+        var audioVisualManifestation = (AudioVisualPublication) audioVisualManifestationOptional.get();
+        assertThat(audioVisualManifestation.getMediaType().getType().getValue(), is(equalTo(expectedMediaSubType)));
+        assertThat(audioVisualManifestation.getIsrc(), is(equalTo(new Isrc(expectedIsrc))));
+        var actualPublisher = (UnconfirmedPublisher) audioVisualManifestation.getPublisher();
+        assertThat(actualPublisher.getName(), is(equalTo(expectedUnconfirmedPublisher)));
+    }
```
brinxmat, line 252: "Best practice" would be to split the various things into statements "And…has an X", but it isn't necessary to change this as I don't think these steps will be re-used

brinxmat: In fact, the first step exists at the bottom of the file.

cristin-import/src/main/java/no/unit/nva/cristin/mapper/artisticproduction/CristinArtisticProduction.java
```diff
@@ -130,4 +164,187 @@ private String[] descriptionFields() {
...
+        var matcher = pattern.matcher(medium);
+        return matcher.matches();
+    }
+
+
+    @JsonIgnore
+    private boolean isCompactDisc(String medium) {
+        var pattern = Pattern.compile(
+            ".*((CD)|(Album))+.*",
+            Pattern.CASE_INSENSITIVE);
```
brinxmat, line 287: I guess this is where you are referring to regex…I'm not sure that there is a better way…GIGO, which makes life miserable.

You can try (though it's case sensitive, so you might want to uppercase/lowercase):

```
        var input = "hello CD, this is CD";
        var things = Stream.of("CD ", "Album");
        var result = things.anyMatch(input::contains);
```

Bear in mind though that the regex could be better if using `\b`

brinxmat (changes requested):

cristin-import/src/main/java/no/unit/nva/cristin/mapper/artisticproduction/CristinArtisticProduction.java
anetteOlli: sadly the mapping rules needs to be case insensitive.

cristin-import/src/test/java/cucumber/ArtisticFeatures.java
anetteOlli: Fixed here and for other places too.

PR #1311, opened 2023-09-22
Np 45262 cristin import artistic details mapped to description

anetteOlli:


cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinMapper.java
```diff
@@ -62,6 +63,7 @@ public class CristinMapper extends CristinMappingModule {
     public static final String UNIT_INSTITUTION_CODE = "UNIT";
     public static final ResourceOwner SIKT_OWNER = new CristinLocale("SIKT", "20754", "0", "0",
                                                                      "0").toResourceOwner();
+    private static final String NEW_LINE = "\n";
```
torarnet, line 66: There also is System.lineSeparator() or something which uses the correct separator for each system (linux, windows etc) but im not sure if its needed here.

cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinMapper.java
```diff
@@ -270,16 +272,24 @@ private Instant extractEntryLastModifiedDate() {
...
+            .withNpiSubjectHeading(extractNpiSubjectHeading())
+            .withAbstract(extractAbstract())
+            .withTags(extractTags())
+            .withAlternativeAbstracts(Collections.emptyMap())
+            .withDescription(extractDescription())
+            .build();
+    }
+
+    private String extractDescription() {
+        return Optional.ofNullable(cristinObject.getCristinArtisticProduction())
```
torarnet, line 289: is CristinArtisticProduction the only one that uses description?

torarnet (commented):
See comments. Will approve afterwards

cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinMapper.java
anetteOlli: So far yes, (and most likely to be so forever).

It all the fields from cristin that doesn't quite fit the NVA-datamodel. Not a design award this implementation, but it's either that or  dataloss.

torbjokv (approved):

torarnet (approved):

PR #1315, opened 2023-09-25
NP-45320-cristin-import-latest-s3-uri-in-error-reports

anetteOlli:


anetteOlli:
There's pros and cons for this PR.

We need an easy method for re-importing single cristin-result that fails because of  some IOexception.

The cons of this is that we now need to be very-careful and move/delete everything in the error path before running a new import-job, otherwise we will just overwrite the previous error reports.

cristin-import/src/test/java/no/unit/nva/cristin/lambda/CristinEntryEventConsumerTest.java
```diff
@@ -526,9 +518,21 @@ private static JavaType constructImportResultJavaType() {
...
     }
 
-    private SQSEvent createEventReference(FileContentsEvent eventBody) throws IOException {
+    private EventReference createEventReference(FileContentsEvent eventBody) throws IOException {
         var eventFileUri = s3Driver.insertEvent(UnixPath.EMPTY_PATH, eventBody.toJsonString());
-        var eventReference = new EventReference(randomString(), EVENT_SUBTOPIC, eventFileUri);
+        return new EventReference(randomString(), EVENT_SUBTOPIC, eventFileUri);
+    }
+
+    private SQSEvent createSQSevent(FileContentsEvent eventBody) throws IOException {
```
brinxmat, line 526: `createSqsEvent` to avoid checkstyle complaints

brinxmat (changes requested):

torarnet (approved):

brinxmat (approved):

PR #1320, opened 2023-09-27
NP-45324-ignore-unresolvable-dois

anetteOlli:


torarnet (approved):

anetteOlli:
There are 5000 unresolvable DOIs. The curators can't fix them all manually. 
And we can't be sure 100 % of them are in fact broken not just "at this time responding with an error". So I think we should keep them.

brinxmat (approved):

PR #1329, opened 2023-10-06
set publishedDate to now and remove contributors restriction

truhacevkir:


brage-import/src/main/java/no/sikt/nva/brage/migration/mapper/BrageNvaMapper.java
```diff
@@ -68,8 +69,8 @@ public static Publication toNvaPublication(Record brageRecord)
         var publication = new Publication.Builder()
                               .withHandle(extractHandle(brageRecord))
                               .withEntityDescription(extractEntityDescription(brageRecord))
-                              .withCreatedDate(extractPublishedDate(brageRecord))
-                              .withPublishedDate(extractPublishedDate(brageRecord))
+                              .withCreatedDate(ZonedDateTime.now().toInstant())
```
anetteOlli, line 72: Uff da. Somebody got a correction list of bad formatted dates in Brage they needed to correct. 

They apparently didn't need to fix that.

anetteOlli (approved):

PR #1330, opened 2023-10-09
added customer name to all reports paths

truhacevkir:


brage-import/src/main/java/no/sikt/nva/brage/migration/lambda/BrageEntryEventConsumer.java
```diff
@@ -267,11 +264,16 @@ private String determineBestEventReference(S3Event event) {
...
+        return UriWrapper.fromUri(ERROR_BUCKET_PATH)
+                   .addChild(extractInstitutionNameFromRecord())
+                   .addChild(timePath(event))
+                   .addChild(exception.getClass().getSimpleName())
+                   .addChild(UriWrapper.fromUri(extractObjectKey(event)).getLastPathElement());
+    }
+
+    private String extractInstitutionNameFromRecord() {
+        var record = attempt(() -> parseBrageRecordJson(brageRecordFile)).orElseThrow();
+        return record.getCustomer().getName();
```
anetteOlli, line 276: Det er ikke fra rekorden, men fra s3eventen som inneholder institusjonsnavn.

anetteOlli (changes requested):

brage-import/src/main/java/no/sikt/nva/brage/migration/lambda/BrageEntryEventConsumer.java
truhacevkir: I mean we do following:
S3Event -> reading s3 resource uri from event -> fetching s3 resource -> convert s3 resource to record -> reading customer value from record.

anetteOlli (approved):

PR #1337, opened 2023-10-17
Log instead of error when no import candidate is found with given sco…

joachimjorgensen:
Asserts that an import candidate is found for a given scopus identifier before continuing. Logs info message if no import candidate is found.

Emulating "Import Candidate Delete Event" is changed to not be wrapped by EventReference so that the handler can correctly handle the event.

### DIFF
--- a/publication-event-handlers/src/main/java/no/unit/nva/publication/events/handlers/delete/DeleteImportCandidateEventConsumer.java
+++ b/publication-event-handlers/src/main/java/no/unit/nva/publication/events/handlers/delete/DeleteImportCandidateEventConsumer.java
@@ -16,7 +16,10 @@ import no.unit.nva.publication.service.impl.ResourceService;
 import nva.commons.apigateway.exceptions.BadGatewayException;
 import nva.commons.core.Environment;
 import nva.commons.core.JacocoGenerated;
+import nva.commons.core.attempt.Try;
 import nva.commons.core.paths.UriWrapper;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class DeleteImportCandidateEventConsumer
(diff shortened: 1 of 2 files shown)


publication-event-handlers/src/main/java/no/unit/nva/publication/events/handlers/delete/DeleteImportCandidateEventConsumer.java
```diff
@@ -31,6 +35,10 @@ public class DeleteImportCandidateEventConsumer
     public static final int UNIQUE_HIT_FROM_SEARCH_API = 0;
     public static final String COULD_NOT_FETCH_UNIQUE_IMPORT_CANDIDATE_MESSAGE = "Could not fetch unique import "
                                                                                  + "candidate";
+    public static final String NO_IMPORT_CANDIDATE_FOUND = "No import candidate found with scopus identifier %s";
+    private static final Logger logger = LoggerFactory.getLogger(DeleteImportCandidateEventConsumer.class);
+    public static final int ZERO_HITS = 0;
```
hackon, line 40: private

hackon: This goes for all constant fields in a class that are not used elsewhere. FYI When creating them in Intellij they default public.

hackon (approved):

publication-event-handlers/src/main/java/no/unit/nva/publication/events/handlers/delete/DeleteImportCandidateEventConsumer.java
joachimjorgensen: Fixed 👍

PR #1343, opened 2023-10-18
Set project name to an empty string when generating publications with projects in Cristin import

joachimjorgensen:
When projects are generated from Presentational Work when generating a Publication in Cristin import, the project name will now be an empty string.


cristin-import/src/main/java/no/unit/nva/cristin/lambda/constants/HardcodedValues.java
```diff
@@ -6,7 +6,7 @@ public final class HardcodedValues {
     public static final String UNIT_CUSTOMER_ID = "f54c8aa9-073a-46a1-8f7c-dde66c853934";
     public static final String HARDCODED_PUBLICATIONS_OWNER = "someone@unit.no";
     //RESEARCH_PROJECT
-    public static final String HARDCODED_RESEARCH_PROJECT_NAME = "Some project name";
```
anetteOlli, line 9: Change it to null-value so it is transparent that we haven't mapped it because there are no project name strings present. Not that the Cristin data contained project names with empty strings.

anetteOlli (changes requested):

anetteOlli (approved):

PR #1345, opened 2023-10-20
adding test to verify functionality

truhacevkir:
We already create HANDLE_REPORT for brage record which is merged with existing nva publication:
Merged record is reported in both UPDATE_REPORT (publication json before merging as file content) and handle report (handle of incoming brage record as file content)

It is not the problem that some publications will have multiple reports, because brage catalogs do not have duplicates (we remove duplicates when creating brage records), and HANDLE_REPORTS have institution name in s3 path.

Multiple brage records (from different institution) merged with nva publication will have following handle reports:

ntnu/timestamp/publicationId
nmbu/timestamp/publicationId
uio/timestamp/publicationId

### DIFF
--- a/brage-import/src/main/java/no/sikt/nva/brage/migration/lambda/BrageEntryEventConsumer.java
+++ b/brage-import/src/main/java/no/sikt/nva/brage/migration/lambda/BrageEntryEventConsumer.java
@@ -43,7 +43,7 @@ public class BrageEntryEventConsumer implements RequestHandler<S3Event, Publicat
     public static final Random RANDOM = new Random(System.currentTimeMillis());
     public static final int MAX_EFFORTS = 10;
     public static final String SOURCE_CRISTIN = "Cristin";
-    public static final String BRAGE_MIGRATION_ERROR_BUCKET_NAME = "BRAGE_MIGRATION_ERROR_BUCKET_NAME";
+    public static final String BRAGE_MIGRATION_REPORTS_BUCKET_NAME = "BRAGE_MIGRATION_ERROR_BUCKET_NAME";
     public static final String YYYY_MM_DD_HH_FORMAT = "yyyy-MM-dd:HH";
     public static final String ERROR_BUCKET_PATH = "ERROR";
     public static final String HANDLE_REPORTS_PATH = "HANDLE_REPORTS";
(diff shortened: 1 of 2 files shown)


brage-import/src/main/java/no/sikt/nva/brage/migration/lambda/BrageEntryEventConsumer.java
```diff
@@ -43,7 +43,7 @@ public class BrageEntryEventConsumer implements RequestHandler<S3Event, Publicat
     public static final Random RANDOM = new Random(System.currentTimeMillis());
     public static final int MAX_EFFORTS = 10;
     public static final String SOURCE_CRISTIN = "Cristin";
-    public static final String BRAGE_MIGRATION_ERROR_BUCKET_NAME = "BRAGE_MIGRATION_ERROR_BUCKET_NAME";
+    public static final String BRAGE_MIGRATION_REPORTS_BUCKET_NAME = "BRAGE_MIGRATION_ERROR_BUCKET_NAME";
```
truhacevkir, line 46: Renamin, cause it is actually a report bucket

brage-import/src/test/java/no/sikt/nva/brage/migration/lambda/BrageEntryEventConsumerTest.java
```diff
@@ -1108,9 +1128,9 @@ private String extractActualHandleReportFromS3Client(S3Event s3Event, Publicatio
     private UriWrapper constructHandleReportFileUri(S3Event s3Event, Publication actualPublication) {
         var timestamp = s3Event.getRecords().get(0).getEventTime().toString(YYYY_MM_DD_HH_FORMAT);
         return UriWrapper.fromUri(HANDLE_REPORTS_PATH)
-                   .addChild(actualPublication.getResourceOwner().getOwner().getValue().split("@")[0])
-                   .addChild(timestamp)
-                   .addChild(actualPublication.getIdentifier().toString());
+                                    .addChild("institution")
```
anetteOlli, line 1131: Do we know that there is no two Brage posts in the same archive that points to the same NVA publication?

If yes, then this is fine, if no. We need to change the path to somehow reflect the Brage-handle and not the NVA publication identifier.

anetteOlli (approved):

PR #1347, opened 2023-10-24
NP-45511-cristin-import-bug-in-exhibition-mapping

anetteOlli:


torarnet:
Why is the build sometimes failing?

anetteOlli:
It randomly fails most likely to some random data generation. It is almost impossible to find the exceptions in the long list of logs at GitHub. So I only create jira task for random build fails when it happens locally on my computer.

PR #1351, opened 2023-10-27
NP-45418 Check for existing dois in nva-publications for Cristin-import

joachimjorgensen:
When importing from Cristin, uses the search-api to check if there are existing nva-publications with the same doi, and throws an exception if so.

### DIFF
--- a/cristin-import/src/main/java/no/unit/nva/cristin/lambda/DoiDuplicateChecker.java
+++ b/cristin-import/src/main/java/no/unit/nva/cristin/lambda/DoiDuplicateChecker.java
@@ -36,14 +36,6 @@ public class DoiDuplicateChecker {
                    .orElse(pubRep);
     }
 
-    private PublicationRepresentations checkForHits(String response, URI doi, PublicationRepresentations pubRep) {
-        var searchRespnse = toResponse(response);
-        if (searchRespnse.totalHits() > 0) {
-            throw new DuplicateDoiException(doi);
-        }
-        return pubRep;
-    }
-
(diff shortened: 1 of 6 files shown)


cristin-import/src/main/java/no/unit/nva/cristin/lambda/DoiDuplicateChecker.java
```diff
@@ -0,0 +1,63 @@
...
+
+    private PublicationRepresentations fetchNvaPublicationsByDoi(URI doi, PublicationRepresentations pubRep) {
+        var request = constructSearchUri(doi);
+        return getResponseBody(request)
+                   .map(response -> checkForHits(response, doi, pubRep))
+                   .orElse(pubRep);
+    }
+
+    private PublicationRepresentations checkForHits(String response, URI doi, PublicationRepresentations pubRep) {
+        var searchRespnse = toResponse(response);
```
truhacevkir, line 40: typo
*searchResponse

cristin-import/src/main/java/no/unit/nva/cristin/lambda/DoiDuplicateChecker.java
```diff
@@ -0,0 +1,63 @@
...
+        if (searchRespnse.totalHits() > 0) {
```
truhacevkir, line 41: I would extract to method like hasHits() or containsHits(), but it is up to you :)

truhacevkir (approved):

anetteOlli (approved):

PR #1390, opened 2023-11-15
filtering out inactive affiliations - scopus import

truhacevkir:


ketilaa (approved):

scopus-import/src/test/java/no/sikt/nva/scopus/conversion/ContributorExtractorTest.java
```diff
@@ -108,6 +110,32 @@ void shouldCreateContributorByFetchingCristinPersonByOrcIdWhenNoResponseFromPia(
         assertThat(contributor.getIdentity().getName(), is(equalTo(expectedContributorName)));
     }
 
+    @Test
+    void shouldMapActiveAffiliationsFromCristinPersonOnlyWhenCreatingOrganizationsForContributor() {
+        var document = ScopusGenerator.createWithNumberOfContributorsFromAuthorTp(1).getDocument();
+        var cristinPerson = mockCristinPersonWithSingleActiveAffiliationResponse();
```
brinxmat, line 116: This should be a contributor with an active and inactive affiliation to actually test what you want. Either the naming is wrong, or the method needs creating.

brinxmat (changes requested):

scopus-import/src/test/java/no/sikt/nva/scopus/conversion/ContributorExtractorTest.java
truhacevkir: Test name is not the best yes, but it is actually testing that we extract only active affiliations, cause I am mocking one active and one inactive affiliation.

```
    private static Set<Affiliation> generateAffiliationsWithSingleActiveAffiliation() {
        return Set.of(generateInactiveAffiliation(), generateActiveAffiliation());
    }
```

PR #1396, opened 2023-11-20
createdDate for importCandidate

truhacevkir:


scopus-import/src/main/java/no/sikt/nva/scopus/ScopusConverter.java
```diff
@@ -116,7 +116,7 @@ public ImportCandidate generateImportCandidate() {
                    .withResourceOwner(constructResourceOwner())
                    .withAdditionalIdentifiers(generateAdditionalIdentifiers())
                    .withEntityDescription(generateEntityDescription())
-                   .withModifiedDate(Instant.now())
+                   .withCreatedDate(Instant.now())
```
brinxmat, line 119: Generally, the modified date is set to the created date when initialised, but this maybe isn't relevant

brinxmat (approved):

PR #1418, opened 2023-11-28
attempting to fix ticket indexing

truhacevkir:
This fixes ticket indexing issue and autocompleted ticket will never be "not seen"


publication-commons/src/main/java/no/unit/nva/publication/model/business/PublishingRequestCase.java
```diff
@@ -274,6 +276,11 @@ && getStatus() == that.getStatus()
                && Objects.equals(getOwnerAffiliation(), that.getOwnerAffiliation());
     }
 
+    public PublishingRequestCase persistAutoComplete(TicketService ticketService) throws ApiGatewayException {
+        this.setStatus(TicketStatus.COMPLETED);
+        return ticketService.createTicket(this);
+    }
```
truhacevkir, line 282: Better naming?
Other method names are -> persistNew(), complete(), updateAssignee(), so I think method ticket.persistNewTicketWithStatusPublished() does not fit so well ? Any thoughts?

brinxmat (approved):

PR #1427, opened 2023-11-30
topLevelOrg isNvaCustomer should be fixed

truhacevkir:


scopus-import/src/main/java/no/sikt/nva/scopus/conversion/NvaCustomerConnection.java
```diff
@@ -36,7 +36,7 @@ public boolean isNvaCustomer(CristinOrganization cristinOrganization) {
                    .map(this::fetchResponse)
                    .map(HttpResponse::statusCode)
                    .map(NvaCustomerConnection::isHttpOk)
-                   .orElseThrow();
```
anetteOlli, line 39: Not sure if I agree on ignoring the exception and return false. This API is our own api, it should be up and running at all times.

anetteOlli (approved):

scopus-import/src/main/java/no/sikt/nva/scopus/conversion/NvaCustomerConnection.java
truhacevkir: Endrer tilbake til exception

PR #1439, opened 2023-12-11
fix: directly trigger DOI Registrar Client update from publication api #NP-45622

torbjokv:


publication-rest/src/main/java/no/unit/nva/publication/delete/DeletePublicationHandler.java
```diff
@@ -127,6 +141,12 @@ private void handleSoftDeletion(RequestInfo requestInfo, Publication publication
         var duplicate = requestInfo.getQueryParameterOpt(DUPLICATE_QUERY_PARAM).orElse(null);
         resourceService.unpublishPublication(toPublicationWithDuplicate(duplicate, publication));
         persistNotification(publication);
+        eventBridgeClient.putEvents(PutEventsRequest.builder().entries(PutEventsRequestEntry.builder()
```
monaullah, line 144: extract to own method?

publication-rest/src/main/java/no/unit/nva/publication/delete/DeletePublicationHandler.java
```diff
@@ -127,6 +141,12 @@ private void handleSoftDeletion(RequestInfo requestInfo, Publication publication
...
+               .source(NVA_PUBLICATION_DELETE_SOURCE)
+               .detailType(LAMBDA_DESTINATIONS_INVOCATION_RESULT_SUCCESS)
+               .detail(DoiMetadataUpdateEvent.createUpdateDoiEvent(publication).toJsonString())
+               .resources(publication.getIdentifier().toString())
+               .build()).build());
```
truhacevkir, line 149: Should be persisted if needed, etc. when publication has nva provided doi. Or not?

truhacevkir: I would extract this piece of code to new method:
`persistDoiUpdateEvent(publication);`

publication-rest/src/test/java/no/unit/nva/publication/delete/DeletePublicationHandlerTest.java
```diff
@@ -98,7 +107,11 @@ public void setUp() throws NotFoundException {
         prepareIdentityServiceClient();
         publicationService = new ResourceService(client, Clock.systemDefaultZone());
         ticketService = new TicketService(client);
-        handler = new DeletePublicationHandler(publicationService, ticketService, environment, identityServiceClient);
+        eventBridgeClient = mock(EventBridgeClient.class);
```
truhacevkir, line 110: We have test implementation of EventBridgeClient.

`eventBridgeClient = new FakeEventBridgeClient(ApplicationConstants.EVENT_BUS_NAME);
`
You can than use something like this to verify that event has been persisted:

```
return eventBridgeClient.getRequestEntries()
                   .stream()
                   .map(PutEventsRequestEntry::detail)
                   .map(event -> 
attempt(() -> JsonUtils.dtoObjectMapper.readValue(event, YourClass.class)).orElseThrow())
```

publication-rest/src/test/java/no/unit/nva/publication/delete/DeletePublicationHandlerTest.java
```diff
@@ -286,6 +301,8 @@ void shouldSuccessWhenUserIsContributorWhenUnpublishingPublication()
 
         var response = GatewayResponse.fromOutputStream(outputStream, Void.class);
         assertThat(response.getStatusCode(), is(equalTo(SC_ACCEPTED)));
+
+        verify(eventBridgeClient, Mockito.times(1)).putEvents(argThat(new EventBridgeRequestMatcher()));
```
monaullah, line 305: I think I would assert this explicitly in another test: `shouldProduceUpdateDoiEventWhenDeletionIsSuccessful()`, and assert that the PutEventsRequestEntry contains the expected detail (DoiMetadataUpdateEvent)

monaullah (changes requested):

PR #1441, opened 2023-12-12
fix: wrap detail payload in LambdaDestinationInvocationResult in DeletePublicationHandler #NP-45622

torbjokv:
…

### DIFF
--- a/publication-rest/src/main/java/no/unit/nva/publication/delete/LambdaDestinationInvocationResult.java
+++ b/publication-rest/src/main/java/no/unit/nva/publication/delete/LambdaDestinationInvocationResult.java
@@ -1,0 +1,11 @@
+package no.unit.nva.publication.delete;
+
+import static nva.commons.core.attempt.Try.attempt;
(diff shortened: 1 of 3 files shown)


publication-rest/src/main/java/no/unit/nva/publication/delete/LambdaDestinationInvocationResult.java
```diff
@@ -0,0 +1,11 @@
+package no.unit.nva.publication.delete;
+
+import static nva.commons.core.attempt.Try.attempt;
+import no.unit.nva.commons.json.JsonUtils;
+import no.unit.nva.publication.events.bodies.DoiMetadataUpdateEvent;
+
+public record LambdaDestinationInvocationResult(DoiMetadataUpdateEvent responsePayload) {
```
monaullah, line 7: `LambdaDestinationInvocationDetail`?

monaullah (approved):

PR #1468, opened 2024-01-08
Np 45880 do not accept files if not allowed

ketilaa:
Step one - validation on create.

We will probably have more stuff to validate, so next refactoring step would be to have plugable validator where each validator checks only one thing.


publication-rest/src/main/java/no/unit/nva/publication/validation/DefaultPublicationValidator.java
```diff
@@ -0,0 +1,57 @@
...
+            return null;
+        }
+    }
+
+    private boolean hasAtLeastOneFile(Publication publication) {
+        return publication.getAssociatedArtifacts() != null
+               &&
+               publication.getAssociatedArtifacts().stream()
+                   .anyMatch(
+                       associatedArtifact -> associatedArtifact instanceof File);
```
truhacevkir, line 55: You can use:
`.anyMatch(File.class::isInstance)`

publication-rest/src/main/java/no/unit/nva/publication/validation/config/CustomerApiFilesAllowedForTypesConfigSupplier.java
```diff
@@ -0,0 +1,72 @@
...
+        if (allowFileUploadForTypesNode != null && allowFileUploadForTypesNode.isArray()) {
+            allowFileUploadForTypesNode.elements().forEachRemaining(node -> types.add(node.asText()));
+            return types;
+        } else {
+            throw new ConfigNotAvailableException(String.format("Response from %s did not contain expected "
+                                                                + "field %s of type array!",
+                                                                request.uri().toString(),
+                                                                ALLOW_FILE_UPLOAD_FOR_TYPES_FIELD_NAME));
+        }
+    }
```
truhacevkir, line 71: I would refactor this function:

```
private static Set<String> extractFilesAllowedForTypesFromJsonResponse(HttpResponse<String> response,
                                                                           HttpRequest request)
        throws JsonProcessingException {

        var node = parseJson(response.body())
                       .get(ALLOW_FILE_UPLOAD_FOR_TYPES_FIELD_NAME);
        if (nodeIsArray(node)) {
            return extractTypes(node);
        } else {
            throw new ConfigNotAvailableException(String.format("Response from %s did not contain expected "
                                                                + "field %s of type array!",
                                                                request.uri().toString(),
                                                                ALLOW_FILE_UPLOAD_FOR_TYPES_FIELD_NAME));
        }
    }
```

Where extractTypes is:

```
    private static Set<String> extractTypes(JsonNode node) {
        return StreamSupport.stream(node.spliterator(), false)
                   .map(JsonNode::asText)
                   .collect(Collectors.toSet());
    }
```

publication-rest/src/main/java/no/unit/nva/publication/create/CreatePublicationHandler.java
```diff
@@ -136,4 +158,34 @@ private boolean isThesis(CreatePublicationRequest request) {
...
+        final var userInstance = UserInstance.createExternalUser(resourceOwner, client.getCustomerUri());
+
+        return new CustomerAwareUserContext(userInstance, customerUri);
+    }
+
+    private CustomerAwareUserContext customerAwareUserContextFromInternalUser(RequestInfo requestInfo)
+        throws UnauthorizedException {
+        var resourceOwner = createInternalResourceOwner(requestInfo);
+        var customerUri = requestInfo.getCurrentCustomer();
+        return new CustomerAwareUserContext(UserInstance.create(resourceOwner, customerUri), customerUri);
```
truhacevkir, line 185: Instead of sending customerUri two times, maybe consider something like this:
```

private CustomerAwareUserContext customerAwareUserContextFromInternalUser(RequestInfo requestInfo)
        throws UnauthorizedException {
        var resourceOwner = createInternalResourceOwner(requestInfo);
        var userInstance = UserInstance.create(resourceOwner, requestInfo.getCurrentCustomer());
        return CustomerAwareUserContext.fromUserInstance(userInstance);
    }
        
        public static CustomerAwareUserContext fromUserInstance(UserInstance userInstance) {
            return new CustomerAwareUserContext(userInstance, userInstance.getOrganizationUri());
        }
```

truhacevkir (approved):
Some comments :)

truhacevkir (approved):

PR #1480, opened 2024-01-12
fix: avoid cache on permanent redirect and rewrite unit tests for un-…

torbjokv:
…publish #NP-45946

### DIFF
--- a/publication-rest/src/main/java/no/unit/nva/publication/fetch/FetchPublicationHandler.java
+++ b/publication-rest/src/main/java/no/unit/nva/publication/fetch/FetchPublicationHandler.java
@@ -1,5 +1,6 @@
 package no.unit.nva.publication.fetch;
 
+import static com.google.common.net.HttpHeaders.CACHE_CONTROL;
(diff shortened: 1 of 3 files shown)


publication-rest/src/main/java/no/unit/nva/publication/fetch/FetchPublicationHandler.java
```diff
@@ -116,7 +117,8 @@ private String produceRemovedPublicationResponse(Publication publication) throws
 
     private String produceRedirect(URI duplicateOf) {
         statusCode = HTTP_MOVED_PERM;
-        addAdditionalHeaders(() -> Map.of(LOCATION, duplicateOf.toString()));
+        // cache control header here to avoid permanent browser caching of the redirect
+        addAdditionalHeaders(() -> Map.of(LOCATION, duplicateOf.toString(), CACHE_CONTROL, "no-cache"));
```
brinxmat, line 121: Let's hope :D

brinxmat (approved):

PR #1487, opened 2024-01-18
NP-46008 Add publications with status UNPUBLISHED and DELETED to the search index

joachimjorgensen:
Remove the event that removes unpublished and deleted publications from the s3 bucket. Instead add publications with theses statuses to the index.

Will merge this after changes in search-api is merged, so that unpublished and deleted publications won't show up in open search api.

Not really confident in the event-space in aws and template.yaml etc., so a thorough review is appreciated :)


anetteOlli:
Will the event still be picket up by DOI-service api?

StigNorland (approved):

brinxmat (approved):

PR #1488, opened 2024-01-18
now makes authenticated request to company api to get file config

ketilaa:
Temporarily disable generation of not supported publication instance types in tests to make sure they are stable. Will fix it in nva-data-model eventually.

### DIFF
--- a/publication-rest/src/main/java/no/unit/nva/publication/create/CreatePublicationHandler.java
+++ b/publication-rest/src/main/java/no/unit/nva/publication/create/CreatePublicationHandler.java
@@ -6,6 +6,7 @@ import com.amazonaws.services.dynamodbv2.AmazonDynamoDBClientBuilder;
 import com.amazonaws.services.lambda.runtime.Context;
 import com.google.common.net.HttpHeaders;
 import java.net.URI;
+import java.net.http.HttpClient;
 import java.time.Clock;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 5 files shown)


template.yaml
```diff
@@ -1241,7 +1237,6 @@ Resources:
           PERSISTED_ENTRIES_BUCKET: !Ref ExpandedEntriesPersistenceBucketName
           TABLE_NAME: !Ref NvaImportCandidatesTable
           API_HOST: !Ref ApiDomain
-          BACKEND_CLIENT_AUTH_URL: !Ref CognitoAuthorizationUri
```
truhacevkir, line 1244: Why?

publication-rest/src/main/java/no/unit/nva/publication/validation/config/CustomerApiFilesAllowedForTypesConfigSupplier.java
```diff
@@ -14,50 +14,70 @@
...
+        return new CognitoCredentials(credentials::getId, credentials::getSecret, uri);
+    }
+
+    private URI getCognitoTokenUrl() {
+        return URI.create(backendClientAuthUrl);
+    }
+
     private static Set<String> extractFilesAllowedForTypesFromJsonResponse(HttpResponse<String> response,
-                                                                           HttpRequest request)
+                                                                           URI requestUri)
```
truhacevkir, line 80: Can we do this ? 
`response.request().uri().toString()`

publication-rest/src/main/java/no/unit/nva/publication/create/CreatePublicationHandler.java
```diff
@@ -73,12 +77,22 @@ public CreatePublicationHandler() {
...
+
+        var backendClientAuthUrl = environment.readEnv("BACKEND_CLIENT_AUTH_URL");
+        var backendClientSecretName = environment.readEnv("BACKEND_CLIENT_SECRET_NAME");
+
+        this.publicationValidator =
+            new DefaultPublicationValidator(
+                new CustomerApiFilesAllowedForTypesConfigSupplier(httpClient,
+                                                                  secretsManagerClient,
+                                                                  backendClientAuthUrl,
+                                                                  backendClientSecretName));
```
truhacevkir, line 95: Looks good, but could we use AuthorizedUriRetriever here?
It looks like you do the same authorization which `AuthorizedBackendUriRetriever` does "by default"

truhacevkir (approved):

PR #1510, opened 2024-02-01
Attempting to fix update publication bug

truhacevkir:


publication-commons/src/main/java/no/unit/nva/publication/model/business/PublishingRequestCase.java
```diff
@@ -140,7 +140,6 @@ public void validateCreationRequirements(Publication publication)
         if (PublicationStatus.DRAFT_FOR_DELETION == publication.getStatus()) {
             throw new ConflictException(MARKED_FOR_DELETION_ERROR);
         }
-        assertThatPublicationHasMinimumMandatoryFields(publication);
```
truhacevkir, line 143: This line makes it impossible to update publication which contains only administrative agreements and misses entitydescription.reference.doi. We have an issue i TEST where user can not update publication because of it. 
We have to discuss if we want to have this check at all, evt. lack of what fields makes publication unpublishable

publication-rest/src/main/java/no/unit/nva/publication/update/UpdatePublicationHandler.java
```diff
@@ -286,10 +287,16 @@ private boolean userUnauthorizedToPublishThesisAndIsNotExternalClient(RequestInf
...
+                           .map(Optional::orElseThrow)
+                           .orElseThrow(this::throwException);
         return attempt(() -> JsonUtils.dtoObjectMapper.readValue(response,
-                CustomerPublishingWorkflowResponse.class)).orElseThrow();
+                CustomerPublishingWorkflowResponse.class)).orElseThrow(this::throwException);
+    }
+
+    private BadGatewayException throwException(Failure<?> failure) {
+        logger.error(failure.getException().getMessage());
+        return new BadGatewayException(UNABLE_TO_FETCH_CUSTOMER_ERROR_MESSAGE);
```
truhacevkir, line 299: Logging actual exception message.

publication-commons/src/main/java/no/unit/nva/publication/model/business/PublishingRequestCase.java
brinxmat: If the reqs have changed, we should know this…

brinxmat (approved):

PR #1511, opened 2024-02-01
Taking in use tike mime type detector

truhacevkir:
We may also consider to use TikaInputStream to fetch the file. It may have utils for extracting filenames as well.

var inputStream =  TikaInputStream.get(file.downloadFileUrl());
var mimeTypeDetector = TikaConfig.getDefaultConfig().getDetector();
var mediaType = mimeTypeDetector.detect(inputStream, new Metadata());

### DIFF
--- a/scopus-import/src/main/java/no/sikt/nva/scopus/conversion/files/ScopusFileConverter.java
+++ b/scopus-import/src/main/java/no/sikt/nva/scopus/conversion/files/ScopusFileConverter.java
@@ -246,7 +246,7 @@ public class ScopusFileConverter {
         var filename = getFilename(fetchFileResponse);
 
         return file.copy()
-                   .withMimeType(mediaType.getType())
+                   .withMimeType(mediaType.toString())
                    .withContent(inputStream)
                    .withName(filename).build();
     }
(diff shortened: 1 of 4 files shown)


scopus-import/src/main/java/no/sikt/nva/scopus/conversion/files/ScopusFileConverter.java
```diff
@@ -236,20 +238,20 @@ private List<AssociatedArtifact> extractAssociatedArtifactsFromDoi(DocTp docTp)
...
+                   .withName(filename).build();
     }
 
-    private static String extractMimeType(ScopusFile file, HttpResponse<InputStream> fetchFileResponse) {
-        return hasSupportedMimeType(file)
-                   ? file.mimeType()
-                   : fetchFileResponse.headers().firstValue(Header.CONTENT_TYPE).orElse(null);
+    private String getMimeType(InputStream inputStream) throws IOException {
+        return TikaConfig.getDefaultConfig().getDetector()
+                   .detect(new BufferedInputStream(inputStream), new Metadata()).toString();
```
truhacevkir, line 254: Problem:

Headers often miss content-type header, creating s3 file without mimeType result in default mimeType application/octet-stream. Our endpoint downloads this file instead of showing it. Tika seems to do what we need with minimum code.

brinxmat (approved):

PR #1515, opened 2024-02-01
Showing nva customers in import candidates index

truhacevkir:
Task:

Show only organizations which are NVA customers in import-candidate index. 

Problem:

When creating import-candidate, we already make a call to fetch nva customer for every contributor. 
To avoid to do the same api calls twice, I am adding nvaContributors to import-candidate which contain data about nvaCustomer. 
It's pretty ugly solution, but I do not see other ways of doing it. At the same time it saves 2 api calls for every contributor (one to cristin and one to customer-api)


scopus-import/src/main/java/no/sikt/nva/scopus/conversion/NvaCustomerConnection.java
```diff
@@ -30,8 +31,23 @@ public NvaCustomerConnection(AuthorizedBackendUriRetriever uriRetriever) {
...
 
-    public boolean isNvaCustomer(CristinOrganization cristinOrganization) {
-        return isPresentAndHasTopLevelOrg(cristinOrganization) && isCustomer(cristinOrganization);
+    public NvaCustomer fetchCustomer(CristinOrganization cristinOrganization) {
+        return Optional.ofNullable(cristinOrganization)
+                   .filter(NvaCustomerConnection::hasTopLevelOrg)
+                   .filter(this::isCustomer)
+                   .map(org -> new NvaCustomer(true, getCristinTopLevelOrg(cristinOrganization)))
+                   .orElse(defaultNvaCustomer());
+    }
```
brinxmat, line 40: We can talk about caching something here…I don't think it's necessary to repeat the call…

brinxmat (approved):

PR #1520, opened 2024-02-02
filtering out non customer organizations

truhacevkir:


expansion/src/main/java/no/unit/nva/expansion/model/ExpandedImportCandidate.java
```diff
@@ -417,22 +435,88 @@ private static String extractPublicationYear(ImportCandidate importCandidate) {
...
+                                                                                 RawContentRetriever uriRetriever) {
+
+        return getOrganizationIdList(importCandidate)
+                   .map(id -> fetchCristinOrg(id, uriRetriever))
+                   .filter(Optional::isPresent)
+                   .map(Optional::get)
+                   .distinct()
+                   .filter(org -> isNvaCustomer(org, uriRetriever))
+                   .map(ExpandedImportCandidateOrganization::fromCristinOrganization)
+                   .collect(Collectors.toSet());
```
brinxmat, line 448: I guess this is the expensive code? I guess we know the mappings here — we should look at providing this via the listCustomers API as some kind of hash table.

brinxmat (approved):

PR #1521, opened 2024-02-02
using tika for fetching files

truhacevkir:


scopus-import/src/main/java/no/sikt/nva/scopus/conversion/files/ScopusFileConverter.java
```diff
@@ -239,21 +246,16 @@ private List<AssociatedArtifact> extractAssociatedArtifactsFromDoi(DocTp docTp)
...
+        var response = fetchResponseAsInputStream(file.downloadFileUrl());
+        var inputStream = tikaUtils.fetch(response.request().uri());
+        var filename = getFilename(response);
+        var size = inputStream.getLength();
         return file.copy()
-                   .withMimeType(getMimeType(fetchFileResponse.body()))
-                   .withContent(fetchFileResponse.body())
+                   .withContent(inputStream)
                    .withSize(size)
                    .withName(filename).build();
```
truhacevkir, line 256: Consuming inputstream two times. One for fetching response and extracting filename. Second one for fetching file using tika. `TikaInputStream + TikaConfig` seems to do much better work than `InputStream + Tikaconfig`.

scopus-import/src/main/java/no/sikt/nva/scopus/conversion/files/ScopusFileConverter.java
```diff
@@ -110,7 +112,8 @@ public List<AssociatedArtifact> fetchAssociatedArtifacts(DocTp docTp) {
                                                        .filter(File.class::isInstance)
                                                        .filter(this::isValid)
                                                        .toList();
-        return !associatedArtifactsFromXmlReferences.isEmpty() ? associatedArtifactsFromXmlReferences
+        return !associatedArtifactsFromXmlReferences.isEmpty()
```
brinxmat, line 115: Can you invert the condition to remove the secret "!"?

brinxmat (approved):

PR #1529, opened 2024-02-09
Tidy unpublish request

brinxmat:
- Don't use Optional for fields
- Validation of publication URI is probably a separate concern
  - Unfortunately, the SortableIdentifier has a weird constraint
  - No reason to use a regex since it will (by experience) fail


torbjokv (approved):
looks good to me!

monaullah (approved):
Nice!

PR #1543, opened 2024-02-16
removed handler as its not used

sondrev:


publication-rest/src/main/java/no/unit/nva/publication/update/PublishPublicationHandler.java
```diff
@@ -1,74 +0,0 @@
...
-import no.unit.nva.publication.model.PublishPublicationStatusResponse;
-import no.unit.nva.publication.model.business.UserInstance;
-import no.unit.nva.publication.service.impl.ResourceService;
-import nva.commons.apigateway.ApiGatewayHandler;
-import nva.commons.apigateway.RequestInfo;
-import nva.commons.apigateway.exceptions.ApiGatewayException;
-import nva.commons.core.Environment;
-import nva.commons.core.JacocoGenerated;
-
-public class PublishPublicationHandler extends ApiGatewayHandler<Void, PublishPublicationStatusResponse> {
```
sondrev, line 20: @brinxmat  i couldent find any usages of PublishPublicationHandler, and its not in template.yaml so I assume this is legacy code and can be deleted?

brinxmat: @truhacevkir @anetteOlli Is it this that I imagined was a publishing backdoor? Since it isn't in the template, I suspect not…

anetteOlli: I imagine it was in use before we started to use the tickets system, as the last commits to this function was in 2021, and most commits to it was in 2020.

I think it's safe to delete if it's not in any template. My "backdoor" functions are stored in another repository (nva-database-tools is stored in gitlab as I don't want them public).

PR #1545, opened 2024-02-20
Verified contributors

sondrev:


publication-commons/build.gradle
```diff
@@ -23,6 +23,7 @@ dependencies {
 
     implementation libs.guava
```
brinxmat, line 24: Do we actually use Guava?

publication-commons/src/main/java/no/unit/nva/publication/utils/JenaUtils.java
```diff
@@ -0,0 +1,86 @@
...
+import org.apache.jena.rdf.model.ModelFactory;
+import org.apache.jena.rdf.model.RDFNode;
+import org.apache.jena.riot.Lang;
+import org.apache.jena.riot.RDFDataMgr;
+import org.apache.jena.riot.RiotException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@JacocoGenerated
+public class JenaUtils {
```
brinxmat, line 25: RdfUtils

brinxmat (changes requested):

publication-commons/src/main/java/no/unit/nva/publication/permission/strategy/PermissionStrategy.java
```diff
@@ -14,17 +14,20 @@
...
 import nva.commons.apigateway.AccessRight;
 
 public abstract class PermissionStrategy {
 
     protected final Publication publication;
     protected final UserInstance userInstance;
+    protected final UriRetriever uriRetriever;
 
-    protected PermissionStrategy(Publication publication, UserInstance userInstance) {
+    protected PermissionStrategy(Publication publication, UserInstance userInstance, UriRetriever uriRetriever) {
```
torbjokv, line 27: I don't think we need to extend all classes, only Curator-class

torbjokv (approved):

PR #1546, opened 2024-02-21
increase-update-handler-memory-size

anetteOlli:


anetteOlli:
Bumping memory increase from 1800 to 4096 didn't give noticeable results (both around 2000 ms duration)

Bumping to 8192 MB reduced the duration ( 700 - 800 ms duration at 8192 MB memory).

If I calculated the price change this will actually decrease our costs:

2000 ms * 0.0000000250 dollar for 1800mb = 0,00005 dollar per invocation.

700 ms * 0.0000001333 dollar for 8192 MB = 0,0000001333 dollar per invocation.

Price list: https://aws.amazon.com/lambda/pricing/

Eiriknil (approved):

brinxmat (approved):

PR #1547, opened 2024-02-21
docs

sondrev:


docs/openapi.yaml
```diff
@@ -810,9 +810,15 @@ components:
...
+              Returns a list of allowed operations for the current user on the publication:
+                * update - Can edit the publication. Accessed though PUT  
+                * delete - Permanently delete. Accessed though DELETE  
+                * terminate - Can mark publication as deleted and remove associated files. Accessed though PUT
+                * unpublish - Can change published publication status to unpublished. Accessed though PUT
+                * ticket/publish - Allowed to approve a PublishingRequest ticket
             items:
               type: string
-              description: List of allowed operations for the current user
               enum: ["update", "delete", "terminate", "unpublish", "ticket/publish"]
```
sondrev, line 822: Could not find a way to document each value individually on openapi 3.0.
It might be possible in 3.1, but swaggerGenerator might need some updates to support 3.1
https://stackoverflow.com/questions/66465888/how-to-define-enum-mapping-in-openapi

docs/openapi.yaml
```diff
@@ -810,9 +810,15 @@ components:
         properties:
           allowedOperations:
             type: array
+            description: >
+              Returns a list of allowed operations for the current user on the publication:
```
ketilaa, line 814: If bearer token is provided. If not, the list will be empty.

ketilaa (approved):

PR #1549, opened 2024-02-21
TrustedThirdPartyStrategy fix, move and change tests, avoid that external users triggers other internal strategies …

torbjokv:
…#NP-46258

avoided that external users triggers other internal strategies 

add ContributorPermissionStrategyTest, EditorPermissionStrategyTest, TrustedThirdPartyStrategyTest, NonDegreePermissionStrategyTest 

moved tests and removed redundant tests

disabled two tests because I could not manage to mock a failing userinfo request without a bigger refactoring of our ApiGatewayHandler/RestRequestHandler/RequestInfo.

This cascading getter setup makes it hard for us to control our input and we also have no good way to inject environment or httpClient as far as I can see:
public class RequestInfo {
....
    public URI getCurrentCustomer() throws UnauthorizedException {
        return (URI)this.fetchCustomerId().or(this::fetchCustomerIdFromContext).or(this::extractCustomerIdForTests).orElseThrow(UnauthorizedException::new);
    }


publication-commons/src/main/java/no/unit/nva/publication/permission/strategy/grant/CuratorPermissionStrategy.java
```diff
@@ -24,7 +24,7 @@ public CuratorPermissionStrategy(Publication publication, UserInstance userInsta
 
     @Override
     public boolean allowsAction(PublicationOperation permission) {
-        if (!canManageStandardResources() || !userRelatesToPublication()) {
+        if (!canManageStandardResources() || !userRelatesToPublication() || userInstance.isExternalClient()) {
             return false;
```
sondrev, line 28: Is there a simpler way to do this? Seems wrong to include a negative isExternalClient here. Is there something else we can check? Can we do a DenyStrategy?

publication-commons/src/main/java/no/unit/nva/publication/permission/strategy/grant/EditorPermissionStrategy.java
```diff
@@ -14,7 +14,7 @@ public EditorPermissionStrategy(Publication publication, UserInstance userInstan
 
     @Override
     public boolean allowsAction(PublicationOperation permission) {
-        if (!hasAccessRight(MANAGE_RESOURCES_ALL)) {
+        if (!hasAccessRight(MANAGE_RESOURCES_ALL) || userInstance.isExternalClient()) {
```
sondrev, line 17: :(

publication-commons/src/main/java/no/unit/nva/publication/permission/strategy/grant/ResourceOwnerPermissionStrategy.java
```diff
@@ -15,18 +13,14 @@ public ResourceOwnerPermissionStrategy(Publication publication, UserInstance use
 
     @Override
     public boolean allowsAction(PublicationOperation permission) {
+        if (userInstance.isExternalClient() || !isOwner()) {
+            return false;
```
sondrev, line 17: :(

publication-commons/src/test/java/no/unit/nva/publication/permission/strategy/CuratorPermissionStrategyTest.java
```diff
@@ -8,17 +9,17 @@
...
 import org.junit.jupiter.params.provider.EnumSource.Mode;
 
 class CuratorPermissionStrategyTest extends PublicationPermissionStrategyTest {
 
-    //region Non-degree publications
     @ParameterizedTest(name = "Should allow Curator {0} operation on non-degree resources belonging to the "
                               + "institution based on publication owner")
     @EnumSource(value = PublicationOperation.class, mode = Mode.EXCLUDE, names = {"DELETE", "TERMINATE",
-        "TICKET_PUBLISH"})
+        "TICKET_PUBLISH"})// er TICKET_PUBLISH her riktig?
```
sondrev, line 22: as long as it has MANAGE_PUBLISHING_REQUEST curator should be able to publish

sondrev (changes requested):

publication-commons/src/main/java/no/unit/nva/publication/permission/strategy/grant/CuratorPermissionStrategy.java
torbjokv: Was thinking about it, but could not figure it out. I decided that the easiest and best solution is this for now and later make sure external clients have scopes and use the normal policies. Then we could have a simple deny for external if customer id is not matching. It you have other ideas, let me know :-)

torbjokv: I will remove this check for now!

sondrev (approved):

publication-commons/src/main/java/no/unit/nva/publication/RequestUtil.java
```diff
@@ -84,7 +84,9 @@ private static UserInstance createExternalUserInstance(RequestInfo requestInfo,
             client.getCristinUrgUri()
         );
 
-        return UserInstance.createExternalUser(resourceOwner, client.getCustomerUri());
+        var accessRights = requestInfo.getAccessRights();
```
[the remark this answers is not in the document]
torbjokv: ah did not think of this. Do we use idToken? I thought we mainly used accessToken and in this case the iss value to check that it's an external user pool? I think I will leave this for now and then we can clean up stuff if we want afterwards after checking the client credentials support for setting scopes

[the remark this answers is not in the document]
torbjokv: I will deploy to sandbox to verify that it does not break stuff

PR #1550, opened 2024-02-21
fix: update publication should include allowed operations #NP-46268

torbjokv:


sondrev (changes requested):
You also need to update opanapi.yaml to reflect the new response.

sondrev (approved):

PR #1561, opened 2024-02-28
multipart copy - brage migration

truhacevkir:
- Multipart copy to be able to copy associatedArtifacts from temporary `brage-input-files` to `persisted-storage-bucket` when running brage-import.


anetteOlli (approved):
Antar du har testet med noen større og mindre filer?

PR #1564, opened 2024-03-01
Add has file enum to expansion

truhacevkir:
This makes aggregation for files way easier. Implemented as enum to be able to extend in future to support statuses as NO_PUBLIC_FILE


StigNorland (approved):
This will make aggregation easier!

PR #1567, opened 2024-03-01
Np 46336 scopus import parallellize fetch organization

anetteOlli:


scopus-import/src/main/java/no/sikt/nva/scopus/conversion/AffiliationGenerator.java
```diff
@@ -1,121 +1,144 @@
...
+        try (var customThreadPool = new ForkJoinPool(8)) {
+            var result =
+                customThreadPool.submit(() -> authorGroupList.parallelStream().map(this::retrieveCristinId))
+                    .join()
+                    .toList();
+            customThreadPool.shutdown();
+            return result;
+        } catch (Exception e) {
+            logger.warn("Parallel fetching of corporations id from PIA failed", e);
+            return authorGroupList.stream().map(this::retrieveCristinId).toList();
```
brinxmat, line 77: I'm not sure how working with physical threads on Lambda works…there are some weird side effects of using lambda that we have to think about (see e.g. https://medium.com/p/multithreading-in-aws-lambda-part-1-performance-scaling-6e2bc3d7796f), we could try VirtualThreads (we're on Java 21…).

anetteOlli: I read some articles before I began this task. I think it as long as you call .join() to make sure the AWS Lambda was waiting for the result and shutdown() to prevent memory leak you were in safe?

anetteOlli: I will change to virtual threads to avoid some aws internals.

anetteOlli:
This PR does two things: 
One is apparent: cristin organisation is fetched for lists of author groups in parallel. 

The other is that previously the institution that is stored on the author group was called for each author. So if there was two authors at NTNU, then fetch NTNU was called twice. This is now fixed.

anetteOlli:
Codacy fails because of threading, is it okay if I make codacy ignore the error?

scopus-import/src/main/java/no/sikt/nva/scopus/conversion/model/CorporationWithContributors.java
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    public Builder copy() {
+        return new Builder().withAuthorGroupTp(scopusAuthors)
+                   .withCristinOrganizationId(cristinOrganizationId)
+                   .withCristinCorporations(cristinOrganizations);
+    }
+
+    private static boolean isNotNorway(Map<String, String> labels) {
+        return labels.values().stream().map(String::toLowerCase).noneMatch(NORWAY::equals);
```
brinxmat, line 79: equalsIgnoreCase?

scopus-import/src/main/java/no/sikt/nva/scopus/conversion/model/CorporationWithContributors.java
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    private static String toOrganizationName(List<OrganizationTp> organizationTps) {
+        return organizationTps.stream()
+                   .map(organizationTp -> extractContentString(organizationTp.getContent()))
+                   .collect(Collectors.joining(AFFILIATION_DELIMITER));
+    }
+
+    private static String getIso6391LanguageCodeForSupportedNvaLanguage(String possiblyUnsupportedLanguageIso6391code) {
+        var language = LanguageMapper.getLanguageByIso6391Code(possiblyUnsupportedLanguageIso6391code);
```
brinxmat, line 89: IIRC, it doesn't return null, it returns `und`

scopus-import/src/main/java/no/sikt/nva/scopus/conversion/model/CorporationWithContributors.java
```diff
@@ -0,0 +1,148 @@
...
+                         : extractCountryNameAsAffiliation();
+        return isNotNorway(labels.orElse(Map.of()))
+                   ? List.of(new UnconfirmedOrganization(name.orElse(null)))
+                   : List.of();
+    }
+
+    private Optional<Map<String, String>> extractCountryNameAsAffiliation() {
+        return Optional.ofNullable(scopusAuthors.getAffiliation())
+                   .map(AffiliationTp::getCountry)
+                   .map(country -> Map.of(getLanguageIso6391Code(country), country));
```
brinxmat, line 112: Country is a language code?

anetteOlli: No, but the country name might be "España", so we want the label to have "spa": "España".

anetteOlli: At least that was the intent of this code, it is from 2021 and just moved from somewhere else.

brinxmat (approved):

scopus-import/src/main/java/no/sikt/nva/scopus/conversion/model/CorporationWithContributors.java
anetteOlli: Actually the build fails here are null pointer exception when I changed.

PR #1572, opened 2024-03-05
NP-45823-brage-merging-rollback-merge

anetteOlli:


anetteOlli:
Scrapped the idea of using python scripts for these rollbacks. Updating dynamoDB objets are complicated since we compress the data and the uncompressed metadata must be stored in a certain format. Changes in resource service would make such a python script very fragile. Better to keep it inside publication-api as a lambda I can trigger by a python script, also it maybe makes it easier for other developers to use.

template.yaml
```diff
@@ -964,6 +966,33 @@ Resources:
                   - Name: suffix
                     Value: '.json'
 
+  BrageMergingRollbackFunction:
+    Type: AWS::Serverless::Function
+    Properties:
+      CodeUri: brage-import
+      Handler: no.sikt.nva.brage.migration.lambda.BrageMergingRollbackHandler::handleRequest
+      Role: !GetAtt LambdaRole.Arn
```
anetteOlli, line 974: @torbjokv du hadde en plass noe ny måte å gi tillatelser til lambdaer som var mer eksplisitt?

template.yaml
```diff
@@ -964,6 +966,33 @@ Resources:
...
+          Properties:
+            EventBusName: !GetAtt InternalBus.Name
+            Pattern:
+              detail:
+                topic: [ "BrageMerging.Rollback.Request" ]
+      EventInvokeConfig:
+        DestinationConfig:
+          OnFailure:
+            Type: SQS
+            Destination: !GetAtt BrageMergingRollbackHandlerDLQ.Arn
```
anetteOlli, line 994: DLQ vil kun benyttes for ting som tryner så saftig at det ikke lages en error report.

template.yaml
torbjokv: ja! eksempel her https://github.com/BIBSYSDEV/nva-publication-api/blob/main/template.yaml#L650
en må ta bort Role på funksjonen og gjør policy en skal referere til om til managed en

torbjokv (approved):

truhacevkir (approved):

PR #1573, opened 2024-03-05
refactor(PermissionStrategy): extract CuratingInstitutionsUtil #NP-46339

torbjokv:


publication-commons/src/main/java/no/unit/nva/publication/permission/strategy/grant/CuratorPermissionStrategy.java
```diff
@@ -52,42 +46,22 @@ private boolean userIsFromSameInstitutionAsPublication() {
             return false;
         }
 
-        var resourceOwnerTopLevel = getTopLevelOrgUri(uriRetriever,
```
torbjokv, line 55: removed getTopLevelOrgUri lookup here. had a look at resourceOwner on both a imported and a newly created resource and it looks like its top level every time

publication-commons/src/main/java/no/unit/nva/publication/model/utils/CuratingInstitutionsUtil.java
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    private static Stream<Contributor> getVerifiedContributors(Publication publication) {
+        return publication.getEntityDescription().getContributors()
+                   .stream()
+                   .filter(CuratingInstitutionsUtil::isVerifiedContributor);
+    }
+
+    private static boolean isVerifiedContributor(Contributor contributor) {
+        return contributor.getIdentity() != null && contributor.getIdentity().getId() != null;
```
brinxmat, line 37: Can you Objects::notNull both of these?

brinxmat (changes requested):

brinxmat (approved):

PR #1602, opened 2024-03-21
adding recovery queue to DataEntryUpdateHandler

truhacevkir:


template.yaml
```diff
@@ -1442,6 +1442,7 @@ Resources:
           EVENTS_BUCKET: !Ref NvaEventsBucketsName
           EVENT_BUS_NAME: !GetAtt InternalBus.Name
           HANDLER_EVENTS_FOLDER: "not used"
+          RECOVERY_QUEUE: !Ref RecoveryQueue
```
anetteOlli, line 1445: Same goes here: recovery queue has a lifespan of 14 days, the event bucket only 1 day.

truhacevkir: I don't think we need this, what do you think?
If something goes wrong when reading event -> event will end up on dlq. RecoveryQueue is meant for entries we have managed to identify, so we can rerun them. When we are not able to read event, we don't have access to resource identifier -> put on dlq. (the same flow as before now)

brinxmat (approved):

PR #1604, opened 2024-03-22
implementing recovery batch scan

truhacevkir:
Sorry for large pr.

Reading messages from recovery queue and refreshing resources.
- For now refreshing only publications, will implement for other entity types later (have to find out how to query tickets and messages by using only identifiers, or extend RecoveryEntry)

- When resource is deleted and refreshing returns NotFoundException -> skipping resource and removing message from queue. 

- Have to find a strategy for how to handle other exception types.

### DIFF
--- a/publication-commons/src/main/java/no/unit/nva/publication/service/impl/ResourceService.java
+++ b/publication-commons/src/main/java/no/unit/nva/publication/service/impl/ResourceService.java
@@ -70,8 +70,9 @@ public class ResourceService extends ServiceWithTransactions {
 
     public static final Supplier<SortableIdentifier> DEFAULT_IDENTIFIER_SUPPLIER = SortableIdentifier::next;
     public static final int AWAIT_TIME_BEFORE_FETCH_RETRY = 50;
-    public static final String INVALID_PATH_ERROR =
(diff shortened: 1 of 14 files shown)


publication-commons/src/main/java/no/unit/nva/publication/service/impl/ResourceService.java
```diff
@@ -311,16 +311,20 @@ public Stream<TicketEntry> fetchAllTicketsForResource(Resource resource) {
...
 
-    private static boolean isNotRemoved(TicketEntry ticket) {
-        return !TicketStatus.REMOVED.equals(ticket.getStatus());
+    public void refresh(SortableIdentifier identifier) {
+        try {
+            updatePublication(getPublicationByIdentifier(identifier));
+            logger.info(RESOURCE_REFRESHED_MESSAGE, identifier);
+        } catch (NotFoundException e) {
+            logger.error(RESOURCE_TO_REFRESH_NOT_FOUND_MESSAGE, identifier);
+        }
```
truhacevkir, line 327: Sorry for line breaks -> this is new method.

ketilaa (approved):

PR #1606, opened 2024-03-26
NP-46416 Add identifier to expanded organization

joachimjorgensen:
Add identifier to the expanded organization so that search-api can implement parameters for organization identifiers as well as id's. (For use in viewingScope-parameter)

### DIFF
--- a/expansion/src/main/java/no/unit/nva/expansion/ResourceExpansionServiceImpl.java
+++ b/expansion/src/main/java/no/unit/nva/expansion/ResourceExpansionServiceImpl.java
@@ -108,7 +108,11 @@ public class ResourceExpansionServiceImpl implements ResourceExpansionService {
                                              .map(UriWrapper::getLastPathElement)
                                              .orElse(null);
 
-            var partOf = RdfUtils.getAllNestedPartOfs(uriRetriever, organizationId);
+            var partOf = RdfUtils.getAllNestedPartOfs(uriRetriever, organizationId)
+                             .stream()
+                             .map(uri -> new ExpandedOrganization(uri, UriWrapper.fromUri(uri).getLastPathElement(), null))
+                             .toList();
+
             return new ExpandedOrganization(organizationId, organizationIdentifier, partOf);
         }
         return null;
--- a/expansion/src/main/java/no/unit/nva/expansion/ResourceExpansionServiceImpl.java
+++ b/expansion/src/main/java/no/unit/nva/expansion/ResourceExpansionServiceImpl.java
@@ -102,8 +102,12 @@ public class ResourceExpansionServiceImpl implements ResourceExpansionService {
                                       : resourceService.getResourceByIdentifier(ticketEntry.getResourceIdentifier())
                                           .getResourceOwner().getOwnerAffiliation();
 
+            var organizationIdentifier = Objects.nonNull(organizationId)
+                                             ? UriWrapper.fromUri(organizationId).getLastPathElement()
+                                             : "";
+
             var partOf = RdfUtils.getAllNestedPartOfs(uriRetriever, organizationId);
-            return new ExpandedOrganization(organizationId, partOf);
+            return new ExpandedOrganization(organizationId, organizationIdentifier, partOf);
         }
         return null;
     }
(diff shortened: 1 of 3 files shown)


expansion/src/main/java/no/unit/nva/expansion/ResourceExpansionServiceImpl.java
```diff
@@ -102,8 +102,12 @@ public ExpandedOrganization getOrganization(Entity dataEntry) throws NotFoundExc
                                       : resourceService.getResourceByIdentifier(ticketEntry.getResourceIdentifier())
                                           .getResourceOwner().getOwnerAffiliation();
 
+            var organizationIdentifier = Objects.nonNull(organizationId)
+                                             ? UriWrapper.fromUri(organizationId).getLastPathElement()
+                                             : "";
```
truhacevkir, line 107: What about null instead of empty string?

Can be wrapped in optional of nullable:
```
var organizationIdentifier = Optional.ofNullable(organizationId)
                                             .map(UriWrapper::fromUri)
                                             .map(UriWrapper::getLastPathElement)
                                             .orElse(null);
```

joachimjorgensen: Good catch! Will do

sondrev (approved):

joachimjorgensen:
Might discard this PR, as it could be a possibility to let search-api convert the identifier to an id in the endpoint instead. This would avoid the need of re-indexing. As per now, viewingScope is the only one that needs this identifier accessible from the search-api.

StigNorland (approved):

expansion/src/test/java/no/unit/nva/expansion/ResourceExpansionServiceTest.java
```diff
@@ -321,6 +321,21 @@ void shouldGetOrganizationPartOfsForAffiliations(Class<? extends TicketEntry> ti
...
 
+    @ParameterizedTest
+    @MethodSource("no.unit.nva.publication.ticket.test.TicketTestUtils#ticketTypeAndPublicationStatusProvider")
+    void shouldGetOrganizationIdentifierForAffiliations(Class<? extends TicketEntry> ticketType,
+                                                     PublicationStatus status)
+        throws ApiGatewayException {
+        var publication = TicketTestUtils.createPersistedPublication(status, resourceService);
+        var ticket = TicketTestUtils.createPersistedTicket(publication, ticketType, ticketService);
+
+        var publicationOwnerId = publication.getResourceOwner().getOwnerAffiliation();
```
monaullah, line 332: var `publicationOwnerAffiliationId`?

expansion/src/main/java/no/unit/nva/expansion/ResourceExpansionServiceImpl.java
```diff
@@ -102,8 +103,13 @@ public ExpandedOrganization getOrganization(Entity dataEntry) throws NotFoundExc
                                       : resourceService.getResourceByIdentifier(ticketEntry.getResourceIdentifier())
                                           .getResourceOwner().getOwnerAffiliation();
 
+            var organizationIdentifier = Optional.ofNullable(organizationId)
```
monaullah, line 106: Can `organizationId` be `null`?
If no, maybe rewrite:
`var organizationIdentifier = UriWrapper.fromUri(organizationId).getLastPathElement();`

joachimjorgensen: There exists a test that specifically tests with null value here (and it fails when not checking for null-value):
[shouldUseResourceOwnerAffiliationWhenTicketHasNoOwnerAffiliation(...)](https://github.com/BIBSYSDEV/nva-publication-api/blob/main/expansion/src/test/java/no/unit/nva/expansion/ResourceExpansionServiceTest.java#L302)

StigNorland:
> Might discard this PR, as it could be a possibility to let search-api convert the identifier to an id in the endpoint instead. This would avoid the need of re-indexing. As per now, viewingScope is the only one that needs this identifier accessible from the search-api.

We should move away from id's in opensearch, there is no need to index id's that share the same 40-50 characters before the actual identifier. And, search urls becomes ridiculous long.

monaullah (approved):

brinxmat (approved):

StigNorland (approved):

PR #1613, opened 2024-04-09
Np 46542 update publisher when merging files brage migration

truhacevkir:
When importing publication from Brage and updating files for existing publication we do following:

- Updating publisher to publisher from Brage publication (customer)
- Updating resourceOwner to resourceOwner from Brage publication
  
   On persistence:
   - Removing existing publication from database
   - Persisting empty publication with publisher and resourceOwner from Brage publication
   - Update empty publication to "updated" existing publication

### DIFF
--- a/brage-import/src/main/java/no/sikt/nva/brage/migration/lambda/BrageEntryEventConsumer.java
+++ b/brage-import/src/main/java/no/sikt/nva/brage/migration/lambda/BrageEntryEventConsumer.java
@@ -258,9 +258,18 @@ public class BrageEntryEventConsumer implements RequestHandler<S3Event, Publicat
     }
 
     private BrageMergingReport persistInDatabaseAndCreateMergeReport(Publication publicationForUpdate,
-                                                                     Publication existinPublication) {
-        var newImage = resourceService.updatePublication(publicationForUpdate);
-        return new BrageMergingReport(existinPublication, newImage);
+                                                                     Publication existingPublication) {
+        if (resouseOwnerIsNotOverriden(publicationForUpdate, existingPublication)) {
+            var newImage = resourceService.updatePublication(publicationForUpdate);
+            return new BrageMergingReport(existingPublication, newImage);
+        } else {
+            var newImage = resourceService.recreateWith(publicationForUpdate);
+            return new BrageMergingReport(existingPublication, newImage);
+        }
+    }
+
+    private static boolean resouseOwnerIsNotOverriden(Publication publicationForUpdate, Publication existingPublication) {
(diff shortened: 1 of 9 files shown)


brage-import/src/main/java/no/sikt/nva/brage/migration/lambda/BrageEntryEventConsumer.java
```diff
@@ -258,9 +260,23 @@ private UriWrapper discardedFilesReportUri(Publication publication, S3Event s3Ev
...
+        var resourceOwner = publicationForUpdate.getResourceOwner();
+        var publisher = publicationForUpdate.getPublisher();
+        if (customerIsNotOverridden(existingPublication, resourceOwner, publisher)) {
+            var newImage = resourceService.updatePublication(publicationForUpdate);
+            return new BrageMergingReport(existingPublication, newImage);
+        } else {
+            resourceService.permanentlyRemovePublication(existingPublication);
+            resourceService.createEmptyPublicationForCustomer(resourceOwner, publisher, existingPublication.getIdentifier());
+            var newImage = resourceService.updatePublication(publicationForUpdate);
+            return new BrageMergingReport(existingPublication, newImage);
```
truhacevkir, line 273: This can be combined to one method in resourceService, but I think it is best to keep it like this, to make it visible whats going on. 
Decided to use existing method when "updating" publication.

publication-commons/src/main/java/no/unit/nva/publication/service/impl/UpdateResourceService.java
```diff
@@ -357,6 +372,25 @@ private TransactWriteItem updateResource(Resource resourceUpdate) {
...
+        keyToDelete.put("PK0", new AttributeValue(resourceDao.getPrimaryKeyPartitionKey()));
+        keyToDelete.put("SK0", new AttributeValue(resourceDao.getPrimaryKeySortKey()));
+        Delete delete = new Delete()
+                            .withTableName(tableName)
+                            .withKey(keyToDelete)
+                            .withConditionExpression(PRIMARY_KEY_EQUALITY_CHECK_EXPRESSION)
+                            .withExpressionAttributeNames(PRIMARY_KEY_EQUALITY_CONDITION_ATTRIBUTE_NAMES)
+                            .withExpressionAttributeValues(primaryKeyConditionAttributeValues);
+
+        return new TransactWriteItem().withDelete(delete);
```
anetteOlli, line 391: Create a transactWriteItem that contains both the removal and the insertion of the new publication so that we get an atomic operation.

anetteOlli (changes requested):

PR #1616, opened 2024-04-15
bump-nva-datamodel

anetteOlli:


brinxmat (approved):
We're setting a lot of nulls here…

PR #1622, opened 2024-04-17
merging brage publication

truhacevkir:
Adding new merge flow for Brage entries. 

- When we get single hit from search-api when searching for existing publication using incoming publication title and instanceType, we merge these two publications.

- Adding new path subpath to UPDATE_REPORTS for merged Brage publications. Adding merge source to update reports path, so we know the source/ cause of this merge. -> Ex. UPDATE_REPORTS/DOI  

- So from now on we will be able to see how many publications have been merged by cristinIdentifier, doi (via search-api) and publication content (search-api) as well:

```
public enum MergeSource {
    CRISTIN, DOI, SEARCH
}
```

### DIFF
--- a/brage-import/src/main/java/no/sikt/nva/brage/migration/lambda/BrageEntryEventConsumer.java
+++ b/brage-import/src/main/java/no/sikt/nva/brage/migration/lambda/BrageEntryEventConsumer.java
@@ -1,5 +1,6 @@
 package no.sikt.nva.brage.migration.lambda;
 
+import static java.net.HttpURLConnection.HTTP_OK;
 import static java.util.Objects.isNull;
 import static java.util.Objects.nonNull;
 import static nva.commons.core.attempt.Try.attempt;
@@ -9,6 +10,7 @@ import com.amazonaws.services.lambda.runtime.events.S3Event;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.google.common.collect.Iterables;
 import java.net.URI;
+import java.net.http.HttpResponse;
 import java.util.List;
 import java.util.Optional;
 import java.util.Random;
@@ -23,6 +25,7 @@ import no.sikt.nva.brage.migration.merger.DuplicatePublicationException;
 import no.sikt.nva.brage.migration.merger.UnmappableCristinRecordException;
 import no.sikt.nva.brage.migration.record.Record;
 import no.unit.nva.commons.json.JsonUtils;
+import no.unit.nva.identifiers.SortableIdentifier;
 import no.unit.nva.model.AdditionalIdentifier;
 import no.unit.nva.model.Publication;
 import no.unit.nva.model.exceptions.InvalidIsbnException;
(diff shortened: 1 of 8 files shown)


brage-import/src/main/java/no/sikt/nva/brage/migration/lambda/BrageEntryEventConsumer.java
```diff
@@ -66,14 +69,17 @@ public class BrageEntryEventConsumer implements RequestHandler<S3Event, Publicat
     private static final String S3_URI_TEMPLATE = "s3://%s/%s";
     private static final String ERROR_SAVING_BRAGE_IMPORT = "Error saving brage import for record with object key: ";
     private static final Logger logger = LoggerFactory.getLogger(BrageEntryEventConsumer.class);
-    private static final String RESOURCES_2 = "resources2";
+    private static final String RESOURCES = "resources";
```
truhacevkir, line 72: The `/resources2` has been moved to `/resource` a while back.

brage-import/src/main/java/no/sikt/nva/brage/migration/lambda/BrageEntryEventConsumer.java
```diff
@@ -109,10 +115,76 @@ private boolean shouldMergePublications(Publication publication) throws NotFound
...
+
+    private boolean existingPublicationHasSamePublicationContent(Publication publication) {
+        publicationsToMerge = searchForPublicationsByTypeAndTitle(publication);
+        return !publicationsToMerge.isEmpty();
+
+    }
+
+    private List<Publication> searchForPublicationsByTypeAndTitle(Publication publication) {
+        var response = fetchResponse(searchByTypeAndTitleUri(publication));
+        return response.map(this::toResponse)
```
brinxmat, line 141: Not unsurprisingly, there is a likelihood that this will miss more often than not. I'd consider a different approach with a similarity/distance metric between similar titles and for example (weirdly) the number of pages

brinxmat: Type is good though

brage-import/src/main/java/no/sikt/nva/brage/migration/lambda/BrageEntryEventConsumer.java
```diff
@@ -109,10 +115,76 @@ private boolean shouldMergePublications(Publication publication) throws NotFound
...
+                   .collect(Collectors.toList()).reversed();
+    }
+
+    private Publication getPublicationByIdentifier(SortableIdentifier identifier) {
+        return attempt(() -> resourceService.getPublicationByIdentifier(identifier)).orElseThrow();
+    }
+
+    private Optional<String> fetchResponse(URI uri) {
+        var response = uriRetriever.fetchResponse(uri);
+        if (!isHttpOk(response)) {
```
brinxmat, line 157: Do you use this without negation?

brinxmat (changes requested):

brage-import/src/main/java/no/sikt/nva/brage/migration/lambda/BrageEntryEventConsumer.java
truhacevkir: nope, fixing

PR #1623, opened 2024-04-17
fixing bucket path

truhacevkir:


template.yaml
```diff
@@ -1868,7 +1868,7 @@ Resources:
           DOMAIN_NAME: !Ref ApiDomain
           EVENT_BUS_NAME: !GetAtt InternalBus.Name
           CRISTIN_UNITS_S3_OBJECT_URI: !Ref CristinUnitsS3ObjectUri
-          CRISTIN_IMPORT_BUCKET: !Sub 's3://cristin-import-${AWS::AccountId}'
+          CRISTIN_IMPORT_BUCKET: !Sub 'cristin-import-${AWS::AccountId}'
```
truhacevkir, line 1871: It seems that it has to be bucket name only, `S3Driver` does not like `s3://`

publication-commons/src/main/java/no/unit/nva/publication/model/PublicationSummary.java
```diff
@@ -58,10 +58,16 @@ public static PublicationSummary create(Publication publication) {
...
-        publicationSummary.setAbstract(publication.getEntityDescription().getAbstract());
+        publicationSummary.setAbstract(extractAbstract(publication));
         return publicationSummary;
     }
 
+    private static String extractAbstract(Publication publication) {
+        return Optional.ofNullable(publication.getEntityDescription())
+                   .map(EntityDescription::getAbstract)
+                   .orElse(null);
+    }
```
truhacevkir, line 69: Endpoint returns drafts as well, so entityDescription is nullable.

PR #1628, opened 2024-04-23
fix(ListTicketsForPublicationHandler): forbidden response when publis…

torbjokv:
…her is different from current user #NP-46595

### DIFF
--- a/publication-commons/src/main/java/no/unit/nva/publication/service/impl/ResourceService.java
+++ b/publication-commons/src/main/java/no/unit/nva/publication/service/impl/ResourceService.java
@@ -18,7 +18,6 @@ import com.amazonaws.services.dynamodbv2.model.TransactWriteItem;
 import com.amazonaws.services.dynamodbv2.model.TransactWriteItemsRequest;
 import com.amazonaws.services.dynamodbv2.model.WriteRequest;
 import com.google.common.collect.Lists;
-import java.net.URI;
 import java.time.Clock;
 import java.time.Instant;
 import java.util.ArrayList;
@@ -32,6 +31,7 @@ import java.util.stream.Stream;
 import no.unit.nva.identifiers.SortableIdentifier;
 import no.unit.nva.model.Organization;
 import no.unit.nva.model.Publication;
+import no.unit.nva.model.PublicationOperation;
 import no.unit.nva.model.PublicationStatus;
 import no.unit.nva.publication.external.services.UriRetriever;
 import no.unit.nva.publication.model.DeletePublicationStatusResponse;
@@ -55,6 +55,7 @@ import no.unit.nva.publication.model.storage.TicketDao;
 import no.unit.nva.publication.model.storage.UniqueDoiRequestEntry;
 import no.unit.nva.publication.model.storage.WithPrimaryKey;
 import no.unit.nva.publication.model.utils.CuratingInstitutionsUtil;
+import no.unit.nva.publication.permission.strategy.PublicationPermissionStrategy;
 import no.unit.nva.publication.storage.model.DatabaseConstants;
 import nva.commons.apigateway.exceptions.ApiGatewayException;
 import nva.commons.apigateway.exceptions.BadMethodException;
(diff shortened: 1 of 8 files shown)


torbjokv:
We should look into making a dedicated permission strategy for tickets, but for now this is probably better than the current strategy that assumes publication.publisher matches user.customerId in fetchAllTicketsForElevatedUser

```java
// look up resource in fetchForElevatedUserQueryObject
Resource.builder().withIdentifier(resourceIdentifier)
                   .withPublisher(new Organization.Builder().withId(customerId).build())
                   .build()
```

publication-commons/src/test/java/no/unit/nva/publication/service/impl/TicketServiceTest.java
```diff
@@ -580,24 +585,27 @@ void shouldReturnAllTicketsForPublicationWhenRequesterIsPublicationOwner() throw
...
+                                               randomUri());
+
+        var fetchedTickets = resourceService.fetchAllTicketsForUser(elevatedUser, publication.getIdentifier())
                                  .collect(Collectors.toList());
         assertThat(fetchedTickets, containsInAnyOrder(originalTickets.toArray(TicketEntry[]::new)));
     }
 
     @Test
-    void shouldThrowNotFoundExceptionWhenAlienElevatedUserAttemptsToFetchPublicationTickets()
+    void shouldThrowNotForbiddenWhenAlienElevatedUserAttemptsToFetchPublicationTickets()
```
ketilaa, line 601: Typo in method name. `NotForbidden` -> `Forbidden`

publication-commons/src/main/java/no/unit/nva/publication/service/impl/ResourceService.java
```diff
@@ -337,10 +339,18 @@ public void refresh(SortableIdentifier identifier) {
...
-        var resource = fetchResourceForElevatedUser(userInstance.getCustomerId(), publicationIdentifier);
+    public Stream<TicketEntry> fetchAllTicketsForUser(UserInstance userInstance,
+                                                      SortableIdentifier publicationIdentifier)
+        throws NotFoundException, ForbiddenException {
+        var resource = getResourceByIdentifier(publicationIdentifier);
+
+        var permissionStrategy = PublicationPermissionStrategy.create(resource.toPublication(), userInstance,
+                                                                      uriRetriever);
+
+        if (!permissionStrategy.allowsAction(PublicationOperation.UPDATE)) {
```
ketilaa, line 350: Do this let contributors see all tickets for a publication too? Is that ok? I am not sure, so I ask to make sure it has been considered.

truhacevkir: I would move this validation out of `ResourceService` to handler, because now 
fetchAllTicketsForUser() does not return all tickets for user if you are not right user. I think it makes it clear when validation is lying on the top layer + single responsibility/ separation of concerns .

brinxmat (commented):
I think @truhacevkir's comment is correct

tickets/src/test/java/no/unit/nva/publication/ticket/read/ListTicketsForPublicationHandlerTest.java
```diff
@@ -154,12 +173,26 @@ void shouldListTicketsOfTypeCuratorHasAccessRightToOperateOn(
         assertThat(body.getTickets().getFirst().ticketType(), is(equalTo(ticketType)));
     }
 
+    private static Set<URI> getCuratingInstitutions(Publication publication) {
```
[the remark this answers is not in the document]
torbjokv: removed one of the methods

torbjokv:
> Generally, would it be better to return 200 Ok with empty list if user do not have access to see any tickets? 4xx results in an error message in the frontend...

discussed with Kjetil and for now we think its nice to expose this extra information to the client/consumer

publication-commons/src/main/java/no/unit/nva/publication/service/impl/ResourceService.java
torbjokv: I agree! refactored it now 👍

truhacevkir (approved):

PR #1632, opened 2024-04-24
removing duplicated tags when scraping tags from Cristin entry

truhacevkir:
Removing duplicated tags from Cristin entry

### DIFF
--- a/cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinMapper.java
+++ b/cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinMapper.java
@@ -26,11 +26,13 @@ import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashSet;
 import java.util.List;
+import java.util.LinkedHashMap;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import no.unit.nva.cristin.mapper.artisticproduction.CristinArtisticProduction;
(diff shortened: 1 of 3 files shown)


cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinMapper.java
```diff
@@ -527,21 +527,13 @@ private String extractAbstract() {
...
-                listOfTags.add(cristinTags.getEnglish());
-            }
-            if (cristinTags.getNynorsk() != null) {
-                listOfTags.add(cristinTags.getNynorsk());
-            }
-        }
-        return listOfTags;
+        return tags.stream()
+                   .flatMap(tag -> Stream.of(tag.getBokmal(), tag.getEnglish(), tag.getNynorsk()))
+                   .filter(Objects::nonNull)
```
anetteOlli, line 536: .distinct().toList().

anetteOlli, line 536: Maybe also trim the strings before checking for unique strings?

anetteOlli (changes requested):

cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinMapper.java
truhacevkir: Adding following:

```
tags.stream()
                   .flatMap(tag -> Stream.of(tag.getBokmal(), tag.getEnglish(), tag.getNynorsk()))
                   .filter(Objects::nonNull)
                   .collect(Collectors.toMap(String::trim, Function.identity(), (a, b) -> a, LinkedHashMap::new))
                   .values()
                   .stream()
                   .toList();
```

It trims string before comparison, removes duplicates, but return original values.

anetteOlli (approved):

PR #1634, opened 2024-04-25
mapping articleNumber for journal article

truhacevkir:


cristin-import/src/main/java/no/unit/nva/cristin/mapper/JournalBuilder.java
```diff
@@ -94,15 +94,15 @@ private PublicationInstance<? extends Pages> createJournalArticle() {
 
         var secondaryCategory = getCristinObject().getSecondaryCategory();
         if (CristinSecondaryCategory.JOURNAL_ARTICLE.equals(secondaryCategory)) {
-            return new ProfessionalArticle(numberOfPages, extractVolume(), extractIssue(), null);
+            return new ProfessionalArticle(numberOfPages, extractVolume(), extractIssue(), extractArticleNumber());
```
anetteOlli, line 97: Can null pointers occur in the extractArticleNumber() method?

anetteOlli (approved):

cristin-import/src/main/java/no/unit/nva/cristin/mapper/JournalBuilder.java
truhacevkir: Wrapping in optional of nullable

PR #1639, opened 2024-05-02
updating some more fields when updating existing Cristin publication

truhacevkir:


cristin-import/src/main/java/no/unit/nva/cristin/lambda/PublicationUpdater.java
```diff
@@ -30,9 +31,45 @@ public static PublicationRepresentations update(PublicationRepresentations publi
...
     }
 
+    private static List<ResearchProject> updateProjects(PublicationRepresentations publicationRepresentations) {
+        var existingProjects = publicationRepresentations.getExistingPublication().getProjects();
+        var incomingProjects = publicationRepresentations.getExistingPublication().getProjects();
+        return shouldBeUpdated(existingProjects, incomingProjects) ? incomingProjects : existingProjects;
+    }
+
+    private static boolean shouldBeUpdated(List<?> oldList, List<?> newList) {
+        return oldList.isEmpty() && !newList.isEmpty();
```
torbjokv: null check maybe?

torbjokv: as long as we don't allow cristin writes while importing I guess this Is ok

torbjokv (approved):

cristin-import/src/main/java/no/unit/nva/cristin/lambda/PublicationUpdater.java
torbjokv: (as long as you add null check that is)

cristin-import/src/main/java/no/unit/nva/cristin/lambda/PublicationUpdater.java
```diff
@@ -30,9 +31,45 @@ public static PublicationRepresentations update(PublicationRepresentations publi
...
+        existingPublication.setFundings(updateFundings(publicationRepresentations));
+        existingPublication.setProjects(updateProjects(publicationRepresentations));
+        existingPublication.setHandle(updateHandle(publicationRepresentations));
+        existingPublication.setLink(updateLink(publicationRepresentations));
         return publicationRepresentations;
     }
 
+    private static List<ResearchProject> updateProjects(PublicationRepresentations publicationRepresentations) {
+        var existingProjects = publicationRepresentations.getExistingPublication().getProjects();
+        var incomingProjects = publicationRepresentations.getExistingPublication().getProjects();
```
joachimjorgensen, line 43:
```suggestion
        var incomingProjects = publicationRepresentations.getIncomingPublication().getProjects();
```

joachimjorgensen: Might need to update some tests so that this would've been caught

PR #1641, opened 2024-05-02
Fixing last pr

truhacevkir:


cristin-import/src/test/java/no/unit/nva/cristin/lambda/CristinEntryEventConsumerTest.java
```diff
@@ -606,6 +606,19 @@ void shouldAddNewFieldsWhenExistingPublicationIsEmpty() throws IOException {
...
+    void shouldNotOverrideCuratingInstitutionsWhenUpdatingExistingPublication() throws IOException {
+        var cristinObject = CristinDataGenerator.randomObject();
+        var existingPublication = persistPublicationWithCristinId(cristinObject.getId(), Lecture.class);
+        var eventBody = createEventBody(cristinObject);
+        var sqsEvent = createSqsEvent(eventBody);
+        var updatedPublication = handler.handleRequest(sqsEvent, CONTEXT).getFirst();
+
+
+        assertThat(existingPublication.getCuratingInstitutions(),
+                is(equalTo(updatedPublication.getCuratingInstitutions())));
```
joachimjorgensen, line 619: Won't this test always pass? Shouldn't there be a line trying to override the `curatingInstitutions`? With my untrained eye it looks like we are persisting a publication, fetching it, and essentially just evaluating if the fetched  publication has the same property as the persisted one.
(I might be entirely wrong though, but just from the test it looks a bit weird)

joachimjorgensen (approved):

cristin-import/src/test/java/no/unit/nva/cristin/lambda/CristinEntryEventConsumerTest.java
truhacevkir: We first persist a publication that have currating institution:

`var existingPublication = persistPublicationWithCristinId(cristinObject.getId(), Lecture.class);`

After that we create event that updates existing publication:

```
        var eventBody = createEventBody(cristinObject);
        var sqsEvent = createSqsEvent(eventBody);
        var updatedPublication = handler.handleRequest(sqsEvent, CONTEXT).getFirst();
```

Event for resource to be created and existing publication have the same CristinId, so instead of creating new publication,  `handler.handleRequest(sqsEvent, CONTEXT).getFirst()` will update existing one. And then one make a check that curratingInstitutions should not be overriden after update.

PR #1646, opened 2024-05-07
mapping npi subject heading for chapter

truhacevkir:
- Mapping `npiSubjectHeading` for `Chapter`
- Removing exception when npiSubjectHeading is missing for `Book` or `Report`
- Updating npiSubjectHeading when existing publication is missing npiSubjectHeading

### DIFF
--- a/cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinBookOrReportPartMetadata.java
+++ b/cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinBookOrReportPartMetadata.java
@@ -19,14 +19,14 @@ import nva.commons.core.JacocoGenerated;
 @Getter
 @Setter
 @AllArgsConstructor(access = AccessLevel.PACKAGE)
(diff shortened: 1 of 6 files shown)


cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinBookOrReportPartMetadata.java
```diff
@@ -19,14 +19,14 @@
 @Getter
 @Setter
 @AllArgsConstructor(access = AccessLevel.PACKAGE)
-@JsonIgnoreProperties({"antall_sider_totalt",
-    "status_utgitt_av_forlag", "delangivelse", "sprakkode_oversatt_fra", "fagfelt"})
+@JsonIgnoreProperties({"antall_sider_totalt", "status_utgitt_av_forlag", "delangivelse", "sprakkode_oversatt_fra"})
```
anetteOlli, line 22: Flott at du husket denne.

anetteOlli (approved):

PR #1650, opened 2024-05-10
Persist updated nvi report

truhacevkir:
- Persisting nvi report both on creation and update. 
- Makes it possible to generate data for "new" nvi-import, so each to we run cristin-import, we are also able to run new nvi import.

### DIFF
--- a/cristin-import/src/main/java/no/unit/nva/cristin/lambda/CristinEntryEventConsumer.java
+++ b/cristin-import/src/main/java/no/unit/nva/cristin/lambda/CristinEntryEventConsumer.java
@@ -159,10 +159,17 @@ public class CristinEntryEventConsumer
         return attempt(() -> getExistingPublication(publicationRepresentation))
                    .map(publicationRepresentation::withExistingPublication)
                    .map(PublicationUpdater::update)
-                   .map(publicationRepresentations -> publicationRepresentations.updateHasEffectiveChanges()
-                                                          ? update(publicationRepresentations)
-                                                          : publicationRepresentations.getExistingPublication())
+                   .map(this::getPublication)
                    .orElseThrow();
+    }
+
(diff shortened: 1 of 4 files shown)


cristin-import/src/main/java/no/unit/nva/cristin/lambda/CristinEntryEventConsumer.java
```diff
@@ -159,15 +159,22 @@ private Publication performUpdate(PublicationRepresentations publicationRepresen
...
                    .map(PublicationUpdater::update)
-                   .map(publicationRepresentations -> publicationRepresentations.updateHasEffectiveChanges()
-                                                          ? update(publicationRepresentations)
-                                                          : publicationRepresentations.getExistingPublication())
+                   .map(this::getPublication)
                    .orElseThrow();
     }
 
+    private Publication getPublication(PublicationRepresentations publicationRepresentations) {
+        if (publicationRepresentations.updateHasEffectiveChanges()) {
```
brinxmat, line 167: the "ifNeeded" in the else indicate that there is some logic for unneeded that may be missing (and I can't see)

truhacevkir: I wanted to indicate that we do not persist nvi report for "all" entries, but only those which satisfy some conditions, i.e. are nvi reported publications.
private void persistNviReportIfNeeded(PublicationRepresentations publicationRepresentations) {
        if (hasScientificResource(publicationRepresentations)) {
            var nviReport = NviReport.fromPublicationRepresentation(publicationRepresentations);
            saveNviReport(publicationRepresentations, nviReport);
        }
    }

brinxmat (approved):

PR #1651, opened 2024-05-10
Calculating Levenshtein distance between title when merging

truhacevkir:
- We calculate the Levenshtein Distance between the main titles of publications when merging. We assume that a distance of >10 is not significant, so we merge the publications

### DIFF
--- a/brage-import/src/main/java/no/sikt/nva/brage/migration/lambda/BrageEntryEventConsumer.java
+++ b/brage-import/src/main/java/no/sikt/nva/brage/migration/lambda/BrageEntryEventConsumer.java
@@ -45,6 +45,7 @@ import nva.commons.core.StringUtils;
 import nva.commons.core.attempt.Failure;
 import nva.commons.core.attempt.Try;
 import nva.commons.core.paths.UriWrapper;
+import org.apache.commons.text.similarity.LevenshteinDistance;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import software.amazon.awssdk.services.s3.S3Client;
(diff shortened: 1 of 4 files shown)


brage-import/src/test/java/no/sikt/nva/brage/migration/lambda/BrageEntryEventConsumerTest.java
```diff
@@ -1065,18 +1069,19 @@ void shouldSaveSuccessReportContainingIncomingHandleWhenUpdatingExistingPublicat
...
 
-    //TODO: flaky test
     @Test
     void shouldPersistMergeReport()
         throws IOException, BadRequestException, nva.commons.apigateway.exceptions.NotFoundException {
         var cristinIdentifier = randomString();
         var nvaBrageMigrationDataGenerator = new NvaBrageMigrationDataGenerator.Builder().withPublishedDate(null)
-                                                 .withType(TYPE_BOOK)
+                                                 .withType(TYPE_FILM)
                                                  .withCristinIdentifier(cristinIdentifier)
```
truhacevkir, line 1078: Fixing this flaky test. Random book with ISBN makes it difficult to create book with ISBN, as random isbn is often not valid isbn and ends up as additionalIdentifier instead of ISBN.

truhacevkir: Pr that fixes that:

https://github.com/BIBSYSDEV/nva-datamodel-java/pull/484

brage-import/src/main/java/no/sikt/nva/brage/migration/lambda/BrageEntryEventConsumer.java
```diff
@@ -79,6 +80,7 @@ public class BrageEntryEventConsumer implements RequestHandler<S3Event, Publicat
     public static final String AGGREGATION = "aggregation";
     public static final String NONE = "none";
     public static final String ISBN = "isbn";
+    public static final int MAX_ACCEPTABLE_LEVENSHTEIN_DISTANCE = 10;
```
truhacevkir, line 83: This means that we have to remove max 10 chars from string to achieve to similar string

brinxmat (approved):

PR #1652, opened 2024-05-13
fix(NonDegreePermissionStrategy): allow third party client access to degrees

torbjokv:
and:
refactor(PublicationPermissionStrategyTest): split in own classes #NP-46258

### DIFF
--- a/publication-commons/src/main/java/no/unit/nva/publication/permission/strategy/restrict/NonDegreePermissionStrategy.java
+++ b/publication-commons/src/main/java/no/unit/nva/publication/permission/strategy/restrict/NonDegreePermissionStrategy.java
@@ -15,7 +15,7 @@ public class NonDegreePermissionStrategy extends DenyPermissionStrategy {
 
     @Override
     public boolean deniesAction(PublicationOperation permission) {
-        if (isUsersDraft()) {
+        if (isUsersDraft() || userInstance.isExternalClient()) {
             return false; // allow
         }
 
--- a/publication-commons/src/test/java/no/unit/nva/publication/permission/strategy/ContributorPermissionStrategyTest.java
+++ b/publication-commons/src/test/java/no/unit/nva/publication/permission/strategy/ContributorPermissionStrategyTest.java
@@ -59,21 +59,4 @@ class ContributorPermissionStrategyTest extends PublicationPermissionStrategyTes
                                    .create(publication, userInstance, uriRetriever)
                                    .allowsAction(operation));
     }
-
-    @Disabled("Not valid anymore?")
-    @Test
-    void shouldDenyPermissionToUnpublishPublicationWhenUserIsContributorButNotCreator()
-        throws JsonProcessingException, UnauthorizedException {
-        var contributorName = randomString();
-        var contributorCristinId = randomUri();
-        var contributorInstitutionId = randomUri();
-
-        var requestInfo = createUserRequestInfo(contributorName, contributorInstitutionId, contributorCristinId, randomUri());
-        var publication = createPublicationWithContributor(contributorName, contributorCristinId, null, randomUri(), randomUri());
-
-        Assertions.assertFalse(PublicationPermissionStrategy
-                                   .create(publication, RequestUtil.createUserInstanceFromRequest(
-                                       requestInfo, identityServiceClient), uriRetriever)
-                                   .allowsAction(UNPUBLISH));
-    }
 }
--- a/publication-commons/src/test/java/no/unit/nva/publication/permission/strategy/ContributorPermissionStrategyTest.java
+++ b/publication-commons/src/test/java/no/unit/nva/publication/permission/strategy/ContributorPermissionStrategyTest.java
@@ -10,7 +10,6 @@ import no.unit.nva.model.role.Role;
 import no.unit.nva.publication.RequestUtil;
 import nva.commons.apigateway.exceptions.UnauthorizedException;
 import org.junit.jupiter.api.Assertions;
-import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.EnumSource;
--- a/publication-commons/src/test/java/no/unit/nva/publication/permission/strategy/ContributorPermissionStrategyTest.java
+++ b/publication-commons/src/test/java/no/unit/nva/publication/permission/strategy/ContributorPermissionStrategyTest.java
@@ -1,0 +1,79 @@
+package no.unit.nva.publication.permission.strategy;
+
+import static no.unit.nva.model.PublicationOperation.UNPUBLISH;
+import static no.unit.nva.testutils.RandomDataGenerator.randomString;
+import static no.unit.nva.testutils.RandomDataGenerator.randomUri;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import no.unit.nva.model.PublicationOperation;
+import no.unit.nva.model.associatedartifacts.AssociatedArtifactList;
+import no.unit.nva.model.role.Role;
+import no.unit.nva.publication.RequestUtil;
+import nva.commons.apigateway.exceptions.UnauthorizedException;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.EnumSource;
+import org.junit.jupiter.params.provider.EnumSource.Mode;
+
+class ContributorPermissionStrategyTest extends PublicationPermissionStrategyTest {
+
+    @ParameterizedTest(name = "Should allow verified contributor {0} operation on non-degree resources")
+    @EnumSource(value = PublicationOperation.class, mode = Mode.EXCLUDE, names = {"DELETE", "TERMINATE",
+        "TICKET_PUBLISH"})
+    void shouldAllowVerifiedContributorOnNonDegree(PublicationOperation operation)
+        throws JsonProcessingException, UnauthorizedException {
+
+        var institution = randomUri();
+        var contributor = randomString();
+        var cristinId = randomUri();
+
+        var requestInfo = createUserRequestInfo(contributor, institution, cristinId, randomUri());
+        var publication = createPublicationWithContributor(contributor, cristinId, Role.CREATOR,
+                                                           randomUri(), randomUri());
+        publication.setAssociatedArtifacts(new AssociatedArtifactList());
+
+        var userInstance = RequestUtil.createUserInstanceFromRequest(requestInfo, identityServiceClient);
+
+
+        Assertions.assertTrue(PublicationPermissionStrategy
+                                  .create(publication, userInstance, uriRetriever)
+                                  .allowsAction(operation));
+    }
+
+    @ParameterizedTest(name = "Should deny verified contributor {0} operation on non-degree resources")
+    @EnumSource(value = PublicationOperation.class, mode = Mode.EXCLUDE, names = {"UNPUBLISH", "UPDATE"})
+    void shouldDenyVerifiedContributorOnNonDegree(PublicationOperation operation)
+        throws JsonProcessingException, UnauthorizedException {
+
+        var institution = randomUri();
+        var contributor = randomString();
+        var cristinId = randomUri();
+
+        var requestInfo = createUserRequestInfo(contributor, institution, cristinId, randomUri());
+        var publication = createPublicationWithContributor(contributor, cristinId, Role.CREATOR,
+                                                           randomUri(), randomUri());
+        var userInstance = RequestUtil.createUserInstanceFromRequest(requestInfo, identityServiceClient);
+
+        Assertions.assertFalse(PublicationPermissionStrategy
+                                   .create(publication, userInstance, uriRetriever)
+                                   .allowsAction(operation));
+    }
+
+    @Disabled("Not valid anymore?")
+    @Test
+    void shouldDenyPermissionToUnpublishPublicationWhenUserIsContributorButNotCreator()
+        throws JsonProcessingException, UnauthorizedException {
+        var contributorName = randomString();
+        var contributorCristinId = randomUri();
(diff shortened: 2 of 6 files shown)


publication-commons/src/main/java/no/unit/nva/publication/permission/strategy/restrict/NonDegreePermissionStrategy.java
```diff
@@ -15,7 +15,7 @@ public NonDegreePermissionStrategy(Publication publication, UserInstance userIns
 
     @Override
     public boolean deniesAction(PublicationOperation permission) {
-        if (isUsersDraft()) {
+        if (isUsersDraft() || userInstance.isExternalClient()) {
```
joachimjorgensen, line 18: Should there be a check to verify that the user has something to do with the publication? Like `isUsersDraft()` does? Now, a third party can access any degree. Unless that is the intention tho.

publication-commons/src/test/java/no/unit/nva/publication/permission/strategy/ContributorPermissionStrategyTest.java
```diff
@@ -0,0 +1,79 @@
...
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.EnumSource;
+import org.junit.jupiter.params.provider.EnumSource.Mode;
+
+class ContributorPermissionStrategyTest extends PublicationPermissionStrategyTest {
+
+    @ParameterizedTest(name = "Should allow verified contributor {0} operation on non-degree resources")
+    @EnumSource(value = PublicationOperation.class, mode = Mode.EXCLUDE, names = {"DELETE", "TERMINATE",
```
joachimjorgensen, line 22: Better to `INCLUDE` operations instead of `EXCLUDE`? Easier to read the test, as one don't need to lookup the other operations.

joachimjorgensen: Applies to the test below as well

publication-commons/src/test/java/no/unit/nva/publication/permission/strategy/ContributorPermissionStrategyTest.java
```diff
@@ -0,0 +1,79 @@
...
+        throws JsonProcessingException, UnauthorizedException {
+
+        var institution = randomUri();
+        var contributor = randomString();
+        var cristinId = randomUri();
+
+        var requestInfo = createUserRequestInfo(contributor, institution, cristinId, randomUri());
+        var publication = createPublicationWithContributor(contributor, cristinId, Role.CREATOR,
+                                                           randomUri(), randomUri());
+        publication.setAssociatedArtifacts(new AssociatedArtifactList());
```
joachimjorgensen, line 34: Was wondering why this line had to be included for the test to pass, while the next test did not. In my eyes these two tests should've/could've had the same body. But it turns out that `randomPublicationNonDegree()` in `createPublicationWithContributor(...)` creates a publication with a published file, which is quite essential information in this test, as half of the UNPUBLISH-test relies on the publication containing a publish file or not. Might want to specify why this line had to be added.

publication-commons/src/test/java/no/unit/nva/publication/permission/strategy/ContributorPermissionStrategyTest.java
```diff
@@ -0,0 +1,79 @@
...
+        var publication = createPublicationWithContributor(contributor, cristinId, Role.CREATOR,
+                                                           randomUri(), randomUri());
+        var userInstance = RequestUtil.createUserInstanceFromRequest(requestInfo, identityServiceClient);
+
+        Assertions.assertFalse(PublicationPermissionStrategy
+                                   .create(publication, userInstance, uriRetriever)
+                                   .allowsAction(operation));
+    }
+
+    @Disabled("Not valid anymore?")
```
joachimjorgensen, line 63: I believe this one is not valid anymore. Should check with PO, but as I understand, everyone that can edit/update a publication, can also unpublish it (unless the publication has any published files, as mentioned in my comment above). Might be wise to get verification from OP though.

publication-commons/src/test/java/no/unit/nva/publication/permission/strategy/CuratorPermissionStrategyTest.java
```diff
@@ -87,6 +92,38 @@ void shouldDenyCuratorOnNonDegree(PublicationOperation operation)
...
+        var requestInfo = createUserRequestInfo(username, institution, getAccessRightsForCurator(), randomUri(), cristinId);
+        var publication = createNonDegreePublication(username, institution, cristinId);
+        var permissionStrategy = PublicationPermissionStrategy.create(publication,
+                                                                      RequestUtil.createUserInstanceFromRequest(requestInfo, identityServiceClient),
+                                                                      uriRetriever);
+        assertThat(permissionStrategy.isCuratorOnPublication(), is(equalTo(true)));
+    }
+
+    @Test
+    void shouldNotGivePermissionToUnpublishPublicationWithPublishedFilesWhenUserIsContributor()
```
joachimjorgensen, line 111: Does this test belong here? Shouldn't it be in `ContributorPermissionStrategy.java`?

publication-commons/src/test/java/no/unit/nva/publication/permission/strategy/EditorPermissionStrategyTest.java
```diff
@@ -0,0 +1,101 @@
...
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.EnumSource;
+import org.junit.jupiter.params.provider.EnumSource.Mode;
+
+class EditorPermissionStrategyTest extends PublicationPermissionStrategyTest {
+
+    @ParameterizedTest(name = "Should allow editor {0} operation on non-degree resources")
+    @EnumSource(value = PublicationOperation.class, mode = Mode.EXCLUDE, names = {"DELETE", "TICKET_PUBLISH"})
```
joachimjorgensen, line 21: `Mode.INCLUDE` instead?

publication-commons/src/test/java/no/unit/nva/publication/permission/strategy/EditorPermissionStrategyTest.java
```diff
@@ -0,0 +1,101 @@
...
+        var publication = createPublication(resourceOwner, editorInstitution, cristinId);
+
+        Assertions.assertTrue(PublicationPermissionStrategy
+                                  .create(publication, RequestUtil.createUserInstanceFromRequest(
+                                      requestInfo, identityServiceClient), uriRetriever)
+                                  .allowsAction(UNPUBLISH));
+    }
+
+    @Test
+    void shouldGiveEditorPermissionToUnpublishPublicationWhenPublicationIsFromAnotherInstitution()
```
joachimjorgensen, line 83: Is this correct? Might be, but just want to make sure, as this is much power.

publication-commons/src/test/java/no/unit/nva/publication/permission/strategy/EditorPermissionStrategyTest.java
```diff
@@ -0,0 +1,101 @@
...
+        throws JsonProcessingException, UnauthorizedException {
+
+        var editorName = randomString();
+        var editorInstitution = randomUri();
+        var resourceOwner = randomString();
+        var cristinId = randomUri();
+
+        var requestInfo = createUserRequestInfo(editorName, editorInstitution, getAccessRightsForEditor(),
+                                                cristinId, randomUri());
+        var publication = createPublication(resourceOwner, editorInstitution, cristinId);
```
joachimjorgensen, line 74: This publication contains a published file, which is very essential to the permission check (in generall, not editor specifically). An editor (and curator)  should be able to unpublish whether or not the publication has any published files. Should we make an explicit test for this? As it goes kind of unnoticed now.
Up to you.

publication-commons/src/test/java/no/unit/nva/publication/permission/strategy/TrustedThirdPartyStrategyTest.java
```diff
@@ -0,0 +1,119 @@
...
+        var requestInfo = createThirdPartyRequestInfo();
+
+        Assertions.assertFalse(
+            PublicationPermissionStrategy.create(publication, RequestUtil.createUserInstanceFromRequest(
+                    requestInfo, identityServiceClient), uriRetriever)
+                .allowsAction(UPDATE));
+    }
+
+    @Test
+    void shouldAllowTrustedClientEditPublicationOnDegree()
```
joachimjorgensen, line 108: On _any_ degree?

joachimjorgensen (commented):
A few comments. Some of them are just to be sure, as it is important that we get this right :)

publication-commons/src/main/java/no/unit/nva/publication/permission/strategy/restrict/NonDegreePermissionStrategy.java
brinxmat: I'd verify with PO that the scope is "own publications", where "own" needs to be defined

brinxmat (changes requested):
See other reviewer's comments, I think PO verification is required on a few of these. I _believe_ that Editors can do things across owner-institution boundaries though

publication-commons/src/main/java/no/unit/nva/publication/permission/strategy/restrict/NonDegreePermissionStrategy.java
torbjokv: This change does not allow anything, its just make sure that an external client is not stopped. This is a restrict policy and external clients are today mostly handled on scope and API level granularity, not really doing much with clients inside the strategies.

publication-commons/src/test/java/no/unit/nva/publication/permission/strategy/ContributorPermissionStrategyTest.java
torbjokv: I agree that the readability is worse, but it's done by design so that a change where something is added, it will make stuff fail unless you have taken explicit care of what the test should do. It's of course up for discussion if it's worth it or not.

publication-commons/src/test/java/no/unit/nva/publication/permission/strategy/ContributorPermissionStrategyTest.java
torbjokv: agree! will rewrite a bit!

publication-commons/src/test/java/no/unit/nva/publication/permission/strategy/ContributorPermissionStrategyTest.java
torbjokv: ahh yes I have deleted it before, but a lot of the code here was from an old PR so it snuck in again, will remove

publication-commons/src/test/java/no/unit/nva/publication/permission/strategy/CuratorPermissionStrategyTest.java
torbjokv: good point, will move

publication-commons/src/test/java/no/unit/nva/publication/permission/strategy/TrustedThirdPartyStrategyTest.java
torbjokv: as long as userInstance.getCustomerId().equals(publication.getPublisher().getId()), as defined in TrustedThirdPartyStrategy. this is not the best check, but probably close enough

torbjokv: will update title of test

publication-commons/src/test/java/no/unit/nva/publication/permission/strategy/ContributorPermissionStrategyTest.java
joachimjorgensen: Ah, didn't think of it that way. Makes sense!

publication-commons/src/test/java/no/unit/nva/publication/permission/strategy/EditorPermissionStrategyTest.java
joachimjorgensen: Ignore this one, as I agree with your comment above.

publication-commons/src/test/java/no/unit/nva/publication/permission/strategy/TrustedThirdPartyStrategyTest.java
joachimjorgensen: Ah, must admit I didn't see that bit. My bad. That looks sound! Would a test `shouldDenyTrustedClientEditPublicationOnDegreeFromAnotherPublisher()` be a good addition? I can see that you have a test for if `publisher == null`, which might be enough? 🤷‍♂️ Your call.

publication-commons/src/main/java/no/unit/nva/publication/permission/strategy/restrict/NonDegreePermissionStrategy.java
joachimjorgensen: From your other comment I can see now that third parties without the correct credentials will be "stopped" by the ThirdPartyStrategy  👍  A little bit unfortunate that third parties has to be taken care of here as well, but I can't see another way atm.

publication-commons/src/test/java/no/unit/nva/publication/permission/strategy/TrustedThirdPartyStrategyTest.java
torbjokv: added shouldDenyTrustedClientEditPublicationOnDegreeFromAnotherPublisher 👍

torbjokv:
suggesting doing eventual changes to editor if needed in dedicated PR as its not the scope of this PR.

brinxmat (approved):

PR #1653, opened 2024-05-14
renaming some values

truhacevkir:
Updating values names for PresignedUrl so it matches one from download-file-api.


publication-rest/src/main/java/no/unit/nva/publication/download/PresignedUri.java
```diff
@@ -14,39 +14,39 @@
 import software.amazon.awssdk.services.s3.presigner.model.PresignedGetObjectRequest;
 
 @JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
-@JsonTypeName(PresignedUrl.TYPE)
-public record PresignedUrl(UUID fileIdentifier, @JsonIgnore String bucket, URI uri, Instant expires) {
+@JsonTypeName(PresignedUri.TYPE)
+public record PresignedUri(UUID fileIdentifier, @JsonIgnore String bucket, URI id, Instant expires) {
```
truhacevkir, line 18: Changing only `PresignedUrl` -> `PresignedUri` and `uri` -> `id`

PR #1655, opened 2024-05-14
explisit set style and explode

sondrev:


docs/openapi.yaml
```diff
@@ -865,14 +922,17 @@ components:
...
+                format: uri
               name:
                 type: string
         description:
           type: string
         doi:
           type: string
-        id:
+          format: uri
+        identifier:
```
sondrev, line 933: I could not find a "id" on publication. It should be "identifier" - right?

ketilaa: PublicationResponse implements WithId that adds the `id` field.

docs/openapi.yaml
```diff
@@ -304,7 +314,11 @@ paths:
         name: publicationIdentifier
         schema:
           type: string
+          format: uuid
```
ketilaa, line 317: Identifier is not full URI, just the SortableIdentifier (UUID-like)

docs/openapi.yaml
```diff
@@ -341,12 +355,20 @@ paths:
...
```
ketilaa, line 358: probably not correct

docs/openapi.yaml
```diff
@@ -341,12 +355,20 @@ paths:
...
         required: true
+        style: simple
+        explode: false
+        description: UUID of the publication
       - in: path
         name: ticketIdentifier
         schema:
           type: string
+          format: uuid
```
ketilaa, line 367: probably not correct

docs/openapi.yaml
```diff
@@ -463,12 +485,20 @@ paths:
         name: publicationIdentifier
         schema:
           type: string
+          format: uuid
```
ketilaa, line 488: probably not correct

docs/openapi.yaml
```diff
@@ -463,12 +485,20 @@ paths:
...
         required: true
+        style: simple
+        explode: false
+        description: UUID of the publication
       - in: path
         name: ticketIdentifier
         schema:
           type: string
+          format: uuid
```
ketilaa, line 497: probably not correct

docs/openapi.yaml
```diff
@@ -505,17 +536,29 @@ paths:
         name: publicationIdentifier
         schema:
           type: string
+          format: uuid
```
ketilaa, line 539: probably not correct

docs/openapi.yaml
```diff
@@ -505,17 +536,29 @@ paths:
...
         required: true
+        style: simple
+        explode: false
+        description: UUID of the publication
       - in: path
         name: ticketIdentifier
         schema:
           type: string
+          format: uuid
```
ketilaa, line 548: probably not correct

docs/openapi.yaml
```diff
@@ -505,17 +536,29 @@ paths:
...
         required: true
+        style: simple
+        explode: false
+        description: UUID of the ticket
       - in: path
         name: messageIdentifier
         schema:
           type: string
+          format: uuid
```
ketilaa, line 557: probably not correct

docs/openapi.yaml
```diff
@@ -580,7 +623,11 @@ paths:
         name: publicationIdentifier
         schema:
           type: string
+          format: uuid
```
ketilaa, line 626: I know we do some "magic" with our identifiers to make them sortable, so it is not a standard UUID.

docs/openapi.yaml
```diff
@@ -865,14 +922,17 @@ components:
...
                 type: string
         description:
           type: string
         doi:
           type: string
-        id:
+          format: uri
+        identifier:
           type: string
+          format: uri
```
ketilaa, line 935: identifier is typically not an uri

docs/openapi.yaml
```diff
@@ -1049,13 +1113,15 @@ components:
...
               type: string
+              format: uri
             name:
               type: string
         reference:
           type: object
           properties:
             doi:
               type: string
+              format: uuid
```
ketilaa, line 1124: doi is a `uri`

ketilaa (commented):
Generally, `id`, `xxxId`, `doi` and `handle` is `uri`.
Identifier is just string. I think we extend UUID with some time info to make it sortable, so better not set a format there.

ketilaa (approved):

PR #1660, opened 2024-05-21
mapping book of abstracts

truhacevkir:
Mapping book of abstracts in brage-migration

### DIFF
--- a/brage-import/src/main/java/no/sikt/nva/brage/migration/NvaType.java
+++ b/brage-import/src/main/java/no/sikt/nva/brage/migration/NvaType.java
@@ -5,6 +5,7 @@ import nva.commons.core.JacocoGenerated;
 @JacocoGenerated
 public enum NvaType {
     BOOK("Book"),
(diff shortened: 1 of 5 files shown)


brage-import/src/main/java/no/sikt/nva/brage/migration/NvaType.java
```diff
@@ -5,6 +5,7 @@
 @JacocoGenerated
 public enum NvaType {
     BOOK("Book"),
+    BOOK_OF_ABSTRACTS("Book of abstracts"),
```
anetteOlli, line 8: BookOfAbstracts, whitespace enums for the frontend is a pain to handle.

anetteOlli (approved):

brage-import/src/main/java/no/sikt/nva/brage/migration/NvaType.java
truhacevkir: This is temporary value -> this will be mapped to InstanceType: ReportBookOfAbstracts in NVA.

PR #1664, opened 2024-05-22
mapping full serieNumber from Brage

truhacevkir:
Mapping full series number from Brage, as example:

Before: 42:2022 has been mapped to 42
Now: 42:2022 will be mapped to 42:2022


brage-import/src/test/java/no/sikt/nva/brage/migration/lambda/BrageEntryEventConsumerTest.java
```diff
@@ -245,6 +247,15 @@ public class BrageEntryEventConsumerTest extends ResourcesLocalTest {
     private FakeS3Client s3Client;
     private ResourceService resourceService;
 
+    public static Stream<Arguments> seriesNumberProvider() {
```
anetteOlli, line 250: Nice!

anetteOlli (approved):
Nice solution for a horrible problem

PR #1669, opened 2024-05-27
Brage patch entry event emitter

truhacevkir:
Brage patch entry event emitter


template.yaml
```diff
@@ -1676,7 +1677,7 @@ Resources:
             Pattern:
               detail:
                 topic: [ "PublicationService.DataImport.Filename" ]
-                subtopic: [ "PublicationService.CristinData.DataEntry", "PublicationService.CristinData.PatchEntry", "PublicationService.CristinData.NviPatch" ]
+                subtopic: [ "PublicationService.CristinData.DataEntry", "PublicationService.CristinData.PatchEntry", "PublicationService.CristinData.NviPatch", "PublicationService.BrageData.PatchEntry" ]
```
truhacevkir, line 1680: This Lambda function is currently designed for Cristin-import, but it essentially acts as an event emitter for S3 objects. If necessary, we can duplicate this Lambda function to create a separate handler specifically for emitting Brage events.

anetteOlli (approved):

PR #1671, opened 2024-05-28
NP-46803-expand-channel-registry-journal

anetteOlli:
PR skal sørge for at journal name på importcandidate kommer med på søke-index.

### DIFF
--- a/expansion/src/main/java/no/unit/nva/expansion/model/ExpandedJournal.java
+++ b/expansion/src/main/java/no/unit/nva/expansion/model/ExpandedJournal.java
@@ -4,54 +4,10 @@ import com.fasterxml.jackson.annotation.JsonTypeInfo;
 import com.fasterxml.jackson.annotation.JsonTypeInfo.Id;
 import com.fasterxml.jackson.annotation.JsonTypeName;
 import java.net.URI;
-import java.util.Objects;
-import nva.commons.core.JacocoGenerated;
 
 @JsonTypeInfo(use = Id.NAME, property = "type")
 @JsonTypeName(ExpandedJournal.TYPE)
-public class ExpandedJournal {
+public record ExpandedJournal(URI id, String name) {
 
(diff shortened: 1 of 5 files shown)


expansion/src/main/java/no/unit/nva/expansion/model/ExpandedJournal.java
```diff
@@ -0,0 +1,110 @@
...
+
+    private Optional<String> expandJournalName(Journal journal,
+                                               RawContentRetriever rawContentRetriever) {
+        return nonNull(journal) && nonNull(journal.getId())
+                   ? expandJournalId(journal, rawContentRetriever)
+                   : Optional.empty();
+    }
+
+    private Optional<String> expandJournalId(Journal journal, RawContentRetriever rawContentRetriever) {
+        return rawContentRetriever.fetchResponse(journal.getId(), APPLICATION_JSON)
```
sondrev, line 84: I would recommend making ExpandedJournal thinner. Only have the necesary fields, and let a "JournalExpansionService" do the logic

expansion/src/main/java/no/unit/nva/expansion/model/ExpandedJournal.java
```diff
@@ -0,0 +1,57 @@
...
+import nva.commons.core.JacocoGenerated;
+
+@JsonTypeInfo(use = Id.NAME, property = "type")
+@JsonTypeName(ExpandedJournal.TYPE)
+public class ExpandedJournal {
+
+    public static final String TYPE = "Journal";
+    public URI id;
+
+    public String name;
```
truhacevkir, line 17: Should be private, whole class can be converted to record:

```
@JsonTypeName(ExpandedJournal.TYPE)
public record ExpandedJournal(URI id, String name) {

    public static final String TYPE = "Journal";
}
```

truhacevkir (approved):
Nice!

PR #1679, opened 2024-05-31
update all sets in Brage objects to lists

truhacevkir:


brage-import/src/main/java/no/sikt/nva/brage/migration/lambda/BrageEntryEventConsumer.java
```diff
@@ -75,11 +75,10 @@ public class BrageEntryEventConsumer implements RequestHandler<S3Event, Publicat
     private static final String DOI = "doi";
     private static final String APPLICATION_JSON = "application/json";
     public static final String TITLE = "title";
-    public static final String INSTANCE_TYPE = "instanceType";
+    public static final String CONTEXT_TYPE = "contextType";
```
truhacevkir, line 78: Did not update this one in last pr when updating comparing of instanceType to publicationContext.type.
Can be smart to mock exact uri's so the tests catch up this changes :)

joachimjorgensen (approved):

brinxmat (approved):

PR #1686, opened 2024-06-05
NP-46950 Add filesForAapproval field to ticket dto

joachimjorgensen:


ketilaa (approved):
Are we tracking both `filesForApproval` and `approvedFiles`? I guess they can be different when ticket is created compared to when it is approved/completed?

joachimjorgensen:
Yes, whenever a `PublishingRequest` is created, `filesForApproval` is populated from unpublished files in the publication. When https://github.com/BIBSYSDEV/nva-publication-api/blob/b8d5b979091038f948896b95085a72d937721279/publication-commons/src/main/java/no/unit/nva/publication/model/business/PublishingRequestCase.java#L150 is called, this list is emptied, and `approvedFiles` is populated instead.

PR #1694, opened 2024-06-07
migratin related documents in original order

truhacevkir:


brage-import/src/test/java/no/sikt/nva/brage/migration/lambda/BrageEntryEventConsumerTest.java
```diff
@@ -1023,6 +1025,44 @@ void shouldMergePublicationContextAndUseCourseFromBragePublicationWhenPublicatio
...
+        var actualPublication = handler.handleRequest(s3Event, CONTEXT);
+
+        var actualPublicationContext = ((DegreePhd) actualPublication.publication().getEntityDescription()
+                                                     .getReference()
+                                                     .getPublicationInstance());
+
+        var related = actualPublicationContext.getRelated().stream().toList();
+        assertThat(related.getLast(), is(equalTo(new UnconfirmedDocument("2"))));
+        assertThat(related.get(related.size() - 2), is(equalTo(new UnconfirmedDocument("1"))));
+    }
```
truhacevkir, line 1064: Ugly assertion, but have no idea how to do other way when handling a set.

anetteOlli: Could hasItem plus assertion on the size of the list be better?
https://www.baeldung.com/hamcrest-core-matchers

truhacevkir: Tried it, but i want to assert that items from brage entry keep original order, but has item + size does not care about order.

anetteOlli (approved):

PR #1697, opened 2024-06-10
creating a file from dublin_core.xml

truhacevkir:
- Mapping dublin_core.xml as administrative agreement.
- All unknown files will be mapped to NullAssociatedArtifacts, we should actually not have files mapped to this file type

### DIFF
--- a/brage-import/src/main/java/no/sikt/nva/brage/migration/mapper/BrageNvaMapper.java
+++ b/brage-import/src/main/java/no/sikt/nva/brage/migration/mapper/BrageNvaMapper.java
@@ -47,6 +47,7 @@ import no.unit.nva.model.ResourceOwner;
 import no.unit.nva.model.Username;
 import no.unit.nva.model.associatedartifacts.AssociatedArtifact;
 import no.unit.nva.model.associatedartifacts.AssociatedLink;
+import no.unit.nva.model.associatedartifacts.NullAssociatedArtifact;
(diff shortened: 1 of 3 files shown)


brage-import/src/main/java/no/sikt/nva/brage/migration/mapper/BrageNvaMapper.java
```diff
@@ -217,9 +218,21 @@ private static List<AssociatedArtifact> convertFilesToAssociatedArtifact(List<Co
...
-                   : createAdministrativeAgreement(file, brageRecord);
+        return switch (file.getBundleType()) {
+            case BundleType.ORIGINAL -> createPublishedFile(file, brageRecord, embargoDate, legalNote);
+            case BundleType.LICENSE -> createAdministrativeAgreement(file, brageRecord);
+            case BundleType.IGNORED -> createAdministrativeAgreementForDublinCore(file, brageRecord);
+            default -> new NullAssociatedArtifact();
+        };
+    }
+
+    private static AssociatedArtifact createAdministrativeAgreementForDublinCore(ContentFile file, Record record) {
```
anetteOlli, line 229: I am a bit reluctant to store this metadata file as an admistrative agreement. I think this is an old version of the metadata, and should be stored as such. Administrative agreement is another thing. And storing the old version of the metadata as an associated artefact is also weird at best.

anetteOlli (approved):

PR #1699, opened 2024-06-10
merging more context types

truhacevkir:
This pull request will build once we have merged pr that bumps datamodel from version 24 to 25, cause method `randomPublicationNonDegree()` has been removed from datamodel-repo, and which is still in use here.


cristin-import/src/main/resources/possiblyEmptyFields.txt
```diff
@@ -30,6 +30,7 @@ entityDescription.reference.publicationContext.url
 entityDescription.reference.publicationContext.linkedContext
 entityDescription.reference.publicationContext.additionalIdentifiers
 entityDescription.reference.publicationContext.course
+entityDescription.reference.publicationContext.partOf
```
truhacevkir, line 33: Some test are failing since i added getter for partOf variable in MediaContribution class. I assume that this field can be empty.

anetteOlli (approved):

PR #1704, opened 2024-06-11
Merging other types in brage-migration

truhacevkir:


brinxmat (approved):

brage-import/src/test/java/no/sikt/nva/brage/migration/merger/PublicationInstanceMergerTest.java
```diff
@@ -68,7 +85,7 @@ void shouldUseExistingPublicationInstanceWhenNewPublicationInstanceIsEmpty(Publi
         var updatedPublication = mergePublications(existingPublication, bragePublication);
 
         assertThat(updatedPublication.getEntityDescription().getReference().getPublicationInstance(),
-                   doesNotHaveEmptyValues());
+                   doesNotHaveEmptyValuesIgnoringFields(Set.of("duration")));
```
anetteOlli, line 88: Add the field to ignoredFields.txt instead, that way we keep all our fields that are exempt for doesNotHaveEmptyValues in one place.

anetteOlli (changes requested):
Nice, bare pirk fra meg.

brage-import/src/test/java/no/sikt/nva/brage/migration/merger/PublicationInstanceMergerTest.java
truhacevkir: Have added, but it is a test, this test will still fail if I just do doesNotHaveEmptyValues(), cause random publication instance containing "duration" is always null.

anetteOlli (approved):

PR #1709, opened 2024-06-12
NP-46967-brage-merging-handle-multiple-duplicates

anetteOlli:


brage-import/src/main/java/no/sikt/nva/brage/migration/merger/findexistingpublication/HandleFinder.java
```diff
@@ -15,9 +16,12 @@
...
+    private final DuplicatePublicationReporter duplicatePublicationReporter;
 
-    public HandleFinder(ResourceService resourceService, UriRetriever uriRetriever, String apiHost) {
+    public HandleFinder(ResourceService resourceService, UriRetriever uriRetriever, String apiHost,
+                        DuplicatePublicationReporter duplicatePublicationReporter) {
         this.searchApiFinder = new SearchApiFinder(resourceService, uriRetriever, apiHost);
+        this.duplicatePublicationReporter = duplicatePublicationReporter;
     }
 
     @Override
```
truhacevkir, line 27: It seems like HandleFinder is the only finder that does not filter out publications as others finders do by this method/ class:

`.filter(item -> PublicationComparator.publicationsMatch(item, publicationRepresentation.publication()))`

Is there a special handling of publications fetched by handle? If no, it could also be nice to somehow move this method to the findExistingPublicationService, so we always "filter out" not relevant publications by default.

anetteOlli: The only reason there is a handle in additionalIdentifier is because a brage-migration has previously occurred.
So either we want to update all the values (probably), or discard the entire thing.

truhacevkir (approved):

brage-import/src/main/java/no/sikt/nva/brage/migration/merger/findexistingpublication/TitleAndTypePublicationFinder.java
```diff
@@ -80,14 +86,13 @@ private static Optional<String> getMainTitle(Publication publication) {
     private List<Publication> searchForPublicationsByTypeAndTitle(Publication publication) {
         var response = fetchResponse(searchByTypeAndTitleUri(publication));
         return response.map(this::toResponse)
-                   .filter(SearchResourceApiResponse::containsSingleHit)
```
anetteOlli, line 83: Denne må bort skal dette fungere i praksis.

brage-import/src/main/java/no/sikt/nva/brage/migration/merger/findexistingpublication/TitleAndTypePublicationFinder.java
```diff
@@ -80,14 +86,13 @@ private static Optional<String> getMainTitle(Publication publication) {
...
                    .map(SearchResourceApiResponse::hits)
                    .orElse(List.of())
                    .stream()
                    .map(ResourceWithId::getIdentifier)
                    .map(this::getPublicationByIdentifier)
                    .filter(item -> PublicationComparator.publicationsMatch(item, publication))
-                   .collect(Collectors.toList()).reversed();
```
anetteOlli, line 90: Dette virker merkelig.

PR #1711, opened 2024-06-13
deploying brage patch event consumer

truhacevkir:
Adding default constructor and template resource for BragePatchEventConsumer.

Doing it now since all the needed resources are now set up :)


joachimjorgensen (approved):
Looks good to me, but I'm no expert in template.yaml

anetteOlli (approved):

PR #1721, opened 2024-06-19
Merge projects from brage when empty

truhacevkir:


brage-import/src/main/java/no/sikt/nva/brage/migration/merger/publicationinstancemerger/PublicationInstanceMerger.java
```diff
@@ -40,7 +41,7 @@ public PublicationInstanceMerger(T publicationInstance) {
         this.publicationInstance = publicationInstance;
     }
 
-    public static PublicationInstanceMerger<?> of(PublicationInstance<?> instance) {
+    public static PublicationInstanceMerger<?> of(PublicationInstance<? extends Pages> instance) {
```
truhacevkir, line 44: Intellij does not like that we use PublicationInstance<?> cause publicationInstance extends always PublicationInstance<P extends Pages>

anetteOlli (approved):

PR #1726, opened 2024-06-25
Adding importDetail when updating existing publication by import entry

truhacevkir:


publication-commons/src/main/java/no/unit/nva/publication/service/impl/ResourceService.java
```diff
@@ -179,6 +179,13 @@ public Publication createPublicationFromImportedEntry(Publication inputData, Imp
         return insertResource(newResource);
     }
 
+    public Publication updatePublicationByImportEntry(Publication publication, ImportSource importSource) {
+        if (nonNull(importSource)) {
+            publication.addImportDetail(new ImportDetail(Instant.now(), importSource));
+        }
+        return updateResourceService.updatePublicationButDoNotChangeStatus(publication);
```
truhacevkir, line 186: Do we want the same timestamp as in modifiedDate?

cristin-import/src/main/java/no/unit/nva/cristin/lambda/CristinEntryEventConsumer.java
```diff
@@ -181,7 +181,8 @@ private Publication update(PublicationRepresentations publicationRepresentations
 
     private PublicationRepresentations persistUpdatedPublication(
         PublicationRepresentations publicationRepresentations) {
-        resourceService.updatePublication(publicationRepresentations.getExistingPublication());
+        resourceService.updatePublicationByImportEntry(publicationRepresentations.getExistingPublication(),
+                                                       ImportSource.fromSource(Source.CRISTIN));
```
truhacevkir, line 185: Adding importDetail in cristin-import

brage-import/src/main/java/no/sikt/nva/brage/migration/lambda/BrageEntryEventConsumer.java
```diff
@@ -265,8 +265,11 @@ private UriWrapper discardedFilesReportUri(Publication publication, S3Event s3Ev
     }
 
     private BrageMergingReport persistInDatabaseAndCreateMergeReport(Publication publicationForUpdate,
-                                                                     Publication existinPublication) {
-        var newImage = resourceService.updatePublication(publicationForUpdate);
+                                                                     Publication existinPublication,
+                                                                     PublicationRepresentation publicationRepresentation) {
+        var customerName = publicationRepresentation.brageRecord().getCustomer().getName();
+        var importSource = ImportSource.fromBrageArchive(customerName);
+        var newImage = resourceService.updatePublicationByImportEntry(publicationForUpdate, importSource);
```
truhacevkir, line 272: adding import detail in brage-import

publication-commons/src/main/java/no/unit/nva/publication/service/impl/ResourceService.java
joachimjorgensen: I guess that would be best, if it's not too much hassle

publication-commons/src/main/java/no/unit/nva/publication/service/impl/ResourceService.java
```diff
@@ -179,6 +179,13 @@ public Publication createPublicationFromImportedEntry(Publication inputData, Imp
         return insertResource(newResource);
     }
 
+    public Publication updatePublicationByImportEntry(Publication publication, ImportSource importSource) {
+        if (nonNull(importSource)) {
+            publication.addImportDetail(new ImportDetail(Instant.now(), importSource));
+        }
```
joachimjorgensen, line 185: Throw exception if null instead? Doesn't really make sense to use this method if `importSource` is null

publication-commons/src/test/java/no/unit/nva/publication/service/impl/ResourceServiceTest.java
```diff
@@ -1187,6 +1187,18 @@ void shouldThrowBadRequestWhenAttemptingToDeletePublishedPublication() throws Ap
...
     }
 
+    @Test
+    void shouldUpdatePublicationByImportedEntry() throws ApiGatewayException {
+        var publication = createPublishedResource();
+        publication.setDoi(randomDoi());
+        var updatedPublication = resourceService.updatePublicationByImportEntry(publication,
+                                                                   ImportSource.fromSource(Source.CRISTIN));
+
+        assertThat(updatedPublication.getImportDetails().getFirst().importSource().getSource(),
```
joachimjorgensen, line 1197: Add an assertion that `getImportDetails().size == 1`? To make sure we're not call the method too many times

joachimjorgensen: Or maybe it's better to assert that in the handler tests

joachimjorgensen (approved):

publication-commons/src/test/java/no/unit/nva/publication/service/impl/ResourceServiceTest.java
truhacevkir: Will assert in both of tests

publication-commons/src/main/java/no/unit/nva/publication/service/impl/ResourceService.java
truhacevkir: throwing IllegalArgumentException when passing null to this method

joachimjorgensen (approved):

PR #1727, opened 2024-06-26
NP-46825 Set logged in user as finlizedBy for autocompleted tickets

joachimjorgensen:
Earlier, when tickets were autocompleted, the publication owner was set as `finalizedBy`. Now, the logged in user is set instead.

### DIFF
--- a/publication-commons/src/main/java/no/unit/nva/publication/model/business/PublishingRequestCase.java
+++ b/publication-commons/src/main/java/no/unit/nva/publication/model/business/PublishingRequestCase.java
@@ -353,9 +353,10 @@ public class PublishingRequestCase extends TicketEntry {
                && Objects.equals(getApprovedFiles(), that.getApprovedFiles());
     }
 
-    public PublishingRequestCase persistAutoComplete(TicketService ticketService, Publication publication)
+    public PublishingRequestCase persistAutoComplete(TicketService ticketService, Publication publication,
+                                                     Username finalizedBy)
         throws ApiGatewayException {
-        var completed = this.complete(publication, publication.getResourceOwner().getOwner());
+        var completed = this.complete(publication, finalizedBy);
         return ticketService.createTicket(completed);
     }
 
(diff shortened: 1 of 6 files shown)


publication-commons/src/main/java/no/unit/nva/publication/model/business/PublishingRequestCase.java
```diff
@@ -353,9 +353,10 @@ && getStatus() == that.getStatus()
                && Objects.equals(getApprovedFiles(), that.getApprovedFiles());
     }
 
-    public PublishingRequestCase persistAutoComplete(TicketService ticketService, Publication publication)
+    public PublishingRequestCase persistAutoComplete(TicketService ticketService, Publication publication,
+                                                     Username finalizedBy)
         throws ApiGatewayException {
-        var completed = this.complete(publication, publication.getResourceOwner().getOwner());
+        var completed = this.complete(publication, finalizedBy);
```
joachimjorgensen, line 359: This is the fix

tickets/src/test/java/no/unit/nva/publication/ticket/create/CreateTicketHandlerTest.java
```diff
@@ -657,6 +658,26 @@ void userWithAccessRightManageDoiShouldNotBeAbleToAutoPublishFilesWhenAllowedToP
...
+        var request = createHttpTicketCreationRequestWithApprovedAccessRight(requestBody, publication, curatorName,
+                                                                             publicationPublisher,
+                                                                             MANAGE_PUBLISHING_REQUESTS, MANAGE_DOI);
+        handler.handleRequest(request, output, CONTEXT);
+        var response = GatewayResponse.fromOutputStream(output, Void.class);
+        assertThat(response.getStatusCode(), is(equalTo(HTTP_CREATED)));
+
+        var ticket = fetchTicket(response);
+
+        assertThat(ticket.getFinalizedBy().toString(), is(not(equalTo(publication.getResourceOwner().getOwner()))));
```
joachimjorgensen, line 677: Not really necessary, but added to reiterate what we do not want. Should remove?

truhacevkir (approved):
Looks good, since you have not modified a singe test, it seems like this has not been tested before. Good catch!

PR #1729, opened 2024-06-26
NP-47047-brage-merging-prioritize-publisher-for-degree

anetteOlli:


brage-import/src/main/java/no/sikt/nva/brage/migration/merger/publicationinstancemerger/PublicationInstanceMerger.java
```diff
@@ -69,7 +69,7 @@ public static PublicationInstanceMerger<?> of(PublicationInstance<? extends Page
         };
     }
 
-    public abstract T merge(PublicationInstance<?> newInstance);
+    public abstract PublicationInstance<?> merge(PublicationInstance<?> newInstance);
```
anetteOlli, line 72: Got a bunch of random builds fails because CAP# 1 ett eller annet med den ubrukte Typen der.

truhacevkir (approved):

PR #1730, opened 2024-06-27
removing reserved concurrency

truhacevkir:
@anetteOlli, @brinxmat, @torbjokv, @torarnet have to evaluate if these lambda's need reserved concurrency. 

Removed reserved concurrency from:

 - NvaScopusFunction -> api calls to customer api, Cristin and Pia, bad decision?
 - ScopusDeletionEventEmitter -> probably no need to reserve concurrency
 - DeleteImportCandidateEventConsumer -> probably no need to reserve concurrency, it simply deletes entry from dynamo
 - ResourceExpansionHandler -> api calls to ChannelRegister + Cristin, bad decision?
 - ImportCandidateExpansionHandler -> api calls to CustomerApi, ChannelRegister + Cristin, bad decision?
 - ExpandedResourcesPersistenceHandler -> probably no need to reserve concurrency, it just persists document to s3
 - AnalyticsIntegrationHandler -> have no idea if this is in use, do we need to reserve concurrency for it?
 - AcceptedPublishingRequestHandler -> probably no need to reserve concurrency
 - CristinEntriesEventEmitter -> probably no need to reserve concurrency (no api calls to other services)
 - ScopusUnzipperFunction -> probably no need to reserve concurrency

### DIFF
--- a/template.yaml
+++ b/template.yaml
@@ -1019,7 +1019,6 @@ Resources:
       CodeUri: scopus-import
       Handler: no.sikt.nva.scopus.ScopusHandler::handleRequest
       Role: !GetAtt LambdaRole.Arn
-      ReservedConcurrentExecutions: 50
       Timeout: 900
       Environment:
         Variables:
@@ -1051,7 +1050,6 @@ Resources:
       CodeUri: s3-import-commons
       Handler: no.unit.nva.publication.s3imports.DeleteImportCandidatesEventEmitter::handleRequest
       Role: !GetAtt LambdaRole.Arn
-      ReservedConcurrentExecutions: 50
       Timeout: 900
       Environment:
         Variables:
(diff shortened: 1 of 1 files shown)


template.yaml
```diff
@@ -1019,7 +1019,6 @@ Resources:
       CodeUri: scopus-import
       Handler: no.sikt.nva.scopus.ScopusHandler::handleRequest
       Role: !GetAtt LambdaRole.Arn
-      ReservedConcurrentExecutions: 50
```
anetteOlli, line 1022: Remember that scopes-import will always run in 4 environment in parallell. All 4 environments will receive the same email at the same time. This might stress the channel-registry.

anetteOlli (approved):

torbjokv:
I fear that removing some of these may increase the throttle issue we have when batch-scanning and it will deplete the unreserved concurrency pool even more than before (maybe?). I agree that we probably want to remove some of these concurrency restrictions, but I think we need to look into how batch scan is done and allow us to put on brakes there before removing restrictions completely. When in full production, we are probably very dependent on being able run batch scan without impacting the stability since it takes quite a bit of time.
image

brinxmat (approved):

PR #1733, opened 2024-06-28
NP-47049-brage-merging-overwrite-certain-metadata-fields

anetteOlli:


brage-import/src/main/java/no/sikt/nva/brage/migration/merger/CristinImportPublicationMerger.java
```diff
@@ -94,29 +101,91 @@ private Publication mergePublicationsMetadata()
...
+                   .withAlternativeAbstracts(determineAlternativeAbstracts())
+                   .withTags(determineTags())
                    .build();
     }
 
+    private List<String> determineTags() {
+        return shouldPrioritizeField(PRIORITIZE_TAGS)
+                   ? mergeTags()
+                   : existingPublication.getEntityDescription().getTags();
+    }
```
truhacevkir, line 137: Does it mean that we have to set prioritized values here in order to be able to merge them with existing tags?
Do we want that?

truhacevkir: It looks like it as a field we have not merged before, so it is probably fine :) Im just thinking that tags is type of filed that can be merged no matter what, we just have to filter duplicates

PR #1735, opened 2024-07-01
Remove checkstyle issues

brinxmat:
- Removes checkstyle issues that have accumulated
- Tempted to fail builds that have checkstyle issues in future, we should probably configure this

### DIFF
--- a/scopus-import/src/test/java/no/sikt/nva/scopus/utils/ScopusGenerator.java
+++ b/scopus-import/src/test/java/no/sikt/nva/scopus/utils/ScopusGenerator.java
@@ -82,6 +82,7 @@ import no.unit.nva.language.Language;
 import no.unit.nva.language.LanguageConstants;
 import nva.commons.core.ioutils.IoUtils;
 import nva.commons.core.paths.UriWrapper;
+import org.jetbrains.annotations.NotNull;
 
 public final class ScopusGenerator {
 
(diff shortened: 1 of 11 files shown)


torbjokv (approved):

scopus-import/src/test/java/no/sikt/nva/scopus/utils/ScopusGenerator.java
```diff
@@ -279,6 +280,21 @@ private HeadTp randomHeadTpWithContributorsFromAuthorTp(AuthorTp authorTp) {
...
+    private static @NotNull AffiliationTp randomAffiliationTp() {
+        var organization = randomOrganizationTp();
+        var affiliationTp = new AffiliationTp();
+        affiliationTp.setCountry(randomString());
+        affiliationTp.setAfid(randomString());
+        affiliationTp.getOrganization().add(organization);
+        return affiliationTp;
+    }
+
+    private static @NotNull OrganizationTp randomOrganizationTp() {
```
truhacevkir, line 292: Do we need this Intellij annotation?

truhacevkir (approved):

scopus-import/src/test/java/no/sikt/nva/scopus/utils/ScopusGenerator.java
brinxmat: Possibly not, it was likely auto-format stuff. There are multiple issues in this file, which I have not addressed

PR #1738, opened 2024-07-02
setting finalizedBy to requestee username

truhacevkir:
When updating publication and approving existing pubReq, finalizedBy value should be set to the user who performs update, not publication owner.

### DIFF
--- a/publication-rest/src/main/java/no/unit/nva/publication/update/UpdatePublicationHandler.java
+++ b/publication-rest/src/main/java/no/unit/nva/publication/update/UpdatePublicationHandler.java
@@ -169,8 +169,7 @@ public class UpdatePublicationHandler
         var permissionStrategy = PublicationPermissionStrategy.create(existingPublication, userInstance);
         Publication updatedPublication = switch (input) {
             case UpdatePublicationRequest publicationMetadata ->
-                updateMetadata(publicationMetadata, identifierInPath, existingPublication, permissionStrategy,
-                               requestInfo, userInstance);
+                updateMetadata(publicationMetadata, identifierInPath, existingPublication, permissionStrategy, userInstance);
 
             case UnpublishPublicationRequest unpublishPublicationRequest ->
                 unpublishPublication(unpublishPublicationRequest,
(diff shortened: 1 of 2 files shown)


publication-rest/src/main/java/no/unit/nva/publication/update/UpdatePublicationHandler.java
```diff
@@ -465,31 +469,19 @@ private Stream<PublishingRequestCase> fetchPendingPublishingRequest(Publication
...
-            default -> {
-                return HttpStatus.SC_BAD_REQUEST;
-            }
-        }
+        return switch (input) {
+            case UpdatePublicationRequest ignored -> HttpStatus.SC_OK;
+            case UnpublishPublicationRequest ignored -> HttpStatus.SC_ACCEPTED;
+            case DeletePublicationRequest ignored -> HttpStatus.SC_ACCEPTED;
+            default -> HttpStatus.SC_BAD_REQUEST;
+        };
```
truhacevkir, line 484: Refactoring

publication-rest/src/main/java/no/unit/nva/publication/update/UpdatePublicationHandler.java
```diff
@@ -452,10 +452,14 @@ private void validatePublication(Publication publicationUpdate, Customer custome
         }
     }
 
-    private void autoCompletePendingPublishingRequestsIfNeeded(Publication publication) {
+    private void autoCompletePendingPublishingRequestsIfNeeded(Publication publication, RequestInfo requestInfo) {
         fetchPendingPublishingRequest(publication)
             .map(PublishingRequestCase.class::cast)
-            .forEach(ticket -> ticket.complete(publication, getOwner(publication)).persistUpdate(ticketService));
+            .forEach(ticket -> ticket.complete(publication, getUserName(requestInfo)).persistUpdate(ticketService));
```
sondrev, line 458: getUserName() actually polls conginto each time you call it. Instead do it once outside the forEach

sondrev (changes requested):

publication-rest/src/main/java/no/unit/nva/publication/update/UpdatePublicationHandler.java
truhacevkir: Sending userInstance instead, thank you :)

brinxmat (approved):

PR #1740, opened 2024-07-03
validating curator access right on ticket creation

truhacevkir:
Validation curator access rights when creating new ticket. 
Not sure about what access right curator should have to be able to create UnpublishRequest, so I have set `manage_publishing_request` access right. Without validating access right for UnpublishRequest creation anyone from the same institution will be able to persist UnpublishRequest ticket.

### DIFF
--- a/tickets/src/main/java/no/unit/nva/publication/ticket/create/CreateTicketHandler.java
+++ b/tickets/src/main/java/no/unit/nva/publication/ticket/create/CreateTicketHandler.java
@@ -3,6 +3,9 @@ package no.unit.nva.publication.ticket.create;
 import static no.unit.nva.publication.PublicationServiceConfig.API_HOST;
 import static no.unit.nva.publication.PublicationServiceConfig.ENVIRONMENT;
 import static no.unit.nva.publication.PublicationServiceConfig.PUBLICATION_PATH;
+import static nva.commons.apigateway.AccessRight.MANAGE_DOI;
+import static nva.commons.apigateway.AccessRight.MANAGE_PUBLISHING_REQUESTS;
+import static nva.commons.apigateway.AccessRight.SUPPORT;
 import static nva.commons.core.attempt.Try.attempt;
 import com.amazonaws.services.lambda.runtime.Context;
 import java.net.HttpURLConnection;
@@ -17,17 +20,18 @@ import no.unit.nva.publication.PublicationServiceConfig;
 import no.unit.nva.publication.external.services.AuthorizedBackendUriRetriever;
 import no.unit.nva.publication.external.services.UriRetriever;
 import no.unit.nva.publication.model.business.TicketEntry;
-import no.unit.nva.publication.model.business.UserInstance;
 import no.unit.nva.publication.service.impl.ResourceService;
 import no.unit.nva.publication.service.impl.TicketService;
+import no.unit.nva.publication.ticket.DoiRequestDto;
+import no.unit.nva.publication.ticket.GeneralSupportRequestDto;
+import no.unit.nva.publication.ticket.PublishingRequestDto;
 import no.unit.nva.publication.ticket.TicketDto;
+import no.unit.nva.publication.ticket.UnpublishRequestDto;
 import no.unit.nva.publication.utils.RequestUtils;
-import nva.commons.apigateway.AccessRight;
 import nva.commons.apigateway.ApiGatewayHandler;
 import nva.commons.apigateway.RequestInfo;
 import nva.commons.apigateway.exceptions.ApiGatewayException;
 import nva.commons.apigateway.exceptions.ForbiddenException;
-import nva.commons.apigateway.exceptions.UnauthorizedException;
 import nva.commons.core.JacocoGenerated;
 import nva.commons.core.paths.UriWrapper;
 import org.slf4j.Logger;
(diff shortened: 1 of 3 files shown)


tickets/src/main/java/no/unit/nva/publication/ticket/create/CreateTicketHandler.java
```diff
@@ -99,33 +99,38 @@ private static URI getPublisherId(Publication publication) {
...
 
-    private static boolean hasValidAccessRights(RequestInfo requestInfo) {
-        return requestInfo.userIsAuthorized(AccessRight.MANAGE_DOI);
+    private static boolean hasValidAccessRights(RequestUtils requestUtils, TicketDto ticketDto) {
+        return switch (ticketDto) {
+            case DoiRequestDto doi -> requestUtils.hasAccessRight(MANAGE_DOI);
+            case PublishingRequestDto publishing -> requestUtils.hasAccessRight(MANAGE_PUBLISHING_REQUESTS);
+            case GeneralSupportRequestDto support -> requestUtils.hasAccessRight(SUPPORT);
+            case UnpublishRequestDto unpublish -> requestUtils.hasAccessRight(MANAGE_PUBLISHING_REQUESTS);
+            case null, default -> false;
```
truhacevkir, line 108: Not sure about following:
`case UnpublishRequestDto unpublish -> requestUtils.hasAccessRight(MANAGE_PUBLISHING_REQUESTS);`

But without passing any access right here by simply returning true will allow anyone to create UnpublishRequest.

brinxmat: I think it makes sense, but perhaps the naming is wrong as MANAGE_PUBLISHING_REQUESTS is strongly connected to a particular ticket type. Should probably be MANAGE_PUBLISHING

tickets/src/test/java/no/unit/nva/publication/ticket/create/CreateTicketHandlerTest.java
```diff
@@ -678,6 +678,39 @@ void shouldSetFinalizedByFromRequestUtilsWhenTicketIsAutoApproved()
         assertThat(ticket.getFinalizedBy().toString(), is(equalTo(curatorName)));
     }
 
+    @ParameterizedTest
+    @MethodSource("no.unit.nva.publication.ticket.test.TicketTestUtils#invalidAccessRightForTicketTypeProvider")
+    void shouldNotAllowCuratorWithoutValidAccessRightToCreateTicket(Class<? extends TicketEntry> ticketType,
```
brinxmat, line 683: What is a curator if they don't have the correct access right?

tickets/src/test/java/no/unit/nva/publication/ticket/create/CreateTicketHandlerTest.java
```diff
@@ -678,6 +678,39 @@ void shouldSetFinalizedByFromRequestUtilsWhenTicketIsAutoApproved()
...
+            requestBody, publication, user.getUsername(), user.getCustomerId(), accessRight);
+        handler.handleRequest(input, output, CONTEXT);
+
+        var response = GatewayResponse.fromOutputStream(output, Void.class);
+        assertThat(response.getStatusCode(), is(equalTo(HTTP_FORBIDDEN)));
+    }
+
+    @ParameterizedTest
+    @MethodSource("no.unit.nva.publication.ticket.test.TicketTestUtils#ticketTypeAndAccessRightProvider")
+    void shouldAllowCuratorWithValidAccessRightToCreateTicket(PublicationStatus status,
```
brinxmat, line 699: Similarly

tickets/src/test/java/no/unit/nva/publication/ticket/create/CreateTicketHandlerTest.java
truhacevkir: Curator with accessRight not related to ticket they attempt to create, see:


```
    public static Stream<Arguments> invalidAccessRightForTicketTypeProvider() {
        return Stream.of(Arguments.of(DoiRequest.class, AccessRight.MANAGE_PUBLISHING_REQUESTS),
                         Arguments.of(PublishingRequestCase.class, AccessRight.MANAGE_DOI));
                         Arguments.of(GeneralSupportRequest.class, AccessRight.MANAGE_PUBLISHING_REQUESTS));
    }
```

truhacevkir: It may make more sense if i change Curator to User in test name?

brinxmat (approved):

tickets/src/test/java/no/unit/nva/publication/ticket/create/CreateTicketHandlerTest.java
brinxmat: I think that was what I was getting at. I find the "X Curator" concepts a bit difficult since they aren't codified

tickets/src/main/java/no/unit/nva/publication/ticket/create/CreateTicketHandler.java
torbjokv: I think contributors and resource owner should be able to create a unpublish request if there are no files, maybe api clients should be allowed to do this (it uses different scopes) and when curator does this, we also need to check resource owner and contributors.  something that maybe should be a part of the logic here?

anetteOlli (approved):

tickets/src/main/java/no/unit/nva/publication/ticket/create/CreateTicketHandler.java
torbjokv: my bad, unpublish is through the update publication endpoint, not here, so not relevant. The other curator stuff still may be to loosely checked here, but this change is for the better anyway

torbjokv (approved):

PR #1745, opened 2024-07-04
setting back reserver concurrency for import lambdas

truhacevkir:
Importer med reserved concurrency satt til 50 ser ut til å gå fint.


anetteOlli (approved):
Are some of these operating by sqs queues? Is it possible to use your suggestion of max concurrency?

truhacevkir:
> Are some of these operating by sqs queues? Is it possible to use your suggestion of max concurrency?

Adding following scalingConfig to import lambdas that that read from sqs :

```
            ScalingConfig:
              MaximumConcurrency: 50
```

PR #1748, opened 2024-07-04
introduced input validation in api gateway function handler

ketilaa:
Only locally in `nva-publication-api` yet. And only on `CreatePublicationHandler`. And only on `rightsHolder` field in `CreatePublicationRequest`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -9,10 +9,10 @@ allprojects{
 }
 
 tasks.withType(JavaCompile) {
-    sourceCompatibility = JavaVersion.VERSION_17
+    sourceCompatibility = JavaVersion.VERSION_21
 }
 
 wrapper {
-    gradleVersion = '8.2.1'
+    gradleVersion = '8.5'
     distributionType = Wrapper.DistributionType.ALL
 }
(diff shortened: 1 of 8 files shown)


build.gradle
```diff
@@ -9,10 +9,10 @@ allprojects{
 }
 
 tasks.withType(JavaCompile) {
-    sourceCompatibility = JavaVersion.VERSION_17
+    sourceCompatibility = JavaVersion.VERSION_21
```
ketilaa, line 12: Had some issues getting to project to behave in IntelliJ related to java 21. This was left behind in java 17.

publication-rest/src/main/java/no/unit/nva/publication/ValidatingApiGatewayHandler.java
ketilaa, line 1: Will be moved to `nva-commons` when we know if we want a separate 'ValidatingApiGatewayHandler' or just to extend 'ApiGatewayHandler'.

publication-rest/src/main/java/no/unit/nva/publication/ValidatingApiGatewayHandler.java
```diff
@@ -0,0 +1,51 @@
...
+        if (Objects.nonNull(input)) {
+            validateInput(input);
+        }
+        return processValidatedInput(input, requestInfo, context);
+    }
+
+    private void validateInput(I input) throws UnprocessableContentException {
+        var result = validator.validate(input);
+        if (!result.isEmpty()) {
+            throw new UnprocessableContentException(toMessage(result));
```
ketilaa, line 37: 422 Unprocessable Content ok, or do we prefer 400 Bad Request when input validation fails?

truhacevkir (approved):

torbjokv (approved):

PR #1749, opened 2024-07-04
Fix checkstyle errors — a few remain

brinxmat:
- Fix all checkstyle errors that don't involve
   - Cucumber method names
   - Very long test names
- Maybe worth fixing checkstyle errors in long test names by using `@DisplayName` 
- Unsure if we can fix the Cucumber method names without affecting the actual tests
- We should probably force the build to fail on checkstyle errors to avoid this in future


truhacevkir (approved):
Nice!

ketilaa (approved):

PR #1753, opened 2024-07-08
Expansion, frame hasPart

monaullah:
Fix cases where `hasPart` only contains one object, it is not framed as an `ArrayNode`


truhacevkir (approved):

ketilaa (approved):
Good catch :)

PR #1755, opened 2024-07-08
moving customer config to publication-api - brage-migration

truhacevkir:
- Moving customers config to publication-api so we do not need to run multiple imports for different environments. 
- These "new" way of setting customer and resourceOwner should be compatible with "old" records generated by brage-migration.jar


anetteOlli (approved):
Bra jobba

PR #1756, opened 2024-07-09
absorbed data model

ketilaa:
Absorbed the data model into publication-api. This will make it easier to manage dependencies, and it will be easier to work on input validation that requires annotating the data model.

Tip: filter out *.json and *.java and focus on the gradle stuff


truhacevkir (approved):
Nice, I would deploy it in sandbox and check if everything works as expected :)

anetteOlli (approved):
Bra jobba!

Compiles locally for me.

Checked that I could not find datamodel in .toml.
Checked that .ttl files were moved here.

Also do we really need .json documentation? Does anybody use those?

Recommend deploying this branch to nva-sandbox before merging.

PR #1757, opened 2024-07-09
filtering out null tags

truhacevkir:
Will do the same when crafting tags in brage-migration


anetteOlli (approved):

brage-import/src/main/java/no/sikt/nva/brage/migration/mapper/BrageNvaMapper.java
```diff
@@ -365,7 +365,7 @@ private static URI extractDoi(Record brageRecord) {
     }
 
     private static List<String> extractTags(Record brageRecord) {
-        return brageRecord.getEntityDescription().getTags();
+        return brageRecord.getEntityDescription().getTags().stream().filter(Objects::nonNull).toList();
```
anetteOlli, line 368: User stringUtils isNotBlank

anetteOlli (changes requested):

PR #1767, opened 2024-08-06
Np 47254 fixing timeout when uploading files

truhacevkir:
When lambda times out, it does not persist error report. It makes it hard to see if posts with large files have been imported or not. When lambda almost reaches timeout threshold when moving large s3 objects, throwing exception and persisting report.

### DIFF
--- a/brage-import/src/main/java/no/sikt/nva/brage/migration/merger/S3MultipartCopier.java
+++ b/brage-import/src/main/java/no/sikt/nva/brage/migration/merger/S3MultipartCopier.java
@@ -1,5 +1,6 @@
 package no.sikt.nva.brage.migration.merger;
 
+import com.amazonaws.services.lambda.runtime.Context;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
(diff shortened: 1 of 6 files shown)


brage-import/src/main/java/no/sikt/nva/brage/migration/merger/S3MultipartCopier.java
```diff
@@ -121,14 +125,17 @@ private void performSimpleCopy(S3Client s3Client) {
...
-    private void performMultiPartCopy(S3Client s3Client, HeadObjectResponse headOfObjectToCopy) {
+    private void performMultiPartCopy(S3Client s3Client, Context context, HeadObjectResponse headOfObjectToCopy) {
         var request = initiateMultiUploadRequest(headOfObjectToCopy);
         var response = s3Client.createMultipartUpload(request);
         try {
             long position = 0;
             int partNumber = 1;
             long totalSize = headOfObjectToCopy.contentLength();
             while (position < totalSize) {
+                if (context.getRemainingTimeInMillis() < TIMEOUT_THRESHOLD_30_SECONDS) {
```
torbjokv, line 136: what's happening here, when getRemainingTimeInMillis() is less then 30 000ms, then throw timeout

truhacevkir: Yes, if there is less then 30 seconds left to proceed the record, we throw exception and persist error report. Before exception was thrown without persisting report, which made it difficult to find missing record.

torbjokv: ok :-) sounds good then 👍

torbjokv (approved):

PR #1772, opened 2024-08-09
Np 46995 read orcid from xml when not present in cristin

truhacevkir:
When creating contributor from fetched person from Cristin under scopus-import
And person in Cristin is missing OrcId
And OrcId is present in xml from Scopus
Then use OrcId from xml

### DIFF
--- a/scopus-import/src/test/java/no/sikt/nva/scopus/ScopusHandlerTest.java
+++ b/scopus-import/src/test/java/no/sikt/nva/scopus/ScopusHandlerTest.java
@@ -997,8 +997,8 @@ class ScopusHandlerTest extends ResourcesLocalTest {
                    startsWith(correspondingAuthorTp.getGivenName()));
     }
 
-   @ParameterizedTest(name = "Should have entityDescription with language:{1}")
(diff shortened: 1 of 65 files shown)


scopus-import/src/test/java/no/sikt/nva/scopus/ScopusHandlerTest.java
```diff
@@ -997,8 +997,8 @@ void shouldExtractCorrespondingAuthor() throws IOException {
                    startsWith(correspondingAuthorTp.getGivenName()));
     }
 
-   @ParameterizedTest(name = "Should have entityDescription with language:{1}")
-   @MethodSource("providedLanguagesAndExpectedOutput")
+    @ParameterizedTest(name = "Should have entityDescription with language:{1}")
+    @MethodSource("providedLanguagesAndExpectedOutput")
```
truhacevkir, line 1001: Fixing checkstyle

torbjokv (approved):
would tests shouldExtractOrcIdFromCristinPersonIfPresent and shouldReturnNullOrchidIfNotPresent make sense? otherwise it looks good to me!

PR #1773, opened 2024-08-09
creating contributor with active affiliations only when present - scopus-import

truhacevkir:
When fetched Cristin person contains active affiliations
And fetched Cristin organizations are present
Then create contributor with active affiliations from Cristin person response only.

Fetched organizations are often the same or are top level orgs of person active affiliations, in that way we do not create contributor with multiple affiliations pointing at the same organization. 
For example `UiO`(from fetched organisation) and `UiO, Fakultetet for økonomi` (from fetched person affiliations). 
From now on we keep only `UiO, Fakultetet for økonomi` when both are present.

### DIFF
--- a/scopus-import/src/main/java/no/sikt/nva/scopus/conversion/CristinContributorExtractor.java
+++ b/scopus-import/src/main/java/no/sikt/nva/scopus/conversion/CristinContributorExtractor.java
@@ -4,7 +4,6 @@ import static java.util.Objects.nonNull;
 import static no.sikt.nva.scopus.conversion.ContributorExtractor.extractAdditionalIdentifiers;
 import java.util.Collection;
 import java.util.List;
-import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Stream;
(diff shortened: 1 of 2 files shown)


LarsV123 (approved):

scopus-import/src/main/java/no/sikt/nva/scopus/conversion/CristinContributorExtractor.java
```diff
@@ -85,18 +84,18 @@ private static ContributorVerificationStatus generateVerificationStatus(CristinP
 
     private static List<Corporation> generateOrganizations(Set<Affiliation> affiliations,
                                                            List<CristinOrganization> cristinOrganizations) {
-        var organizations = createOrganizationsFromCristinPersonAffiliations(affiliations).toList();
-        var organisationFromAuthorGroupTp = createOrganizationFromCristinOrganization(cristinOrganizations).toList();
-        return Stream.concat(organizations.stream(), organisationFromAuthorGroupTp.stream())
-                   .filter(Objects::nonNull)
-                   .toList();
+        var cristinPersonActiveAffiliations = createOrganizationsFromActiveCristinPersonAffiliations(affiliations);
+        var organisationsFromAuthorGroupTp = createOrganizationFromCristinOrganization(cristinOrganizations).toList();
```
monaullah, line 88: What is `Tp` in `organisationsFromAuthorGroupTp`? Spell out?

monaullah (approved):

PR #1776, opened 2024-08-13
create period for event when one of period dates is missing

truhacevkir:
When creating Event from Cristin publication and one of period dates is missing - create period with one date.

### DIFF
--- a/cristin-import/src/main/java/no/unit/nva/cristin/mapper/nva/ReferenceBuilder.java
+++ b/cristin-import/src/main/java/no/unit/nva/cristin/mapper/nva/ReferenceBuilder.java
@@ -194,7 +194,7 @@ public class ReferenceBuilder extends CristinMappingModule {
     private Time extractTime() {
         var from = extractCristinEventFromDate();
         var to = extractCristinEventToDate();
-        return nonNull(from) && nonNull(to) ? new Period(from, to) : null;
+        return nonNull(from) || nonNull(to) ? new Period(from, to) : null;
     }
 
     private Instant extractCristinEventFromDate() {
--- a/cristin-import/src/test/java/cucumber/EventFeatures.java
+++ b/cristin-import/src/test/java/cucumber/EventFeatures.java
@@ -105,8 +105,8 @@ public class EventFeatures {
         assertThat(((Period) event.getTime()).getFrom().toString(), is(equalTo(value)));
     }
 
-    @Then("the Event has no a time Period with toDate")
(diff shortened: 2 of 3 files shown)


cristin-import/src/main/java/no/unit/nva/cristin/mapper/nva/ReferenceBuilder.java
```diff
@@ -194,7 +194,7 @@ private PublicationContext buildEventForPublicationContext() {
     private Time extractTime() {
         var from = extractCristinEventFromDate();
         var to = extractCristinEventToDate();
-        return nonNull(from) && nonNull(to) ? new Period(from, to) : null;
+        return nonNull(from) || nonNull(to) ? new Period(from, to) : null;
```
LarsV123, line 197: What is the expected behavior when the "from" date is missing? Does that make sense or would that be an error in the data?

cristin-import/src/test/java/cucumber/EventFeatures.java
```diff
@@ -104,6 +105,15 @@ public void theNvaEventHasAPeriodWithFromDate(String value) {
         assertThat(((Period) event.getTime()).getFrom().toString(), is(equalTo(value)));
     }
 
+    @Then("the Event has no a time Period with toDate")
```
LarsV123, line 108: Suggestion: Rename this to "the Event toDate is null" or "the Event end date is missing" perhaps? Equivalent for the method name could be "theNvaEventPeriodToDateEquals".

LarsV123 (commented):
I'm not sure it makes sense for an event to exist with only a "to" date, but that's what the ticket says :)

cristin-import/src/main/java/no/unit/nva/cristin/mapper/nva/ReferenceBuilder.java
truhacevkir: One of dates can be missing, we are migrating data from Cristin and we have to import dates/periods as is. It will not be an error in the data.

truhacevkir: It can be the case where user has started to create draft publication, but is not sure when event actually starts, and leaves this field empty.

LarsV123: Aha, I see!

LarsV123 (approved):

PR #1780, opened 2024-08-14
not injecting import details when importing the same cristin publication twice

truhacevkir:


truhacevkir:
Purpose of this is not to show imported from Cristin multiple times when running same import over and over agai:
image

LarsV123 (approved):
This seems to work as described, but can you explain why it is necessary to discard the data? If I understand correctly, the "importDetails" field is now an array that will only hold one entry (the original import event). Is there some other way to log/track when a document is updated via import?

It seems odd to not preserve the history of events, just to avoid a frontend issue. I'm probably misunderstanding something here though.

monaullah:
Will this case be relevant in the production environment? If yes, are we sure we don't want to keep the import history for a Publication? What happens to a Publication when the cristin import is run several times? Can it be updated by the second import?

truhacevkir:
> Reference in n

It can, yes, in the case we found out that we forgot to migrate som fields.

truhacevkir:
> This seems to work as described, but can you explain why it is necessary to discard the data? If I understand correctly, the "importDetails" field is now an array that will only hold one entry (the original import event). Is there some other way to log/track when a document is updated via import?
> 
> It seems odd to not preserve the history of events, just to avoid a frontend issue. I'm probably misunderstanding something here though.

CristinImport should be run once, but most likely we will be running it multiple times, we do not want to persist information about multiple Cristin imports. It is also likely that brage import will be done multiple times, but we should not serve this information to user, we have reports in s3 buckets for each iteration, so we know which publication has been updated and what the changes are from these reports.

PR #1788, opened 2024-08-21
implementing service to fetch publication from search-api

truhacevkir:
This is part 1 of implementing logic for repairing imported publications. Implementing SearchService which returns publications fetched by from search-api and ResourceService.


brinxmat (commented):
Fix the codacy errors

PR #1792, opened 2024-08-26
mapping orcId for brage contributors

truhacevkir:
Mapping orcId from Brage.

### DIFF
--- a/brage-import/src/main/java/no/sikt/nva/brage/migration/record/Identity.java
+++ b/brage-import/src/main/java/no/sikt/nva/brage/migration/record/Identity.java
@@ -12,7 +12,8 @@ public class Identity {
     private final String identifier;
     private final String orcId;
 
-    public Identity(@JsonProperty("name") String name, @JsonProperty("identifier") String identifier,
+    public Identity(@JsonProperty("name") String name,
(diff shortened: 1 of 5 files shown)


brage-import/src/main/java/no/sikt/nva/brage/migration/record/Identity.java
```diff
@@ -10,11 +10,17 @@ public class Identity {
 
     private final String name;
     private final String identifier;
+    private final String orcId;
 
-    public Identity(@JsonProperty("name") String name,
-                    @JsonProperty("identifier") String identifier) {
+    public Identity(@JsonProperty("name") String name, @JsonProperty("identifier") String identifier,
```
brinxmat, line 15: Beginning to agree with @LarsV123 at the code style should be stricter since we aren't fastidious in readable formatting. Break the parameters to new lines.

brinxmat (approved):

brage-import/src/main/java/no/sikt/nva/brage/migration/record/Identity.java
truhacevkir: Should we modify our codestyle scheme?

brinxmat (approved):

PR #1793, opened 2024-08-26
Np 47377 fixing migrated licenses

truhacevkir:
Adding support for updating licenses for publications.

### DIFF
--- a/publication-event-handlers/src/main/java/no/unit/nva/publication/events/handlers/batch/ManuallyUpdatePublicationUtil.java
+++ b/publication-event-handlers/src/main/java/no/unit/nva/publication/events/handlers/batch/ManuallyUpdatePublicationUtil.java
@@ -22,14 +22,10 @@ public final class ManuallyUpdatePublicationUtil {
     }
 
     public List<Publication> update(List<Publication> publications, ManuallyUpdatePublicationsRequest request) {
-        if (ManualUpdateType.PUBLISHER.equals(request.type())) {
-            return updatePublisher(publications, request);
-        }
-        if (ManualUpdateType.LICENSE.equals(request.type())) {
-            return updateLicense(publications, request);
-        } else {
-            throw new IllegalArgumentException("Unsupported update type: " + request.type());
-        }
+        return switch (request.type()) {
+            case PUBLISHER -> updatePublisher(publications, request);
+            case LICENSE -> updateLicense(publications, request);
+        };
     }
 
     private static Publication update(Publication publication, String oldPublisher, String newPublisher) {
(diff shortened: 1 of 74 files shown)


publication-model/src/main/java/no/unit/nva/model/associatedartifacts/file/File.java
```diff
@@ -418,5 +422,17 @@ public File buildUnpublishableFile() {
...
+            if (clazz.equals(PublishedFile.class)) {
+                return buildPublishedFile();
+            } else if (clazz.equals(UnpublishedFile.class)) {
+                return buildUnpublishedFile();
+            } else if (clazz.equals(AdministrativeAgreement.class)) {
+                return buildUnpublishableFile();
+            } else {
+                throw new IllegalArgumentException("Invalid file type");
+            }
+        }
```
truhacevkir, line 436: I think this builder is a more convenient to use when you consume file without knowing its actual type.

publication-event-handlers/src/test/java/no/unit/nva/publication/events/handlers/batch/UpdatePublicationsInBatchesHandlerTest.java
```diff
@@ -97,10 +103,73 @@ void shouldNotUpdatePublicationPublisherWhenPublicationsHaveDifferentPublisherTh
...
+                         updatedPublication.getAssociatedArtifacts().size());
+
+            var updatedFiles = getFiles(updatedPublication);
+
+            updatedFiles.forEach(file -> assertEquals(license, file.getLicense()));
+        });
+    }
+
+    @Test
+    void shouldThrowIllegalArgumentExceptionWhenUnknownUpdateType() {
```
brinxmat, line 154: This will fail due to the change made to `ManuallyUpdatePublicationUtil.java`, however, validate the creation of the enum object from JSON

brinxmat (changes requested):

publication-event-handlers/src/main/java/no/unit/nva/publication/events/handlers/batch/ManuallyUpdatePublicationUtil.java
```diff
@@ -0,0 +1,101 @@
...
+                                        .build();
+        publication.getEntityDescription().getReference().setPublicationContext(newPublicationContext);
+        return publication;
+    }
+
+    private static Publisher createNewPublisher(Publication publication, String oldPublisher, String newPublisher) {
+        return Optional.of(publication.getEntityDescription().getReference().getPublicationContext())
+                   .map(Book.class::cast)
+                   .map(Book::getPublisher)
+                   .map(Publisher.class::cast)
```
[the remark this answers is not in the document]
truhacevkir: Changing similar method in test so they are not identical.

brinxmat (approved):

PR #1794, opened 2024-08-27
importing Cristin publications without associatedLinks when invalid link

truhacevkir:
When importing Cristin publication
And publication has link
And link is not valid url
Then import publication without that link

### DIFF
--- a/cristin-import/src/main/java/no/unit/nva/cristin/mapper/nva/exceptions/ExceptionHandling.java
+++ b/cristin-import/src/main/java/no/unit/nva/cristin/mapper/nva/exceptions/ExceptionHandling.java
@@ -33,9 +33,6 @@ public final class ExceptionHandling {
         if (exception instanceof DuplicateDoiException) {
             return (DuplicateDoiException) exception;
         }
-        if (exception instanceof InvalidArchiveException) {
-            return (InvalidArchiveException) exception;
-        }
         if (exception instanceof RuntimeException) {
             return (RuntimeException) exception;
         }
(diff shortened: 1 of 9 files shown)


cristin-import/src/test/resources/features/MediaRules.feature
```diff
@@ -56,8 +56,8 @@ Feature: Rules that apply for Media
...
     Then the NVA publication should contain associatedArtifacts containing associatedLink with "<url>"
     Examples:
-      | urltypekode | url              |
-      | FULLTEKST   | www.example.com  |
-      | DATA        | www.example.com  |
-      | OMTALE      | www.example.com  |
+      | urltypekode | url                     |
+      | FULLTEKST   | https://www.example.com |
+      | DATA        | https://www.example.com |
+      | OMTALE      | https://www.example.com |
```
truhacevkir, line 62: Adding scheme to these test uri's.

cristin-import/src/main/java/no/unit/nva/cristin/mapper/nva/exceptions/ExceptionHandling.java
```diff
@@ -33,9 +33,6 @@ public static RuntimeException castToCorrectRuntimeException(Exception exception
         if (exception instanceof DuplicateDoiException) {
             return (DuplicateDoiException) exception;
         }
-        if (exception instanceof InvalidArchiveException) {
-            return (InvalidArchiveException) exception;
-        }
```
truhacevkir, line 38: Removing InvalidArchive exception. Instead of not importing publications with invalid uri's, we import them without these uri's.

torbjokv (approved):

PR #1796, opened 2024-08-28
feature(publicationPermissions): do not allow publication owner to unpublish when files are approved #NP-47411

torbjokv:


brinxmat (approved):

ketilaa (approved):
I feel like we contradict ourselves when using the term "approved" also for files that have been rejected by a curator. But I agree this will work as expected.

torbjokv:
> I feel like we contradict ourselves when using the term "approved" also for files that have been rejected by a curator. But I agree this will work as expected.

yes was thinking about it and that we should probably find a way to make differentiate rejected files from the other ones, but there is no way today without looking at the tickets as far as I know? 

I'm also using both words accepted and approved here and we should probably use one of those

PR #1799, opened 2024-08-29
updating event time

truhacevkir:
When Cristin publication of type Event has been imported without time
And incoming publication has time
Then update publication with time.

- We have updated import rule for event, before we did not map time when time was missing one of values, start date or end date, now we import time when it is missing one of dates.

### DIFF
--- a/cristin-import/src/test/java/no/unit/nva/cristin/lambda/PublicationUpdaterTest.java
+++ b/cristin-import/src/test/java/no/unit/nva/cristin/lambda/PublicationUpdaterTest.java
@@ -32,7 +32,7 @@ class PublicationUpdaterTest {
     }
 
     @Test
-    void shouldUpdatedEventWithTimeWhenExistingEventIsMissingTimeButNewEventHasTimeMissing() {
+    void shouldUpdatedEventWithTimeWhenExistingEventIsMissingTimeButNewEventHasTime() {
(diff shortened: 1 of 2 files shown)


cristin-import/src/test/java/no/unit/nva/cristin/lambda/PublicationUpdaterTest.java
```diff
@@ -28,6 +31,29 @@ void shouldReturnUpdatedEventWithNullValuesWhenAllEventValuesAreMissing() {
         assertEquals(publication.getEntityDescription().getReference().getPublicationContext(), emptyEvent());
     }
 
+    @Test
+    void shouldUpdatedEventWithTimeWhenExistingEventIsMissingTimeButNewEventHasTimeMissing() {
```
brinxmat, line 35: Name is wrong?

brinxmat (changes requested):

cristin-import/src/test/java/no/unit/nva/cristin/lambda/PublicationUpdaterTest.java
truhacevkir: Yes, thank you, updating to:
`shouldUpdatedEventWithTimeWhenExistingEventIsMissingTimeButNewEventHasTime`

brinxmat (approved):

PR #1803, opened 2024-08-30
support jsonsubtypes as oneOf in openapi

sondrev:
Annotaed datamodel to document inheritence (jsonSubTypes)

Review notes:
openapi.yaml and schema.yaml: Theese files are autogenerated and based on the new annotation. Changes should be equal between the files
 documentation/: ignore

### DIFF
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -1301,6 +1301,11 @@ components:
           type: string
         type:
           type: string
+      oneOf:
+        - $ref: '#/components/schemas/AdditionalIdentifier'
+        - $ref: '#/components/schemas/HandleIdentifier'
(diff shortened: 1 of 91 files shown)


truhacevkir (approved):
Looks good, not sure about AdministrativeAgreement, cause it is deserializing as UnpublishableFile. Should it have the same name as json served via API?

```
@JsonTypeName(AdministrativeAgreement.TYPE)
public class AdministrativeAgreement extends File {

    public static final String TYPE = "UnpublishableFile";
```

StigNorland (approved):
Wow, det var en voksen PR :)

Jeg har sett over alle java filene, og så ingen opplagte feil.

PR #1804, opened 2024-08-30
Np 47466 republish unpublish publication

truhacevkir:
Republishing publication in UpdatePublicationHandler and persisting completed PublishingRequest.

Implementing republishing in UpdatePublicationHandler and not in the CreateTicketHandler because all ticket creation there is based on curator creating ticket for publication at the same institution as curator institution.
Refactoring this implementation is not worth it, in addition I think it makes more sense to do publication update (republishing is update of publication status) in UpdatePublicationHandler. 

Adding new allowed operation -> REPUBLISH

### DIFF
--- a/publication-rest/src/main/java/no/unit/nva/publication/update/RepublishPublicationRequest.java
+++ b/publication-rest/src/main/java/no/unit/nva/publication/update/RepublishPublicationRequest.java
@@ -1,5 +1,11 @@
 package no.unit.nva.publication.update;
 
-public class RepublishPublicationRequest implements PublicationRequest{
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import com.fasterxml.jackson.annotation.JsonTypeName;
 
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
+@JsonTypeName(RepublishPublicationRequest.TYPE)
+public class RepublishPublicationRequest implements PublicationRequest {
+
+    public static final String TYPE = "RepublishPublicationRequest";
 }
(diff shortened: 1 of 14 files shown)


publication-rest/src/main/java/no/unit/nva/publication/update/RepublishPublicationRequest.java
```diff
@@ -0,0 +1,5 @@
+package no.unit.nva.publication.update;
+
+public class RepublishPublicationRequest implements PublicationRequest{
```
brinxmat, line 3: Format the file

publication-rest/src/test/java/no/unit/nva/publication/update/UpdatePublicationHandlerTest.java
```diff
@@ -1746,6 +1747,54 @@ void publicationOwnerShouldBeAbleToUpdateMetadataOfPublishedFile() throws ApiGat
...
+        var gatewayResponse = GatewayResponse.fromOutputStream(output, Publication.class);
+
+        var republishedPublication = resourceService.getPublication(publication);
+
+        assertThat(gatewayResponse.getStatusCode(), is(equalTo(HTTP_OK)));
+        assertThat(republishedPublication.getStatus(), is(equalTo(PUBLISHED)));
+    }
+
+    @Test
+    void shouldReturnForbiddenWhenRepublishingNotUnpublishedPublication()
```
brinxmat, line 1771: Not unpublished, or "published"? If "published" 200 OK is also OK since they get what they want

brinxmat (approved):

PR #1810, opened 2024-09-05
Remove old gmazzo classpath declaration

brinxmat:
For some reason, we had an old-style declaration of the gmazzo-plugin that was causing us to use an ancient version. This in turn added a number of deprecation warnings related to deprecated conventions plugin methods.

The actual usage here is a bit weird, if I'm honest, but I don't see another simple way of saying what the model version is at first glance, so I am just fixing the issue in gradle for now.

Also bumping to gradle 8.10


truhacevkir (approved):

ketilaa (approved):
Always pin dependencies and plugins so that build are reproducible 👍

PR #1812, opened 2024-09-05
Np 47515 allow updating publication with existing files

truhacevkir:
Allowing user to update metadata for publication type where customer config does not allow to update/create new files. 
Not sure if updating existing file metadata should be allowed, for example license. 
Implementing that any file change is not allowed. 

https://sikt.atlassian.net/browse/NP-47515


brinxmat (approved):

publication-rest/src/test/java/no/unit/nva/publication/update/UpdatePublicationHandlerTest.java
ketilaa, line 1: We should keep a test for rejecting if a new files is added in the update requrest, right?

PR #1816, opened 2024-09-06
fix: Expand joinField to include all anthologies

LarsV123:
Ticket: https://sikt.atlassian.net/browse/NP-47516

This expands the injection of "joinField" in expansion logic to include all possible anthology containers as "parents" in the parent/child relation.

### DIFF
--- a/expansion/src/main/java/no/unit/nva/expansion/model/ExpandedResource.java
+++ b/expansion/src/main/java/no/unit/nva/expansion/model/ExpandedResource.java
@@ -45,9 +45,12 @@ import no.unit.nva.model.contexttypes.Anthology;
 import no.unit.nva.model.contexttypes.PublicationContext;
 import no.unit.nva.model.instancetypes.PublicationInstance;
 import no.unit.nva.model.instancetypes.book.BookAnthology;
+import no.unit.nva.model.instancetypes.book.BookMonograph;
 import no.unit.nva.model.instancetypes.chapter.ChapterArticle;
 import no.unit.nva.model.instancetypes.chapter.ChapterConferenceAbstract;
 import no.unit.nva.model.instancetypes.chapter.ChapterInReport;
+import no.unit.nva.model.instancetypes.report.ReportBasic;
+import no.unit.nva.model.instancetypes.report.ReportBookOfAbstract;
 import no.unit.nva.publication.external.services.RawContentRetriever;
 import nva.commons.core.JacocoGenerated;
 import nva.commons.core.paths.UriWrapper;
@@ -203,7 +206,7 @@ public final class ExpandedResource implements JsonSerializable, ExpandedDataEnt
         var instanceType = reference.getPublicationInstance();
         var publicationContext = reference.getPublicationContext();
 
-        if (instanceType instanceof BookAnthology) {
+        if (instanceTypeCanBeAnthology(instanceType)) {
             addJoinField(sortedJson, JOIN_FIELD_PARENT_LABEL, null);
         } else if (isPartOfAnthology(publicationContext, instanceType)) {
             var parentId = ((Anthology) publicationContext).getId();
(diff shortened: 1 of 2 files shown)


expansion/src/main/java/no/unit/nva/expansion/model/ExpandedResource.java
```diff
@@ -216,14 +219,24 @@ private static void addJoinFieldWhenAnthology(ObjectNode sortedJson, Reference r
...
+                || instanceType instanceof ReportBasic
+                || instanceType instanceof BookAnthology
+                || instanceType instanceof ReportBookOfAbstract;
+    }
+
+    private static boolean instanceTypeCanBePartOfAnthology(PublicationInstance<?> instanceType) {
+        return instanceType instanceof ChapterArticle
+               || instanceType instanceof ChapterConferenceAbstract
+               || instanceType instanceof ChapterInReport;
+    }
```
truhacevkir, line 233: Why not to use publicationContext type when we are checking if resource can be an Anthology?
Instead of listing all the possible "subtypes", we can check "main type".

image

expansion/src/main/java/no/unit/nva/expansion/model/ExpandedResource.java
```diff
@@ -189,41 +184,52 @@ private static JsonNode addFields(String json, Publication publication) {
...
+                            try {
+                                var identifier = SortableIdentifier.fromUri(uri);
+                                return identifier.toString();
+                            } catch (IllegalArgumentException e) {
+                                // Missing or invalid identifier, which is expected for many
+                                // different types of documents.
+                                return null;
+                            }
+                        })
+                .orElse(JOIN_FIELD_DUMMY_PARENT_IDENTIFIER);
```
truhacevkir, line 228: You could extract uri to identifier conversion to method:

```
 private static String getParentIdentifier(ObjectNode sortedJson) {
        return extractPublicationContextUri(sortedJson)
                .filter(uri -> isNotBlank(uri.toString()))
                .map(ExpandedResource::getPublicationIdentifier)
                .orElse(JOIN_FIELD_DUMMY_PARENT_IDENTIFIER);
    }

    private static String getPublicationIdentifier(URI uri) {
        return attempt(() -> SortableIdentifier.fromUri(uri)).map(SortableIdentifier::toString).orElse(failure -> null);
    }
```

expansion/src/main/java/no/unit/nva/expansion/model/ExpandedResource.java
```diff
@@ -189,41 +184,52 @@ private static JsonNode addFields(String json, Publication publication) {
...
-                var parentIdentifier = SortableIdentifier.fromUri(parentId).toString();
-                addJoinField(sortedJson, JOIN_FIELD_CHILD_LABEL, parentIdentifier);
-            } else {
-                addJoinField(sortedJson, JOIN_FIELD_CHILD_LABEL, JOIN_FIELD_DUMMY_PARENT_ID);
-            }
+        } else {
+            var parentIdentifier = getParentIdentifier(sortedJson);
+            addJoinField(sortedJson, JOIN_FIELD_CHILD_LABEL, parentIdentifier);
         }
     }
```
truhacevkir, line 212: Nice, but i think we have to be careful here, because for example when publicationContext is Journal, publicationContext.id is pointing to publication channel, see publication:

https://dev.nva.sikt.no/registration/019054f93468-bfe46458-f0c2-4177-ae47-22aeaf2fd727

```
"publicationContext" : {
        "type" : "Journal",
        "id" : "https://api.dev.nva.aws.unit.no/publication-channels-v2/journal/697AD9B8-B974-4F2D-8187-7B235814B622/2014"
      }
```

I think we want to have a similar check for childs as you implemented for parents, something like:

```
if (canBeChild(reference.getPublicationInstance())) {
            var parentIdentifier = getParentIdentifier(sortedJson);
            addJoinField(sortedJson, JOIN_FIELD_CHILD_LABEL, parentIdentifier);
        }
```

Not sure what should be placed in canBeChild method, but this is a good start:

```
    private static boolean canBeChild(PublicationInstance<? extends Pages> publicationInstance) {
        return publicationInstance instanceof ChapterArticle;
    }
```

truhacevkir (commented):
Looks good, see comment :)
I think if you are not sure about childs, you can merge this and if we found out that there are more child types, we will just add them to `canBeChild()` method.

expansion/src/main/java/no/unit/nva/expansion/model/ExpandedResource.java
LarsV123: Does it matter if we set the tag for journals too? This just circles back to what we discussed yesterday, there's no obvious way to know which implementations are relevant here.

monaullah: I don't think this method will be able to extract journal identifiers:
```
private static String getPublicationIdentifier(URI uri) {
    return attempt(() -> SortableIdentifier.fromUri(uri))
            .map(SortableIdentifier::toString)
            .orElse(failure -> null);
}
```
because journal ids have year as the last path element, example: `https://api.dev.nva.aws.unit.no/publication-channels-v2/journal/C8BDE22A-49B3-45DE-8FF6-98BBFC7AA66C/2008`

LarsV123: True, but then it default to the dummy value. Not ideal, but not necessarily a problem?

truhacevkir (approved):

monaullah (approved):

PR #1817, opened 2024-09-07
Np 47527 hinder occult publication model changes

brinxmat:
*Problem:* When we make changes to the data model, we need to update the model version. Previously, we controlled this by releasing the model as a versioned artifact. Since publication model was moved into publication api, we no longer have this control. Here, I introduce a control for this issue.

This change does a number of closely related things:

- Adds a gradle plugin that 
  - checks if a pre-push hook is present, and if it is the same as the current pre-push hook
  - if not, it adds a pre-push hook automatically the first time the code is built
  - Issue (probably, though maybe not, see discussion below): it overwrites existing project pre-push hooks

- Adds a pre-push hook that
  - checks if there are uncommitted changes (this may affect next steps)
  - fetches origin
  - runs a diff of publication-model between main and current branch and checks if the model version is updated if there are changes

- Issues
  - We naturally install something in `./git/hooks`
  - We make it executable
  - We overwrite existing pre-push, which is an issue if someone has custom scripts (not aware of anyone doing this for publication API)
    - This issue is because the hook _must_ be named `pre-push`, we could possibly append, or append an external hook-script execution to existing files, but this assumes that the hook-file itself is bash, which it doesn't need to be
    - This is the publication-api hook, not a general git hook, so I don't see this as wrong


buildSrc/src/main/groovy/nva.publication.api.git-hooks.gradle
```diff
@@ -0,0 +1,26 @@
...
+def prePushHookPath = file('./.git/hooks/pre-push')
+
+tasks.register('copyPrePushHook', Copy) {
+    from scriptsPrePushPath
+    into prePushHookPath.parentFile
+    onlyIf {
+        !prePushHookPath.exists() || scriptsPrePushPath.text != prePushHookPath.text
+    }
+    doLast {
+        if (!System.properties['os.name'].toLowerCase().contains('windows')) {
```
brinxmat, line 15: Apparently, Win does not need this…not sure if it can be done via gradle in this way either

buildSrc/src/main/groovy/nva.publication.api.git-hooks.gradle
```diff
@@ -0,0 +1,26 @@
...
+}
+
+def scriptsPrePushPath = file('.scripts/pre-push')
+def prePushHookPath = file('./.git/hooks/pre-push')
+
+tasks.register('copyPrePushHook', Copy) {
+    from scriptsPrePushPath
+    into prePushHookPath.parentFile
+    onlyIf {
+        !prePushHookPath.exists() || scriptsPrePushPath.text != prePushHookPath.text
```
brinxmat, line 12: We don't want to make any changes if everything is up to date

buildSrc/src/main/groovy/nva.publication.api.git-hooks.gradle
```diff
@@ -0,0 +1,26 @@
+plugins {
+    id 'base'
+}
+
+def scriptsPrePushPath = file('.scripts/pre-push')
+def prePushHookPath = file('./.git/hooks/pre-push')
```
brinxmat, line 6: The hook must be named exactly `pre-push` in `./git/hooks`

buildSrc/src/main/groovy/nva.publication.api.git-hooks.gradle
```diff
@@ -0,0 +1,26 @@
...
+        if (!System.properties['os.name'].toLowerCase().contains('windows')) {
+            println "Making ${prePushHookPath} executable"
+            if (!prePushHookPath.setExecutable(true)) {
+                throw new GradleException("Failed to make ${prePushHookPath} executable.")
+            }
+        }
+    }
+}
+
+build {
```
brinxmat, line 24: Is there a more appropriate task on which to enforce this?

.scripts/pre-push
```diff
@@ -0,0 +1,43 @@
...
+set -o nounset
+set -o errexit
+
+# If the publication-model has updates that change the model structure,
+# the model version should be updated.
+
+TARGET="publication-model"
+BUILD_FILE="${TARGET}/build.gradle"
+PATTERN="^def modelVersion = '"
+MODEL_CHANGE_MESSAGE="Changes in the $TARGET package, which imply\
```
brinxmat, line 11: This message contains what to do if you have changes that don't require a model-version update.

.scripts/pre-push
```diff
@@ -0,0 +1,43 @@
...
+ the model version number may need updating. If there are no effective model structure changes,\
+ override this pre-push hook with $ git push --no-verify"
+NO_MODEL_CHANGE_MESSAGE="No changes to $TARGET. Continuing."
+
+cd "$(git rev-parse --show-toplevel)"
```
brinxmat, line 16: Make sure our relative paths are relative to what they should be

.scripts/pre-push
```diff
@@ -0,0 +1,43 @@
...
+
+if ! git diff-index --quiet HEAD --; then
+  echo "You have uncommitted changes. Please commit or stash them before pushing."
+  exit 1
+fi
```
brinxmat, line 21: Can't do a diff with uncommitted changes, alas

ketilaa (approved):
Seems ok to me. Let us try it in practise.

.scripts/pre-push
```diff
@@ -0,0 +1,43 @@
...
+# If the publication-model has updates that change the model structure,
+# the model version should be updated.
+
+TARGET="publication-model"
+BUILD_FILE="${TARGET}/build.gradle"
+PATTERN="^def modelVersion = '"
+MODEL_CHANGE_MESSAGE="Changes in the $TARGET package, which imply\
+ the model version number may need updating. If there are no effective model structure changes,\
+ override this pre-push hook with $ git push --no-verify"
+NO_MODEL_CHANGE_MESSAGE="No changes to $TARGET. Continuing."
```
monaullah, line 14: Receiving this message also when the model version is updated. Maybe misleading?

StigNorland (approved):
Fungerer på windows.
Java versjon er problematisk i intellij

monaullah (approved):

axthosarouris:
I cannot think of a better solution to be honest. Having said that, I will write some thoughts which probably are not very constructive. I am not even sure if they make sense.

It seems strange to me to have two versions in the same project. To the best of my knowledge, package managers do not allow submodules to have different version numbers, and probably for a good reason ( I am not sure I can tell the reason). 

If we have 2 version numbers, many other questions arise. E.g. If we change the model version, do we need to change the api version? Could we have the following transitions in versions and what do they mean?

```
      1.0.0 (PublicatioApi) 1.0.0 (ModelVersion) -> 1.0.1(PublicationApi) 1.1.0(ModelVersion)
      1.0.0 (PublicatioApi) 1.0.0 (ModelVersion) -> 1.1.0(PublicationApi) 1.0.1(ModelVersion)     
```
To me, it seems that when the publication model changes the version of the publication api should change to indicate breaking changes with the previous version. e.g. to go from 1.0.0. to 1.1.0. When the publication model does not change the publication API is basically compatible with its previous version (since it can read the data and the semantics of the data have not changed ) so the version change should indicate that (e.g. 1.0.0 to 1.0.1).

buildSrc/src/main/groovy/nva.publication.api.git-hooks.gradle
torbjokv: could maybe have it on git push so to not affect build time

torbjokv: ah wait, it is pre-push already. do we need this?

PR #1820, opened 2024-09-13
making brage search less strict

truhacevkir:
When searching for duplicates, searching for publications with titles that almost match (does not have to match 100% the provided title) and removing aggregation param, aggregations are not returned by default.

### DIFF
--- a/brage-import/src/main/java/no/sikt/nva/brage/migration/merger/findexistingpublication/TitleAndTypePublicationFinder.java
+++ b/brage-import/src/main/java/no/sikt/nva/brage/migration/merger/findexistingpublication/TitleAndTypePublicationFinder.java
@@ -27,10 +27,8 @@ import org.slf4j.LoggerFactory;
 
 public class TitleAndTypePublicationFinder implements FindExistingPublicationService {
 
-    public static final String TITLE = "title";
+    public static final String TITLE_SHOULD = "titleShould";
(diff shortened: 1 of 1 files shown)


brage-import/src/main/java/no/sikt/nva/brage/migration/merger/findexistingpublication/TitleAndTypePublicationFinder.java
```diff
@@ -125,9 +123,8 @@ private UriWrapper getStandardSearchUri(Publication publication) {
         return UriWrapper.fromHost(apiHost)
                    .addChild(SEARCH)
                    .addChild(RESOURCES)
-                   .addQueryParameter(TITLE, getMainTitle(publication).get())
-                   .addQueryParameter(CONTEXT_TYPE, getPublicationContextType(publication))
-                   .addQueryParameter(AGGREGATION, NONE);
+                   .addQueryParameter(TITLE_SHOULD, getMainTitle(publication).get())
+                   .addQueryParameter(CONTEXT_TYPE, getPublicationContextType(publication));
     }
```
truhacevkir, line 128: We do not mock whole call in test, so no existing tests that fail. Should consider to mock exact api calls in test setup.

brinxmat (approved):

PR #1826, opened 2024-09-18
Np 47374 setting sequence number to related documents

truhacevkir:
When importing phd degrees from Brage, we should present related documents in the same order as in Brage. Attempting to achieve it by setting sequence number.

### DIFF
--- a/brage-import/src/test/java/no/sikt/nva/brage/migration/lambda/BrageEntryEventConsumerTest.java
+++ b/brage-import/src/test/java/no/sikt/nva/brage/migration/lambda/BrageEntryEventConsumerTest.java
@@ -149,7 +149,6 @@ import no.unit.nva.model.instancetypes.degree.DegreeBachelor;
 import no.unit.nva.model.instancetypes.degree.DegreeMaster;
 import no.unit.nva.model.instancetypes.degree.DegreePhd;
 import no.unit.nva.model.instancetypes.degree.OtherStudentWork;
-import no.unit.nva.model.instancetypes.degree.UnconfirmedDocument;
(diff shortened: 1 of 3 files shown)


brage-import/src/test/java/no/sikt/nva/brage/migration/lambda/BrageEntryEventConsumerTest.java
```diff
@@ -1114,50 +1113,6 @@ void shouldMergePublicationContextAndUseCourseFromBragePublicationWhenPublicatio
         assertThat(((UnconfirmedCourse) actualPublicationContext.getCourse()).code(), is(equalTo(subjectCode)));
     }
 
-    @Test
-    void shouldKeepOrderOfRelatedDocumentsWhenMergingDegreePhd()
```
truhacevkir, line 1118: Not relevant longer, we set sequence number instead of sorting values in a list.

brage-import/src/test/java/no/sikt/nva/brage/migration/mapper/BrageNvaMapperTest.java
```diff
@@ -111,23 +109,6 @@ void shouldNotCreateAlternativeAbstractsWhenSingleAbstractInBrage()
         assertThat(publication.getEntityDescription().getAlternativeAbstracts(), is(anEmptyMap()));
     }
 
-    @Test
-    void shouldCreatePublicationWithRelatedDocumentsSortedAlphabeticallyWhenMappingDoctoralThesis()
```
truhacevkir, line 115: No relevant longer, we are setting sequence number now.

brinxmat (approved):

PR #1827, opened 2024-09-18
mapping wiseflow and inspera identifiers from Brage

truhacevkir:
Mapping wiseflow and inspera identifiers from Brage to additional identifiers.


brage-import/src/main/java/no/sikt/nva/brage/migration/mapper/BrageNvaMapper.java
```diff
@@ -163,12 +166,27 @@ private static AssociatedLink extractAssociatedLink(Record brageRecord) {
...
+                         extractWiseflowIdentifier(brageRecord))
                    .filter(Optional::isPresent)
                    .map(Optional::get)
                    .collect(Collectors.toSet());
     }
 
+    private static Optional<? extends AdditionalIdentifierBase> extractInsperaIdentifier(Record brageRecord) {
+        return isNull(brageRecord.getInsperaIdentifier())
+                   ? Optional.empty()
+                   : Optional.of(new AdditionalIdentifier(INSPERA_SOURCE, brageRecord.getInsperaIdentifier()));
```
torbjokv, line 181: Should we create dedicated AdditionalIdentifier types for this? especially if this is new data in our system, I think it would be a good idea, if this already exists in prod its probably ok

torbjokv (changes requested):

brage-import/src/main/java/no/sikt/nva/brage/migration/mapper/BrageNvaMapper.java
brinxmat: See other PR…I imagine these values are URNs

brage-import/src/test/java/no/sikt/nva/brage/migration/mapper/BrageNvaMapperTest.java
```diff
@@ -33,6 +34,7 @@
 import no.unit.nva.model.instancetypes.degree.DegreePhd;
 import no.unit.nva.model.instancetypes.degree.UnconfirmedDocument;
 import nva.commons.core.Environment;
+import org.jetbrains.annotations.NotNull;
```
brinxmat, line 37: ?

brage-import/src/main/java/no/sikt/nva/brage/migration/mapper/BrageNvaMapper.java
truhacevkir: I am thinking about keeping it as AdditionalIdentifier for now, because instance SourceName.instanceName for this identifiers does not make much sense, cause it will be missing when importing from Inspera and Wiseflow directly, and refactoring to serialize SourceName as object requires migration in test.

brinxmat (approved):

PR #1852, opened 2024-10-02
Filter tickets for other institution from listing

truhacevkir:
Changes: 
- When creating ticket, setting curatingCustomer - maybe should change to curatingInstitution (cristinTopLevelOrg)
- CuratingCustomer is a customer of the user who creates ticket.
- When listing tickets for publication-> list all tickets with the same curating institution and all completed publishingRequests and DoiRequests not depending of institution. 
- Removing restriction that person at the same customer as ticket is allowed to fetch ticket.


brinxmat (approved):

brinxmat:
I suspect that there are a number of pitfalls with this approach, such as DOI…whose DOI is it? The owner institution or the curator institution?

PR #1857, opened 2024-10-03
setting only upload files as filesForApproval when updating publication

truhacevkir:
Changes:
- When updating publication, adding only new approved files to existing publishing request. 

Cases:

```
 @DisplayName("When publishing curator updates publication with unpublished file" +
                 "and removes existing unpublished file" +
                 "and there exists pending publishing request for curator institution with unpublished file to remove" +
                 "then new unpublished file only is added to approved files in PublishingRequest")
```

```
    @DisplayName("When publishing curator updates publication with unpublished file" +
                 "and publication already contains another unpublished file" +
                 "and there exists pending publishing request for curator institution with existing unpublished file" +
                 "then both existing and new unpublished files are added to approved files in PublishingRequest")
```

```
    @DisplayName("When publishing curator updates publication with unpublished file" +
                 "and publication already contains another unpublished file" +
                 "and there is no pending publishing request for the publication with curator owner affiliation" +
                 "then only new unpublished file is added to approved files in PublishingRequest")
```

```
    @DisplayName("When contributor updates publication with unpublished file" +
                 "and publication already contains another unpublished file" +
                 "and there is no pending publishing request for the publication with contributor owner affiliation" +
                 "then only new unpublished file is added to files for approval in PublishingRequest")
```

There are probably more complex cases that we will find out when we will test behavior in frontend. 

I think we will need to refactor the whole ticket/filesForApproval/approvedFiles logic to own service that will hold all the logic, otherwise it will be a mess (it already is)

### DIFF
--- a/publication-commons/src/main/java/no/unit/nva/publication/model/business/PublishingRequestCase.java
+++ b/publication-commons/src/main/java/no/unit/nva/publication/model/business/PublishingRequestCase.java
@@ -10,7 +10,6 @@ import static no.unit.nva.publication.model.business.TicketEntry.Constants.OWNER
 import static no.unit.nva.publication.model.business.TicketEntry.Constants.OWNER_FIELD;
 import static no.unit.nva.publication.model.business.TicketEntry.Constants.STATUS_FIELD;
 import static no.unit.nva.publication.model.business.TicketEntry.Constants.WORKFLOW;
-import static no.unit.nva.publication.model.business.TicketStatus.COMPLETED;
 import static nva.commons.core.attempt.Try.attempt;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonTypeInfo;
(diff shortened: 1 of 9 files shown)


publication-commons/src/main/java/no/unit/nva/publication/model/business/PublishingRequestCase.java
```diff
@@ -149,19 +138,10 @@ public void validateCompletionRequirements(Publication publication) {
     @Override
     public PublishingRequestCase complete(Publication publication, Username finalizedBy) {
         var completed = (PublishingRequestCase) super.complete(publication, finalizedBy);
-        completed.setApprovedFiles(getFilesToApprove(publication));
```
truhacevkir, line 152: We cant do it by default longer, cause we have to manage which files to approve when updating publication. Cause different institutions are now responsible for files uploaded by institution users (curators/contributors)

PR #1860, opened 2024-10-04
feature(resource-model): add new file types K215 #NP-47724

torbjokv:
plan is to add this in backend, so that we can try out frontend a bit, implement access checks and etc in download and view before doing the migration.


brinxmat (approved):
File should have been an interface, I suspect

PR #1864, opened 2024-10-04
Moving pu req logic to own class

truhacevkir:
Moving all PublishingRequest logic to own class when updating publication.


brinxmat (approved):
Given communication outside the PR, I'm approving as-is

PR #1868, opened 2024-10-08
using funding code english when generating funding id

truhacevkir:
`finansieringskildekode_engelsk` is a funding code which is used in cristin-api. This is the code we should use when generating funding id.


cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinMapper.java
truhacevkir, line 1: Some clean up here

LarsV123 (approved):

PR #1872, opened 2024-10-09
updating publication comparator when importing publications

truhacevkir:
When berging publication from Brage with existing publication in NVA, comparing last name of contributor instead of whole name value. Multiple cases where contributors from Cristin have full names including middle names, but Brage contributors have first name and last name only. 

+ 

Ignoring student thesis from UiO when importing from Brage

### DIFF
--- a/brage-import/src/main/java/no/sikt/nva/brage/migration/lambda/PublicationComparator.java
+++ b/brage-import/src/main/java/no/sikt/nva/brage/migration/lambda/PublicationComparator.java
@@ -1,7 +1,6 @@
 package no.sikt.nva.brage.migration.lambda;
 
 import static java.util.Objects.isNull;
-import static java.util.Objects.nonNull;
 import java.util.List;
 import java.util.Locale;
 import java.util.Objects;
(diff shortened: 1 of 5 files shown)


brage-import/src/main/java/no/sikt/nva/brage/migration/lambda/PublicationComparator.java
```diff
@@ -97,15 +100,21 @@ private static boolean atleasOneContributorMatch(Publication incomingPublication
...
+        return Optional.ofNullable(contributor.getIdentity())
                    .map(Identity::getName)
-                   .filter(value -> levenshteinDistanceIsLessThan(
-                       value,
-                       name,
-                       MAX_ACCEPTABLE_CONTRIBUTOR_NAME_LEVENSHTEIN_DISTANCE)
-                   );
+                   .map(name -> name.split(StringUtils.SPACE))
+                   .map(nameArray -> nameArray[nameArray.length - 1])
+                   .orElse(null);
```
brinxmat, line 117: Not sure what these methods are doing here at all

truhacevkir: It gets last name of contributor. Splitting "name" by whitespaces and getting last value.

PR #1880, opened 2024-10-15
Persisting completed pub req in single database transaction

truhacevkir:
Persisting completed PublishingRequest in two transactions causes sometimes processing of EventBridge events in unwanted order. Event with completed ticket is being proceeded before event with new ticket. It causes that search index is not updated with latest image of database entry. 

- Persisting completed PublishingRequest in single database transaction.

### DIFF
--- a/publication-commons/src/main/java/no/unit/nva/publication/model/business/PublishingRequestCase.java
+++ b/publication-commons/src/main/java/no/unit/nva/publication/model/business/PublishingRequestCase.java
@@ -341,9 +341,7 @@ public class PublishingRequestCase extends TicketEntry {
     public PublishingRequestCase persistAutoComplete(TicketService ticketService, Publication publication,
                                                      Username finalizedBy)
         throws ApiGatewayException {
-        var ticket = this.persistNewTicket(ticketService);
-        ticket.complete(publication, finalizedBy).persistUpdate(ticketService);
-        return (PublishingRequestCase) ticket.fetch(ticketService);
+        return (PublishingRequestCase) this.complete(publication, finalizedBy).persistNewTicket(ticketService);
     }
 
     private static PublishingRequestCase createPublishingRequestIdentifyingObject(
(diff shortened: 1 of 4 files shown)


publication-commons/src/main/java/no/unit/nva/publication/model/business/TicketEntry.java
```diff
@@ -127,6 +127,7 @@ public static void setServiceControlledFields(TicketEntry ticketEntry,
         ticketEntry.setCreatedDate(now);
         ticketEntry.setModifiedDate(now);
         ticketEntry.setIdentifier(identifierProvider.get());
+        ticketEntry.setFinalizedDate(nonNull(ticketEntry.getFinalizedDate()) ? now : null);
```
truhacevkir, line 130: Is it okay to handle it here? Instead of creating identical service methods for persisting completed ticket with only difference that finalized data is set by default.

publication-commons/src/main/java/no/unit/nva/publication/model/business/PublishingRequestCase.java
```diff
@@ -341,9 +341,7 @@ && getStatus() == that.getStatus()
     public PublishingRequestCase persistAutoComplete(TicketService ticketService, Publication publication,
                                                      Username finalizedBy)
         throws ApiGatewayException {
-        var ticket = this.persistNewTicket(ticketService);
-        ticket.complete(publication, finalizedBy).persistUpdate(ticketService);
-        return (PublishingRequestCase) ticket.fetch(ticketService);
```
truhacevkir, line 346: We do not need this fetch, cause `persistNewTicket()` fetches ticket after insertion:
```
    private <T extends TicketEntry> T createTicketForPublication(Publication publication,
                                                                 TicketEntry ticketEntry)
        throws ConflictException {

        setServiceControlledFields(ticketEntry, identifierProvider);
        ticketEntry.validateCreationRequirements(publication);
        var request = ticketEntry.toDao().createInsertionTransactionRequest();
        sendTransactionWriteRequest(request);
        FunctionWithException<TicketEntry, TicketEntry, NotFoundException>
            fetchTicketProvider = this::fetchTicket;
        return (T) fetchEventualConsistentDataEntry(ticketEntry, fetchTicketProvider).orElseThrow();
    }
```

publication-commons/src/main/java/no/unit/nva/publication/model/business/TicketEntry.java
brinxmat: Try it; there might be a few edge-cases we cannot see right now.

brinxmat (approved):

PR #1882, opened 2024-10-15
completing existing publishing request when removing unpublished files

truhacevkir:
It seems like after refactorting I have introduced a bug. 

Not updating filesForApproval after completing publishing request.

### DIFF
--- a/publication-rest/src/main/java/no/unit/nva/publication/update/PublishingRequestResolver.java
+++ b/publication-rest/src/main/java/no/unit/nva/publication/update/PublishingRequestResolver.java
@@ -89,9 +89,11 @@ public final class PublishingRequestResolver {
         var pendingPublishingRequests = fetchPendingPublishingRequestsForUserInstitution(oldImage);
         if (pendingPublishingRequests.isEmpty()) {
             createPublishingRequestOnFileUpdate(oldImage, newImage);
+            return;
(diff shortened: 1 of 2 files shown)


publication-rest/src/main/java/no/unit/nva/publication/update/PublishingRequestResolver.java
```diff
@@ -89,9 +89,11 @@ private void handlePublishingRequest(Publication oldImage, Publication newImage)
...
         if (pendingPublishingRequests.isEmpty()) {
             createPublishingRequestOnFileUpdate(oldImage, newImage);
+            return;
         }
         if (thereAreNoFiles(newImage)) {
             autoCompletePendingPublishingRequestsIfNeeded(newImage, pendingPublishingRequests);
+            return;
         }
         if (updateHasFileChanges(oldImage, newImage)) {
             updateFilesForApproval(oldImage, newImage, pendingPublishingRequests);
```
truhacevkir, line 99: When removing unpublished files, both second and third conditions are true. Second condition approval publishing request. Third one updates files for approval (removes them) and updates completing publishing request to pending. Probably we need to restrict COMPLETED -> PENDING transition.

brinxmat (approved):

PR #1886, opened 2024-10-16
fix(permissions): restrict unpublish for nvi-curators #NP-47874

torbjokv:
https://sikt.atlassian.net/browse/NP-47874

Used the PR to also rename CuratorPermissionStrategy to PublishingCuratorPermissionStrategy

### DIFF
--- a/publication-commons/src/main/java/no/unit/nva/publication/permission/strategy/PublicationPermissionStrategy.java
+++ b/publication-commons/src/main/java/no/unit/nva/publication/permission/strategy/PublicationPermissionStrategy.java
@@ -9,7 +9,7 @@ import no.unit.nva.model.PublicationOperation;
 import no.unit.nva.publication.model.business.UserInstance;
 import no.unit.nva.publication.permission.strategy.grant.BackendClientStrategy;
 import no.unit.nva.publication.permission.strategy.grant.ContributorPermissionStrategy;
-import no.unit.nva.publication.permission.strategy.grant.CuratorPermissionStrategy;
+import no.unit.nva.publication.permission.strategy.grant.PublishingCuratorPermissionStrategy;
 import no.unit.nva.publication.permission.strategy.grant.EditorPermissionStrategy;
 import no.unit.nva.publication.permission.strategy.grant.GrantPermissionStrategy;
 import no.unit.nva.publication.permission.strategy.grant.ResourceOwnerPermissionStrategy;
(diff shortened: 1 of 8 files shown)


torbjokv:
Thinking that we maybe should remove MANAGE_RESOURCES_STANDARD from some roles here https://github.com/BIBSYSDEV/nva-identity-service/blob/main/user-access-handlers/src/main/java/no/unit/nva/handlers/data/DefaultRoleSource.java and also add documentation about what MANAGE_RESOURCES_STANDARD means in words. Or maybe remove it completely

LarsV123:
What is a "PublishingCurator"? I see Confluence lists `Curator` and `Nvi-curator` as distinct roles with no overlap.
If "PublishingCurator" and "Nvi-curator" is the same thing:
- Can we use the same term in all places?
- Is the existing permissions flow also used by `Curator`s (non-NVI), given the name `CuratorPermissionStrategy`?

The Confluence page on roles may be outdated, so please update it if you can: https://sikt.atlassian.net/wiki/spaces/NVAP/pages/2822865096/Users+roles+and+access+rights

torbjokv:
> What is a "PublishingCurator"? I see Confluence lists `Curator` and `Nvi-curator` as distinct roles with no overlap. If "PublishingCurator" and "Nvi-curator" is the same thing:
> 
> * Can we use the same term in all places?
> * Is the existing permissions flow also used by `Curator`s (non-NVI), given the name `CuratorPermissionStrategy`?
> 
> The Confluence page on roles may be outdated, so please update it if you can: https://sikt.atlassian.net/wiki/spaces/NVAP/pages/2822865096/Users+roles+and+access+rights

there is no only "curator" as far as I know, only publishing curator and the other variants (like avi). That is the motivation to rename here. @ketilaa can probably confirm that the column in the nice document there is actually a publishing curator. This change is done to remove some of the overlap that we have in our code today

publication-commons/src/main/java/no/unit/nva/publication/permission/strategy/PublicationPermissionStrategy.java
```diff
@@ -38,7 +38,7 @@ public PublicationPermissionStrategy(
         this.publication = publication;
         this.grantStrategies = Set.of(
             new EditorPermissionStrategy(publication, userInstance, resourceService),
-            new CuratorPermissionStrategy(publication, userInstance, resourceService),
+            new PublishingCuratorPermissionStrategy(publication, userInstance, resourceService),
```
truhacevkir, line 41: Not sure this is the right thing to do. By introducing "PublishingCurator", we lack "SupprotCurator", "DoiCurator", "NviCurator" and other type of curators. I think CuratorPermissionStrategy is general permission check for any type of curator.

torbjokv: hmm yeah I agree! Will change back for this PR, but I think it would be better to separate the concerns in different strategies somehow. Our code today tries to infer roles based on specific access grants and mixes different grants in same strategy.

truhacevkir (approved):

PR #1889, opened 2024-10-16
fakeCustomerApiResponse

monaullah:


expansion/src/main/java/no/unit/nva/expansion/model/ExpandedImportCandidate.java
```diff
@@ -508,9 +504,7 @@ private static boolean isHttpOk(HttpResponse<String> stringHttpResponse) {
     }
 
     private static Optional<CristinOrganization> fetchCristinOrg(URI id, RawContentRetriever uriRetriever) {
-        return Optional.ofNullable(getCristinIdentifier(id))
```
monaullah, line 511: @truhacevkir Can you have a look at this? Is it ok to remove this conversion (cristinId -> identifier -> cristinId with depth and top)?

truhacevkir: I think there is a reason for this variable. It is used to return top level for organization to look up, better to keep ut there. I have checked and `depth=top` returns shorter response for some organizations, in addition Cristin does not need to query through whole table when looking up for organization.

truhacevkir: Maybe we can also use depth=none.

truhacevkir: Have checked correspondence with Tor Arne, when running scopus-import, the are a lot of api calls to Cristin, because of ton of contributors. He suggested these params to reduce a load on Cristin servers

truhacevkir (commented):
Se kommentar, vi bør beholde parameteret som sendes.

expansion/src/main/java/no/unit/nva/expansion/model/ExpandedImportCandidate.java
brinxmat: I'm going to reinstate this code and merge as I need to get on with refactoring. @monaullah, keep this in the back of your mind as we may need to return to the topic as we progress into _greener_ pastures

monaullah: Thanks, Kir!

truhacevkir: Pratet med Tor Arne, Cristin gikk fra å bruke rekursjon for org oppslag til å bruke søk, så nå gjøres det kun to kall på oppslag av organisasjon uansett nivå man søker på. Tor Arne mener at vi kan ta vekk det parameteret (mtp. overbelastning på Cristin). Så jeg tenker det er bare å fjerne `depth=top`

PR #1892, opened 2024-10-17
Np 47907 rRecoveryBatchScanHandler, avoid event loop

monaullah:
Jira task: NP-47907 RecoveryBatchScanHandler - should not run in event loop if retry fails
With the current implementation, `RecoveryBatchScanHandler` will go into an infinite loop if a resource that ends up in DLQ is not successfully expanded on retry.

Change: 
- `RecoveryBatchScanHandler` does not emit new event for self-consumption
- `RecoveryBatchScanHandler` takes `maximumNumberOfMessages` as input


truhacevkir (approved):
Nice!

brinxmat (approved):

PR #1895, opened 2024-10-18
fixing academic commentary + converting contributorCristinIds to Set

truhacevkir:
BatchScan is failing cause AcademicCommentary is not added to JsonSubtypes of InstanceType. 

- Adding AcademicCommentary to JsonSubtypes of InstanceType.
- Converting contributorCristinIds in CuratingInstitution to Set instead of using List.


publication-model/src/main/java/no/unit/nva/model/instancetypes/PublicationInstance.java
```diff
@@ -68,6 +69,7 @@
 
 @JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
 @JsonSubTypes({
+    @JsonSubTypes.Type(name = "AcademicCommentary", value = AcademicCommentary.class),
```
truhacevkir, line 72: AcademicCommentary is a fresh InstanceType that has been added and mapped in last cristin-import run.

torbjokv (approved):

PR #1906, opened 2024-10-30
refactor(publication-rest): upgrade commons to latest to avoid online…

torbjokv:
… requests in tests


buildSrc/src/main/groovy/nva.publication.api.java-conventions.gradle
```diff
@@ -33,6 +33,7 @@ java {
 }
 
 repositories {
+    mavenLocal()
```
torbjokv: ops

brinxmat (approved):

PR #1907, opened 2024-10-30
Rejecting publishing request with new file types that need approval

ketilaa:
PendingInternalFile and PendingOpen file should end up with RejectedFile when publishing request is rejected.


torbjokv (approved):
👍

PR #1917, opened 2024-11-04
fixing publishing of files

truhacevkir:
Refactoring + fixing bug when publishing files.

### DIFF
--- a/publication-event-handlers/src/main/java/no/unit/nva/publication/events/handlers/tickets/AcceptedPublishingRequestEventHandler.java
+++ b/publication-event-handlers/src/main/java/no/unit/nva/publication/events/handlers/tickets/AcceptedPublishingRequestEventHandler.java
@@ -27,7 +27,6 @@ import no.unit.nva.publication.external.services.UriRetriever;
 import no.unit.nva.publication.model.business.DoiRequest;
 import no.unit.nva.publication.model.business.Entity;
 import no.unit.nva.publication.model.business.PublishingRequestCase;
-import no.unit.nva.publication.model.business.PublishingWorkflow;
 import no.unit.nva.publication.model.business.TicketStatus;
 import no.unit.nva.publication.model.business.UserInstance;
 import no.unit.nva.publication.service.impl.ResourceService;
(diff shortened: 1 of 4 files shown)


publication-event-handlers/src/main/java/no/unit/nva/publication/events/handlers/tickets/AcceptedPublishingRequestEventHandler.java
```diff
@@ -180,121 +173,56 @@ private RuntimeException throwException(String message, Failure<?> failure) {
...
-                                                                ? toInternalFile(pendingInternalFile,
-                                                                                 publishingRequestCase)
-                                                                : pendingInternalFile;
-            case PendingOpenFile pendingOpenFile -> publishingRequestCase
-                                                        .getApprovedFiles()
-                                                        .contains(pendingOpenFile.getIdentifier())
-                                                        ? toOpenFile(pendingOpenFile, publishingRequestCase)
-                                                        : pendingOpenFile;
+            case UnpublishedFile unpublishedFile when publishingRequest.fileIsApproved(unpublishedFile) ->
+                toPublishedFile(unpublishedFile, publishingRequest);
```
truhacevkir, line 207: Look at this beauty

publication-event-handlers/src/main/java/no/unit/nva/publication/events/handlers/tickets/AcceptedPublishingRequestEventHandler.java
```diff
@@ -180,121 +173,56 @@ private RuntimeException throwException(String message, Failure<?> failure) {
...
-        logFilePublish(pendingOpenFile, publishingRequestCase);
-        return pendingOpenFile.toOpenFile();
-    }
-
-    private Publication fetchPublication(
-            UserInstance userInstance, SortableIdentifier publicationIdentifier) {
-        return attempt(() -> resourceService.getPublication(userInstance, publicationIdentifier))
-                .orElseThrow();
+    private Publication fetchPublication(SortableIdentifier publicationIdentifier) {
+        return attempt(() -> resourceService.getPublicationByIdentifier(publicationIdentifier)).orElseThrow();
```
truhacevkir, line 217: This line fixes the issue. When fetching publication by using userInstance, it should match resourceOwner of publication. In our case it can differ, since ticket can be owned by other institution than publication.

truhacevkir:
It looks like following method is used only one place in production code, HandleIdentifierEventHandler. 
I suggest to remove this method from resource service, so we do not get the same error other places. 
`resourceService.getPublication(userInstance, publicationIdentifier)`

We should validate access to publication other places than resourceService. User `getPublicationByIdentifier()` method instead.

publication-event-handlers/src/main/java/no/unit/nva/publication/events/handlers/tickets/AcceptedPublishingRequestEventHandler.java
truhacevkir: `case` X `when` condition -> do something

PR #1920, opened 2024-11-05
moving rejecting of files to accepted pub req handler

truhacevkir:
Simplifying logic in backend and moving rejection of file to AcceptedPublishingRequestHandler.
Before: When updating PublishingRequest to status CLOSED in UpdateTicketHandler -> update ticket status to Completed + update files to rejected.
Now: updating ticket status to completed only in UpdateTicketHandler + Handling Closed PubReq in AcceptedPublishingRequestEventHandler and rejecting files there. 

I think it simplifies updates on publication.


torbjokv (approved):
Looks nice! At some point we should probably think about if we should do stuff to dynamodb on dynamodb events at all, but this pr does not do anything new in that sense so probably all good.

PR #1924, opened 2024-11-08
refreshing ticket when publishing metadata

truhacevkir:
Refreshing PublishingRequest when publishing publication metadata. This is needed in order to populate publication status changes in search index. Otherwise we end up with published publication and publishing request which says that publication is not published. It happens as the cause of metadata being published after we update ticket, cause ticket includes in that case "outdated" publication summary.

### DIFF
--- a/expansion/src/test/java/no/unit/nva/publication/indexing/ExpandedResourceTest.java
+++ b/expansion/src/test/java/no/unit/nva/publication/indexing/ExpandedResourceTest.java
@@ -162,19 +162,6 @@ class ExpandedResourceTest extends ResourcesLocalTest {
     }
 
     @Test
-    void shouldReturnIndexDocumentWithValidContributorAffiliationCountryCode() throws Exception {
-        var publication = randomPublication();
-        FakeUriResponse.setupFakeForType(publication, fakeUriRetriever, resourceService);
-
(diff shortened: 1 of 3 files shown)


expansion/src/test/java/no/unit/nva/publication/indexing/ExpandedResourceTest.java
```diff
@@ -161,19 +161,6 @@ void shouldReturnIndexDocumentWithValidReferenceData() throws Exception {
...
-        FakeUriResponse.setupFakeForType(publication, fakeUriRetriever, resourceService);
-
-        var indexDocument = fromPublication(fakeUriRetriever, resourceService, publication);
-        var framedResultNode = indexDocument.asJsonNode();
-
-        var actualCountryCode = framedResultNode.at("/entityDescription/contributors/1/affiliations/0/countryCode")
-                                    .textValue();
-        assertThat(actualCountryCode, is(not(nullValue())));
-    }
-
```
truhacevkir, line 176: This test always fail, @brinxmat knows why.

brinxmat: Have fixed

brinxmat (approved):

PR #1925, opened 2024-11-11
fixing https scheme for cristin imported handles

truhacevkir:
Updating http scheme of Cristin imported handle.

### DIFF
--- a/cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinMapper.java
+++ b/cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinMapper.java
@@ -184,7 +184,12 @@ public class CristinMapper extends CristinMappingModule {
     private URI extractHandle() {
         return Optional.ofNullable(cristinObject.getCristinAssociatedUris())
                    .flatMap(CristinMapper::extractArchiveUri)
+                   .map(this::updateHttpScheme)
                    .orElse(null);
+    }
+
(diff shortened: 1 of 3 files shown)


cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinMapper.java
```diff
@@ -184,9 +184,14 @@ private static PublicationDate convertToPublicationDate(LocalDate publishedDate)
     private URI extractHandle() {
         return Optional.ofNullable(cristinObject.getCristinAssociatedUris())
                    .flatMap(CristinMapper::extractArchiveUri)
+                   .map(this::updateHttpScheme)
                    .orElse(null);
     }
 
+    private URI updateHttpScheme(URI uri) {
+        return UriWrapper.fromHost(uri.toString()).addChild(uri.getPath()).getUri();
```
brinxmat, line 192: why `fromHost(uri.toString())` and not `fromHost(uri.getHost())`?

brinxmat (changes requested):

cristin-import/src/main/java/no/unit/nva/cristin/mapper/CristinMapper.java
truhacevkir: That makes sense

brinxmat (approved):

PR #1932, opened 2024-11-14
fix(MigrationTests): fix flaky test in shouldMigrateFilesForPublishingRequest

torbjokv:


publication-commons/src/test/java/no/unit/nva/publication/service/impl/MigrationTests.java
```diff
@@ -236,7 +237,8 @@ void shouldNotDoOnlineLookupWhenMigrating() {
     @Deprecated
```
LarsV123, line 237: This test is marked as deprecated. Could it simply be removed?

LarsV123 (approved):

PR #1949, opened 2024-11-26
removing administrative agreement boolean variable

truhacevkir:
- Removing administrative agreement boolean parameter from File.class. 
- Removing migration code for file license and setting it to URI.
- Removing class License, has been used in tests only?
- Adding JsonIgnore annotation to isVisibleForNonOwner method so we never serve it to frontend.


publication-model/src/main/java/no/unit/nva/model/associatedartifacts/file/File.java
```diff
@@ -60,8 +58,8 @@ public abstract class File implements JsonSerializable, AssociatedArtifact {
             "https://creativecommons.org/publicdomain/zero/1.0"), "RightsReserved",
                                                               URI.create("http://rightsstatements.org/vocab/InC/1.0/"));
 
-    public static final String MISSING_LICENSE = "The file is not annotated as an administrative agreement and should" +
-                                                 " have a license";
+    public static final String MISSING_LICENSE =
+        "The file is not annotated as an administrative agreement and should" + " have a license";
```
brinxmat, line 62: "This file is served to the public and should therefore have a license", I suspect.

brinxmat (approved):

publication-model/src/main/java/no/unit/nva/model/associatedartifacts/file/File.java
brinxmat: Or "is open"

PR #1953, opened 2024-11-28
Np 48092 updated file should be updated in ticket as well

truhacevkir:
Updating file in filesForApproval + removing needsApproval cause it is the same as File of type PendingFile

### DIFF
--- a/publication-model/src/main/java/no/unit/nva/model/associatedartifacts/file/HiddenFile.java
+++ b/publication-model/src/main/java/no/unit/nva/model/associatedartifacts/file/HiddenFile.java
@@ -1,8 +1,6 @@
 package no.unit.nva.model.associatedartifacts.file;
 
-
 import com.fasterxml.jackson.annotation.JsonCreator;
-import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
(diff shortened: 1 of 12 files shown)


publication-model/src/main/java/no/unit/nva/model/associatedartifacts/file/PendingInternalFile.java
```diff
@@ -89,6 +82,6 @@ public InternalFile approve() {
 
     @Override
     public boolean isNotApprovable() {
-        return isNull(getLicense()) || isNull(getPublisherVersion());
+        return isNull(getLicense());
```
truhacevkir, line 85: We have publication types that does not require files to have version, for example `Utstilling`
Checking license only.

publication-model/src/main/java/no/unit/nva/model/associatedartifacts/file/HiddenFile.java
```diff
@@ -56,12 +54,6 @@ public boolean isVisibleForNonOwner() {
         return false;
     }
 
-    @Override
-    @JsonIgnore
-    public boolean needsApproval() {
-        return false;
-    }
```
truhacevkir, line 63: removing this method, cause needsApproval == PendingFile

brinxmat (approved):

PR #1955, opened 2024-11-29
Np 48156 do not allow publishing files when missing fields flow1

truhacevkir:
Throwing IllegalStateException when attempting to approve not approvable file. Serving Conflict 409 via API.


truhacevkir:
This implementation handles cases where contributor has rights to publish file without approval, but file is not approvable. It will also catch all other cases when attempting to approve not approvable file.

brinxmat (approved):

PR #1956, opened 2024-12-02
creating skeleton handler and openapi spec for fetch log endpoint

truhacevkir:
- Adding new module in publication-api called publication-log with GroupId `no.unit.nva.publication.log`.
- Adding FetchPublicationLogHandler which return response object with empty list. 
- Adding function to template .
- Adding basic endpoint spec to openapi scheme.

### DIFF
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -752,6 +752,50 @@ paths:
         responses: { }
         httpMethod: POST
         type: AWS_PROXY
+  /{publicationIdentifier}/log:
+    get:
+      operationId: fetchPublicationLog
+      summary: Fetch Publication log by identifier
+      parameters:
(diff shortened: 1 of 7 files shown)


docs/openapi.yaml
```diff
@@ -752,6 +752,52 @@ paths:
...
+        responses: {}
+        httpMethod: POST
+        type: AWS_PROXY
+      responses:
+        200:
+          description: OK
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/TicketCollection"
```
ketilaa, line 782: You want to expose tickets directly? Should probably be `PublicationLogResponse`

ketilaa (approved):
Fix type in response in openapi spec.

PR #1960, opened 2024-12-04
Np 48175 creating persist log entry event handler

truhacevkir:
Implementing `PersistLogEntryEventHandler`. 
- Handler consumes DataEntryUpdateEvent -> fetches resources when present -> creates log entry when resource event is present -> removes resource event from publication.

### DIFF
--- a/publication-commons/src/main/java/no/unit/nva/publication/model/business/Resource.java
+++ b/publication-commons/src/main/java/no/unit/nva/publication/model/business/Resource.java
@@ -1,6 +1,7 @@
 package no.unit.nva.publication.model.business;
 
 import static java.util.Objects.nonNull;
+import static nva.commons.core.attempt.Try.attempt;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonTypeInfo;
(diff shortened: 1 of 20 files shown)


publication-commons/src/main/java/no/unit/nva/publication/model/business/Resource.java
```diff
@@ -202,6 +212,10 @@ public List<LogEntry> fetchLogEntries(ResourceService resourceService) {
         return resourceService.getLogEntriesForResource(this);
     }
 
+    public Resource fetch(ResourceService resourceService) {
+        return attempt(() -> resourceService.getResourceByIdentifier(this.getIdentifier())).orElseThrow();
+    }
+
```
truhacevkir, line 218: Adding fetch method to resource.

publication-commons/src/main/java/no/unit/nva/publication/model/business/logentry/LogTopic.java
```diff
@@ -8,7 +8,7 @@ public enum LogTopic {
     FILE_PUBLISHED("FilePublished"),
     FILE_REJECTED("FileRejected"),
     FILE_UPLOADED("FileUploaded"),
-    METADATA_PUBLISHED("PublicationPublished"),
+    PUBLICATION_PUBLISHED("PublicationPublished"),
```
brinxmat, line 11: Maintain order

brinxmat (approved):

PR #1962, opened 2024-12-05
adding business logic to fetch log handler

truhacevkir:
- Fetching publication log from database in FetchLog endpoint.
- Only users that are allowed to do UPDATE on publication are authorized to fetch log. (today it is handled by frontend)

Next step will be to implement fetch customer by Cristin identifier method in IdentityServiceClient in commons, and persist user and customer names when creating log message.


publication-log/src/main/java/no/unit/nva/publication/log/rest/PublicationLogResponse.java
```diff
@@ -4,10 +4,16 @@
...
+import no.unit.nva.publication.model.business.logentry.LogEntry;
 
 @JsonTypeInfo(use = Id.NAME, property = "type")
 @JsonTypeName(PublicationLogResponse.TYPE)
-public record PublicationLogResponse(List<?> logEntries) {
+public record PublicationLogResponse(List<LogEntryDto> logEntries) {
 
     public static final String TYPE = "PublicationLog";
+
+    public static PublicationLogResponse fromLogEntries(List<LogEntry> logEntries) {
```
brinxmat, line 15: Collection

publication-log/src/test/java/no/unit/nva/publication/log/rest/FetchPublicationLogHandlerTest.java
```diff
@@ -1,50 +1,168 @@
...
         var response = GatewayResponse.fromOutputStream(output, PublicationLogResponse.class);
 
         assertEquals(HTTP_OK, response.getStatusCode());
         assertTrue(response.getBodyObject(PublicationLogResponse.class).logEntries().isEmpty());
     }
 
-    private InputStream createRequest(SortableIdentifier identifier) throws JsonProcessingException {
-        var dtoObjectMapper = JsonUtils.dtoObjectMapper;
+    @Test
+    void shouldReturnNotEmptyPublicationLogWhenUserHasRightsToFetchLog() throws IOException, BadRequestException,
```
brinxmat, line 117: I like the pessimism of this coming last

brinxmat (approved):

template.yaml
```diff
@@ -1081,6 +1081,7 @@ Resources:
       Environment:
         Variables:
           ALLOWED_ORIGIN: !Ref AllowedOrigins
+          TABLE_NAME: !Ref NvaResourcesTable
```
LarsV123, line 1084: Why do we need to add this environment variable now? I see there are already references to it in old code (such as `UpdateImportStatusHandler`), so I'm a bit confused.

LarsV123 (approved):

template.yaml
truhacevkir: I have updated body of this handler. It returned "hardcoded" response object with empty list before and did no database interactions. Now it reads entries from database, without providing table name it fails on execution.

PR #1966, opened 2024-12-10
persisting republish log message

truhacevkir:
- Creating own log entry for republish event
- Creating republish method for Resource

Republish entry looks like:
```
 {
            "type": "LogEntry",
            "topic": "PublicationRepublished",
            "timestamp": "2024-12-10T13:25:14.252160195Z",
            "performedBy": {
               
            }
        }
```


publication-commons/src/main/java/no/unit/nva/publication/service/impl/UpdateResourceService.java
```diff
@@ -163,7 +163,6 @@ public void updateOwner(SortableIdentifier identifier, UserInstance oldOwner, Us
     }
 
     public void updateResource(Resource resource) {
-        resource.setModifiedDate(clockForTimestamps.instant());
```
torbjokv, line 166: a bit scare to change this maybe?

torbjokv: we have at least protect against new users not remembering to setModifiedDate when they should

brinxmat: Strong agreement here. It's the responsibility of Resource to set its own modification time

brinxmat (approved):

PR #1968, opened 2024-12-12
moving publishing of resources to resource class

truhacevkir:
- Moving publishing of resource to resource class. 
- Publishing Resource with following method:
 
```
Resource.fromPublication(publication).publish(resourceService, userInstance);
```
Where Resource on its own knows if it is publishable or not:
```
    public void publish(ResourceService resourceService, UserInstance userInstance) throws NotFoundException {
        var resource = this.fetch(resourceService);
        if (PUBLISHED.equals(resource.getStatus())) {
            return;
        }
        if (isNotPublishable(resource)) {
            throw new IllegalStateException("Resource is not publishable.");
        } else {
            resource.publish(userInstance);
            resourceService.updateResource(resource);
        }
    }
```

Instead of using public method from PublishingRequestCase to check if publication is publishable:
`public static void assertThatPublicationHasMinimumMandatoryFields(Publication resource)`


publication-commons/src/test/java/no/unit/nva/publication/service/impl/ResourceServiceTest.java
```diff
@@ -1294,6 +1295,40 @@ void shouldSetResourceEventToNull() throws BadRequestException, NotFoundExceptio
...
+        resourceService.publishPublication(userInstance, peristedPublication.getIdentifier());
+
+        var publishedResource = Resource.fromPublication(peristedPublication).fetch(resourceService);
+
+        Resource.resourceQueryObject(peristedPublication.getIdentifier()).publish(resourceService, userInstance);
+
+        var resource = Resource.fromPublication(peristedPublication).fetch(resourceService);
+
+        assertEquals(PUBLISHED, publishedResource.getStatus());
+        assertEquals(publishedResource.getPublishedDate(), resource.getPublishedDate());
```
brinxmat, line 1313: You might want to`verify` that some method of resource service is not called, rather than just side effects. The first assertion is a bit pointless. Also, you have modified date, I guess?

brinxmat (approved):

torbjokv (approved):

publication-commons/src/main/java/no/unit/nva/publication/model/business/Resource.java
```diff
@@ -219,15 +225,55 @@ public Resource fetch(ResourceService resourceService) throws NotFoundException
...
+        return !isPublished();
+    }
+
+    private boolean isPublished() {
+        return PUBLISHED.equals(this.getStatus());
+    }
+
+    public void republish(ResourceService resourceService, UserInstance userInstance) throws NotFoundException {
+        var resource = this.fetch(resourceService);
+        resource.republish(userInstance);
```
torbjokv, line 265: maybe add same isNotPublished here?

publication-commons/src/main/java/no/unit/nva/publication/model/business/Resource.java
```diff
@@ -219,15 +225,55 @@ public Resource fetch(ResourceService resourceService) throws NotFoundException
...
+            return;
+        }
+        if (this.isNotPublishable()) {
+            throw new IllegalStateException("Resource is not publishable!");
+        } else {
+            this.setStatus(PUBLISHED);
+            var currentTime = Instant.now();
+            this.setPublishedDate(currentTime);
+            this.setResourceEvent(PublishedResourceEvent.create(userInstance, currentTime));
+        }
```
LarsV123, line 247: nitpick: Perhaps the two `publish`methods could be combined? The logic  seem to just check `isNotPublished` twice, in a roundabout way.

LarsV123: Also, I don't think you need to wrap the last bit here in an `else`, since you can't get there if an exception is thrown.

publication-commons/src/main/java/no/unit/nva/publication/model/business/Resource.java
```diff
@@ -219,15 +225,55 @@ public Resource fetch(ResourceService resourceService) throws NotFoundException
...
+        return PUBLISHED.equals(this.getStatus());
+    }
+
+    public void republish(ResourceService resourceService, UserInstance userInstance) throws NotFoundException {
+        var resource = this.fetch(resourceService);
+        resource.republish(userInstance);
+        resourceService.updateResource(resource);
+    }
+
+    private void republish(UserInstance userInstance) {
```
LarsV123, line 269: nitpick: Could `republish` call `publish` directly, after checking if status is valid? The rest of the code appears to be duplicate.

LarsV123 (approved):

publication-commons/src/main/java/no/unit/nva/publication/model/business/Resource.java
truhacevkir: We could do it, but we set to different resource events for these two actions, so I think it is better to differentiate:
When publishing:
`this.setResourceEvent(PublishedResourceEvent.create(userInstance, currentTime));`
When republishing:
`this.setResourceEvent(RepublishedResourceEvent.create(userInstance, timestamp));
`

PR #1969, opened 2024-12-12
adding artistic degree phd to datamodel - need for discussion with po

truhacevkir:
After multiple discussions with Jan Erik and Terje, we have concluded that it is necessary to implement a new type. External testers have identified the `ArtisticPhd` as a critical requirement for our production release.

### DIFF
--- a/publication-model/src/main/java/no/unit/nva/model/instancetypes/degree/ArtisticDegreePhd.java
+++ b/publication-model/src/main/java/no/unit/nva/model/instancetypes/degree/ArtisticDegreePhd.java
@@ -1,0 +1,20 @@
+package no.unit.nva.model.instancetypes.degree;
+
+import static no.unit.nva.model.instancetypes.PublicationInstance.Constants.PAGES_FIELD;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import java.util.Set;
+import no.unit.nva.model.PublicationDate;
+import no.unit.nva.model.pages.MonographPages;
+
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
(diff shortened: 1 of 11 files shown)


publication-model/src/main/resources/publication-ontology.ttl
```diff
@@ -261,6 +261,11 @@ nva:DegreePhd a rdfs:Class ;
   rdfs:isDefinedBy nva: ;
   rdfs:subClassOf nva:Degree .
 
+nva:ArtisticDegreePhd a rdfs:Class ;
+  rdfs:comment "" ;
+  rdfs:isDefinedBy nva: ;
+  rdfs:subClassOf nva:DegreePhd .
+
```
truhacevkir, line 268: Not sure what to add in comment here, to differentiate between DegreePhd and ArtisticDegreePhd.

publication-model/src/main/java/no/unit/nva/model/instancetypes/degree/ArtisticDegreePhd.java
```diff
@@ -0,0 +1,20 @@
...
+
+import static no.unit.nva.model.instancetypes.PublicationInstance.Constants.PAGES_FIELD;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import java.util.Set;
+import no.unit.nva.model.PublicationDate;
+import no.unit.nva.model.pages.MonographPages;
+
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
+public class ArtisticDegreePhd extends DegreePhd {
```
truhacevkir, line 11: I think in case it will be a mistake, migration will be easy to carry out by simply adding @JsonType annotation to DegreePhd.

brinxmat (approved):

PR #1972, opened 2024-12-16
implementing publish publication endpoint

truhacevkir:
- Implementing Publish publication endpoint:
` POST: /{publicationIdentifier}/publish - No request body required - PublicationResponse in response body`
- Not allowing to publish publication with status `DELETED`. This is not something we support in GUI. 

Why?

Since we now persist log entries for publication update statuses, there is no need to publish metadata via `CreateTicketHandler + AcceptedPublishingRequestEventHandler` by persisting PublishingRequestCase and approving it in order to be able to populate log.

### DIFF
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -213,6 +213,72 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Publication'
+  /{publicationIdentifier}/publish:
+    post:
+      operationId: publishPublication
+      summary: Publish publication by identifier
+      security:
+        - CognitoUserPool: [
+          'https://api.nva.unit.no/scopes/backend',
+          'https://api.nva.unit.no/scopes/frontend',
+          'aws.cognito.signin.user.admin',
+          'https://api.nva.unit.no/scopes/third-party/publication-upsert'
+        ]
+      requestBody:
+        required: false
+        content: {}
+      parameters:
+        - in: path
+          name: publicationIdentifier
+          schema:
+            type: string
+          required: true
+          style: simple
+          explode: false
+          description: UUID identifier of the Publication to publish.
+      x-amazon-apigateway-integration:
+        uri:
+          Fn::Sub: arn:aws:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${PublishPublicationFunction.Arn}/invocations
+        responses: { }
+        httpMethod: POST
+        type: AWS_PROXY
+      responses:
+        200:
+          description: OK - Publication is published
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/PublicationResponse'
+        400:
+          description: Bad Request - Publication is not publishable
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Problem"
+        401:
+          description: Unauthorized - User is not authorized
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Problem"
+        403:
+          description: Forbidden - User has no permissions to publish publication
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Problem"
+        404:
+          description: Not found - Publication does not exist
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Problem"
+        502:
+          description: Bad Gateway - Error in external service
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Problem"
   /by-owner:
     get:
       operationId: listPublicationsByOwner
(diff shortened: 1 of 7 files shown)


docs/openapi.yaml
```diff
@@ -213,6 +213,72 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Publication'
+  /{publicationIdentifier}/publish:
```
brinxmat, line 216: Here we're doing HTTP-RPC; it is worth considering if this is the pattern we want. If it is, we need to document that we are doing /{noun}/{id}/{verb} and that this has a particular meaning. I'd be tempted to differentiate the verb further with some marker (such as Google's :verb)

publication-rest/src/main/java/no/unit/nva/publication/update/PublishPublicationHandler.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    public PublishPublicationHandler(ResourceService resourceService) {
+        super(Void.class);
+        this.resourceService = resourceService;
+    }
+
+    @Override
+    protected void validateRequest(Void unused, RequestInfo requestInfo, Context context) throws ApiGatewayException {
+        //No input to validate
```
brinxmat, line 36: This is unformatted

publication-rest/src/test/java/no/unit/nva/publication/update/PublishPublicationHandlerTest.java
```diff
@@ -0,0 +1,175 @@
...
+
+        var response = GatewayResponse.fromOutputStream(output, Problem.class);
+
+        assertEquals(HTTP_NOT_FOUND, response.getStatusCode());
+    }
+
+    @Test
+    void shouldReturnForbiddenWhenUserHasNoPermissionsToPublishPublication() throws IOException, BadRequestException {
+        var publication = createPublication();
+        var request = createRequestWithRandomUser(publication.getIdentifier());
```
brinxmat, line 80: random user implies that it will occasionally have the correct permissions to publish…

publication-rest/src/test/java/no/unit/nva/publication/update/PublishPublicationHandlerTest.java
```diff
@@ -0,0 +1,175 @@
...
+
+    @Test
+    void shouldReturnOkWhenPublishingPublication()
+        throws IOException, BadRequestException {
+        var publication = createPublication();
+        var request = createRequest(publication);
+
+        handler.handleRequest(request, output, context);
+
+        var response = GatewayResponse.fromOutputStream(output, Problem.class);
```
brinxmat, line 124: Problem.class?

publication-rest/src/test/java/no/unit/nva/publication/update/PublishPublicationHandlerTest.java
```diff
@@ -0,0 +1,175 @@
...
+        return new HandlerRequestBuilder<Void>(dtoObjectMapper).withPathParameters(
+                publicationIdentifierPathParam(publicationIdentifier))
+                   .withCurrentCustomer(randomUri())
+                   .withUserName(randomString())
+                   .withPersonCristinId(randomUri())
+                   .withTopLevelCristinOrgId(randomUri())
+                   .build();
+    }
+
+    private InputStream createRequest(Publication publication) throws JsonProcessingException {
```
brinxmat, line 159: The naming here does not indicate that the user has the correct permissions.

brinxmat: I might have derived the permissions from the publication (i.e. the top level, etc.)

brinxmat (changes requested):

publication-rest/src/test/java/no/unit/nva/publication/update/PublishPublicationHandlerTest.java
truhacevkir: renaming to `createRequestWithUserWithoutPermissions()`

publication-rest/src/test/java/no/unit/nva/publication/update/PublishPublicationHandlerTest.java
truhacevkir: renaming to `createRequestWithUserWithPermissionsToPublishPublication(publication)`

docs/openapi.yaml
truhacevkir: Alternative is to place this under `PUT: /publication/{publicationIdentifier}` with customer request body. But I think that update handler does already a lot of things, but thats probably ok?

truhacevkir: Alternative is to place this under `PUT: /publication/{publicationIdentifier}` with custom request body. But I think that update handler does already a lot of things, but thats probably ok?

torbjokv (approved):

PR #1974, opened 2024-12-18
moving multipart-upload handlers to publication-api

truhacevkir:
- Moving all four handlers from multipart-upload to publication-api
- Creating module `publication-file` -> Maybe download file should be moved here as well?
- Minor refactoring:
When completing multipart upload in CompleteUploadHandler, using `S3 GetObjectMetadata` method instead of `GetObject`, we are interesting in S3 object metadata only, no need to fetch whole object. 
Throwing InternalServerError when S3 interruptions instead of NotFound.
Converting responses and requests to records.


truhacevkir:
PR is not so large :) Simply moving moving code. Template and openapi schema will come in another pr :)

publication-rest/src/main/java/no/unit/nva/publication/uploadfile/CompleteUploadHandler.java
```diff
@@ -0,0 +1,53 @@
+package no.unit.nva.publication.uploadfile;
```
truhacevkir, line 1: Placing these handler in publication-rest.uploadfile as I see download file handlers are also placed here. 
Should consider own file module?

StigNorland: YES!

publication-rest/src/main/java/no/unit/nva/publication/uploadfile/restmodel/CompleteUploadRequestBody.java
```diff
@@ -0,0 +1,33 @@
...
+        return parts().stream().filter(CompleteUploadPart::hasValue).map(CompleteUploadPart::toPartETag).toList();
+    }
+
+    public CompleteMultipartUploadRequest toCompleteMultipartUploadRequest(String bucketName) {
+        var completeMultipartUploadRequest = new CompleteMultipartUploadRequest();
+        completeMultipartUploadRequest.setBucketName(bucketName);
+        completeMultipartUploadRequest.setKey(key());
+        completeMultipartUploadRequest.setUploadId(uploadId());
+        completeMultipartUploadRequest.setPartETags(partETags());
+        return completeMultipartUploadRequest;
```
StigNorland, line 21: In `CompleteUploadResponseBody`, you use the constructor with parameters, I guess I prefer a constructor, compared  to setters....

truhacevkir: Its aws class, I take in use "builder" instead:

```
    public CompleteMultipartUploadRequest toCompleteMultipartUploadRequest(String bucketName) {
        return new CompleteMultipartUploadRequest()
                   .withBucketName(bucketName)
                   .withKey(key())
                   .withUploadId(uploadId())
                   .withPartETags(partETags());
    }
```

torbjokv (approved):

PR #1977, opened 2024-12-19
removing not needed method from resource service

truhacevkir:
Following method is not needed and is in use single place where there is no need to use it:
`
public Publication getPublication(UserInstance userInstance, SortableIdentifier resourceIdentifier)
`
We have multiple fetchByIdentifier() methods to fetch publication


publication-event-handlers/src/main/java/no/unit/nva/publication/events/handlers/identifiers/HandleIdentifierEventHandler.java
```diff
@@ -125,8 +124,7 @@ protected Void processInputPayload(EventReference input) {
             var resourceUpdate = parseResourceUpdateInput(eventBlob);
             if (isPublished(resourceUpdate) && isMissingHandle(resourceUpdate)) {
                 logger.info("Creating handle for publication: {}", resourceUpdate.getIdentifier());
-                var userInstance = UserInstance.create(resourceUpdate.getOwner(), resourceUpdate.getCustomerId());
-                var publication = fetchPublication(userInstance, resourceUpdate.getIdentifier());
+                var publication = fetchPublication(resourceUpdate.getIdentifier());
```
truhacevkir, line 127: No need to fetch publication using user instance when you have publication identifier. Using "simple" fetch method instead.

torbjokv (approved):

PR #1978, opened 2024-12-19
Np 48265 Expansion - Update scientific index endpoint

monaullah:


expansion/src/test/java/no/unit/nva/expansion/ResourceExpansionServiceTest.java
```diff
@@ -670,7 +670,7 @@ void shouldExpandPublicationWithNviStatusWhenPublicationIsReportedNviCandidate()
         var nviStatusNode = json.get(ScientificIndex.SCIENTIFIC_INDEX_FIELD);
 
         assertThat(nviStatusNode.get("year").asText(), is(equalTo("2024")));
-        assertThat(nviStatusNode.get("status").asText(), is(equalTo("Reported")));
+        assertThat(nviStatusNode.get("status").asText(), is(equalTo("REPORTED")));
```
monaullah, line 673: @truhacevkir Might this be a problem for `search-api`?

truhacevkir: I dont think so, I mean search does simple aggregation for scientific index statuses, frontend on other side should be aware of this change.

truhacevkir (approved):
Looks good, but look at my comment, important to fix this.

expansion/src/test/java/no/unit/nva/expansion/ResourceExpansionServiceTest.java
monaullah: Changed json serialization, so it should be serialized "Reported" now

PR #1981, opened 2025-01-07
Add read mes to modules for ease of use

brinxmat:
- Since new developers ask, we include the readmes to define what each module contains

### DIFF
--- a/datacite-commons/README.md
+++ b/datacite-commons/README.md
@@ -4,7 +4,9 @@ Provides a handler to allow reservation of DOIs (non-findable DOIs for publicati
 
 ## Why does this module exist?
 
-This is a short-circuit of a process, which would otherwise be handled as an asynchronous process.  This is an unfortunate optimization for a frontend design that had timing that was considered unacceptable by the end users.
+This is a workaround that short-circuits the usual asynchronous process for generating DOIs.
+
+This is an optimization for a frontend design that requires a synchronous response.
 
 ## Notes
 
--- a/datacite-commons/README.md
+++ b/datacite-commons/README.md
@@ -1,0 +1,11 @@
+# datacite-commons
+
+Provides a handler to allow reservation of DOIs (non-findable DOIs for publications with status "Draft")
+
+## Why does this module exist?
+
+This is a short-circuit of a process, which would otherwise be handled as an asynchronous process.  This is an unfortunate optimization for a frontend design that had timing that was considered unacceptable by the end users.
+
+## Notes
+
+1. The code that otherwise handles DOI creation and update can be found in: [https://github.com/BIBSYSDEV/nva-doi-registrar-client](https://github.com/BIBSYSDEV/nva-doi-registrar-client)
--- a/expansion/README.md
+++ b/expansion/README.md
@@ -1,0 +1,15 @@
+# expansion
+
+In NVAs core data model, it is not necessary to know the exact attributes of every data class (such as Journal or person).
+
+In some cases, it is nice to know these things (when humans search or perform data analysis, or computers evaluate data for further processing (NVI)). This module creates a view of the data that contains all of the "missing" data that is nice to have for other processes. 
+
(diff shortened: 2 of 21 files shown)


datacite-commons/README.md
```diff
@@ -0,0 +1,11 @@
+# datacite-commons
+
+Provides a handler to allow reservation of DOIs (non-findable DOIs for publications with status "Draft")
+
+## Why does this module exist?
+
+This is a short-circuit of a process, which would otherwise be handled as an asynchronous process.  This is an unfortunate optimization for a frontend design that had timing that was considered unacceptable by the end users.
```
LarsV123, line 7: Bit difficult to parse without historical context. Perhaps simply rewrite to something like (assuming I understand correctly):

> This is a workaround that short-circuits the usual asynchronous process for generating DOIs. This is necessary to display new DOIs to end users within an acceptable time-frame.

expansion/README.md
```diff
@@ -0,0 +1,15 @@
+# expansion
+
+In NVAs core data model, it is not necessary to know the exact attributes of every data class (such as Journal or person).
```
LarsV123, line 3:
```suggestion
In NVAs core data model, it is not necessary to know the exact attributes of every data class (such as Journal or Person).
```

messages/README.md
```diff
@@ -0,0 +1,5 @@
+# messages
+
+A message is a textual, human-produced note associated with a task (Ticket) related to a Publication.
+
+This module supports the creation, deletion and reading of messages.
```
LarsV123, line 5: Not really relevant to the PR, but this explanation really makes me question why this is a separate module and not merged with `tickets`. Why is this?

publication-constants/README.md
```diff
@@ -0,0 +1,3 @@
+# publication-constants
+
+Contains a database constants that are used in several modules.
```
LarsV123, line 3: Again, not your fault or relevant, but couldn't this simply be moved to `publication-commons`?  :-(

LarsV123, line 3:
```suggestion
Contains common database constants that are used in several modules.
```

publication-event-handlers/README.md
```diff
@@ -0,0 +1,24 @@
...
+  - identifiers (Handles hdl.net identifiers)
+  - initialization (NOT IN USE)
+  - log (Handles log-entry creation)
+  - persistence (Handles file persistence operations)
+  - recovery (Handles recovery of failed events from persistence to expansion)
+  - tickets (Handles events related to tickets)
+
+## Notes
+
+- There are a few candidates for removal/moving here
```
LarsV123, line 23: Perhaps expand on this, or remove it? Seems like it's missing something.

LarsV123 (approved):
Very nice, great to have a basic explanation for each module.

publication-constants/README.md
brinxmat: I think the major issue is that publication-commons contains far too much. Would we prefer smaller, specific commons modules that are truly common to all modules that use them (i.e. no "used by two, imported by ten")? There is a naming issue "publication-constants" --> "database-constants".

LarsV123: Good point, but this module is only used in `publication-commons` and `publication-testing`, so it doesn't actually solve that problem I think.

messages/README.md
brinxmat: It is a good question — we discussed various things that don't make much sense at the current time. We should take the time to analyse how we have structured the code here. There are some artefacts from previous times that we have not had the time to address. See task NP-48338

publication-constants/README.md
brinxmat: See task NP-48338

datacite-commons/README.md
brinxmat: Re-wrote this to match what was intended based on what you commented.

truhacevkir (approved):

PR #1985, opened 2025-01-09
implementing model for file entry

truhacevkir:
Introducing business object and dao for file. 
Adding Jacoco annotation to FileDao, will remove it in next pr, once we implement persistence logic.


publication-commons/src/main/java/no/unit/nva/publication/model/storage/Dao.java
```diff
@@ -104,7 +105,7 @@ public void setVersion(UUID version) {
     }
 
     @Override
-    public final String getPrimaryKeyPartitionKey() {
+    public String getPrimaryKeyPartitionKey() {
```
truhacevkir, line 108: Removing final to be able to override this method in FileDao class, otherwise it will get the same key structure as other entities in databse, i.e. including customer and owner in primary key.

brinxmat (changes requested):

brinxmat (approved):

torbjokv (approved):

LarsV123:
FYI: This build is only broken because it needs changes from #1988

PR #1988, opened 2025-01-10
fix: Remove invalid parameters from OpenAPI spec

LarsV123:
Removes invalid parameters from (currently unused) endpoints in the OpenAPI spec. These errors currently break the build, as the new version of the OpenAPI linter can find them. 

For future reference: This can be reverted if needed, just add the parameter to the path as well.

### DIFF
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -709,14 +709,6 @@ paths:
     post:
       summary: Create a upload
       description: Creates a multipart upload for a file. This is later referenced for preparing file-parts, uploading, and finishing the upload..
-      parameters:
-        - in: path
-          name: publicationIdentifier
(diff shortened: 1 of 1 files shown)


brinxmat:
> Removes invalid parameters from (currently unused) endpoints in the OpenAPI spec. These errors currently break the build, as the new version of the OpenAPI linter can find them.
> 
> For future reference: This can be reverted if needed, just add the parameter to the path as well.

Is this because they are unused, or because we have a misconfiguration in the linter? I believe the syntax is valid.

torbjokv (approved):

LarsV123:
> Is this because they are unused, or because we have a misconfiguration in the linter? I believe the syntax is valid.

The syntax is invalid, because the parameters listed were not present in the path.

PR #1991, opened 2025-01-13
Np 48212 peristing file

truhacevkir:
Adding database methods for persisting, deleting and updating FileEntry. 

The plan is to migrate File identifier from UUID to SortableIdentifier in next pr, so there is no need to map one to another when doing database transactions.

### DIFF
--- a/publication-commons/src/main/java/no/unit/nva/publication/model/business/FileEntry.java
+++ b/publication-commons/src/main/java/no/unit/nva/publication/model/business/FileEntry.java
@@ -1,5 +1,6 @@
 package no.unit.nva.publication.model.business;
 
+import static nva.commons.core.attempt.Try.attempt;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
@@ -7,7 +8,9 @@ import com.fasterxml.jackson.annotation.JsonTypeInfo;
 import com.fasterxml.jackson.annotation.JsonTypeName;
 import java.net.URI;
 import java.time.Instant;
+import java.util.Objects;
 import java.util.Optional;
+import java.util.UUID;
 import no.unit.nva.identifiers.SortableIdentifier;
 import no.unit.nva.model.Publication;
 import no.unit.nva.model.associatedartifacts.file.File;
@@ -50,9 +53,11 @@ public final class FileEntry implements Entity {
                              userInstance.getTopLevelOrgCristinId(), userInstance.getCustomerId(), file);
     }
 
-    public static FileEntry queryObject(SortableIdentifier fileIdentifier, SortableIdentifier resourceIdentifier) {
-        return new FileEntry(resourceIdentifier, null, null, null, null, null,
-                             File.builder().withIdentifier(fileIdentifier).buildHiddenFile());
+    public static FileEntry queryObject(UUID fileIdentifier, SortableIdentifier resourceIdentifier) {
+        return new FileEntry(resourceIdentifier, null, null, null, null, null, File.builder()
+                                                                                   .withIdentifier(UUID.fromString(
+                                                                                       fileIdentifier.toString()))
+                                                                                   .buildHiddenFile());
     }
 
     public static FileEntry fromDao(FileDao fileDao) {
@@ -83,11 +88,9 @@ public final class FileEntry implements Entity {
         throw new UnsupportedOperationException(DO_NOT_USE_THIS_METHOD);
     }
 
-    //TODO: Implement once we implement database logic
-    @JacocoGenerated
     @Override
     public Publication toPublication(ResourceService resourceService) {
-        return null;
+        return attempt(() -> resourceService.getPublicationByIdentifier(getResourceIdentifier())).orElseThrow();
     }
 
     @JacocoGenerated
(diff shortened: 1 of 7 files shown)


publication-commons/src/main/java/no/unit/nva/publication/service/impl/ResourceService.java
```diff
@@ -382,6 +386,28 @@ public List<LogEntry> getLogEntriesForResource(Resource resource) {
...
+        sendTransactionWriteRequest(dao);
+    }
+
+    public Optional<FileEntry> fetchFile(FileEntry fileEntry) {
+        var primaryKey = fileEntry.toDao().primaryKey();
+        var result = client.getItem(new GetItemRequest().withTableName(tableName).withKey(primaryKey));
+        return Optional.ofNullable(result.getItem())
+                   .map(FileDao::fromDynamoFormat)
+                   .map(FileEntry::fromDao);
+    }
```
truhacevkir, line 400: I think it is nice to return Optional here and do not throw ApiGateway NotFoundException as we do some other places, so we handle specific behavior in adapters/ services.

publication-commons/src/main/java/no/unit/nva/publication/service/impl/ResourceService.java
```diff
@@ -382,6 +386,28 @@ public List<LogEntry> getLogEntriesForResource(Resource resource) {
...
+
+    public void deleteFile(FileEntry fileEntry) {
+        var primaryKey = fileEntry.toDao().primaryKey();
+        client.deleteItem(new DeleteItemRequest().withTableName(tableName).withKey(primaryKey));
+    }
```
truhacevkir, line 405: Har deleting file from database.

publication-commons/src/main/java/no/unit/nva/publication/model/storage/FileDao.java
```diff
@@ -65,24 +95,45 @@ public String indexingType() {
...
+    @JsonProperty(BY_TYPE_CUSTOMER_STATUS_INDEX_PARTITION_KEY_NAME)
+    public String getByTypeCustomerStatusPartitionKey() {
+        return null;
+    }
+
+    @Override
+    @JsonProperty(BY_TYPE_CUSTOMER_STATUS_INDEX_SORT_KEY_NAME)
+    public String getByTypeCustomerStatusSortKey() {
+        return null;
+    }
```
truhacevkir, line 108: This is important so we do not persist database entries in unnecessary index, for example as we do with message. 
Something like Customer:315981-12513-125-124124:Status:NOT_IMPLEMENTED. 

This implementation will not persist any entry in ByTypeCustomerStatus secondary index.

publication-commons/src/main/java/no/unit/nva/publication/service/impl/ResourceService.java
```diff
@@ -382,6 +386,28 @@ public List<LogEntry> getLogEntriesForResource(Resource resource) {
...
+
+    public void updateFile(FileEntry fileEntry) {
+        fileEntry.toDao().updateExistingEntry(client);
```
torbjokv, line 408: Should maybe make sure if someone tries to change file reference/id that it is stopped

torbjokv: no worries, already done I see :-) identifiersDoesNotMatch check

torbjokv (approved):

publication-commons/src/main/java/no/unit/nva/publication/model/business/FileEntry.java
```diff
@@ -42,6 +53,25 @@ public static FileEntry create(File file, SortableIdentifier resourceIdentifier,
                              userInstance.getTopLevelOrgCristinId(), userInstance.getCustomerId(), file);
     }
 
+    public static FileEntry queryObject(UUID fileIdentifier, SortableIdentifier resourceIdentifier) {
```
LarsV123, line 56: This appears to only be used in tests, is that correct? Could move it to a test util in that case?

Also, could you rename it to something more  descriptive?

publication-commons/src/test/java/no/unit/nva/publication/service/impl/ResourceServiceTest.java
```diff
@@ -1333,6 +1337,111 @@ void shouldSetAllTicketsToNotApplicableWhenMarkingDraftPublicationAsDraftForDele
         assertEquals(TicketStatus.NOT_APPLICABLE,ticket.fetch(ticketService).getStatus());
     }
 
+    @Test
```
LarsV123, line 1340: These tests all appear to duplicate the same setup, could it be extracted to a common util method?

publication-commons/src/main/java/no/unit/nva/publication/model/business/FileEntry.java
truhacevkir: Yes, all this methods are not used in production yet, it will be the next step (or one of next steps).

publication-commons/src/main/java/no/unit/nva/publication/model/storage/FileDao.java
brinxmat: For once, I'd add a comment to this effect

brinxmat (approved):

PR #1995, opened 2025-01-14
fetching publication with files + adding files to resource

truhacevkir:
- Implementing fetch resources method which should be used once we migrate files as own entries in database. 
- Adding `List<File> files` to Resource. I think it is convenient to have files in own array to avoid casting all the time.

### DIFF
--- a/publication-commons/src/main/java/no/unit/nva/publication/model/business/Resource.java
+++ b/publication-commons/src/main/java/no/unit/nva/publication/model/business/Resource.java
@@ -1,7 +1,6 @@
 package no.unit.nva.publication.model.business;
 
 import static java.util.Objects.nonNull;
-import static no.unit.nva.model.PublicationStatus.DELETED;
 import static no.unit.nva.model.PublicationStatus.DRAFT;
 import static no.unit.nva.model.PublicationStatus.PUBLISHED;
 import static no.unit.nva.model.PublicationStatus.PUBLISHED_METADATA;
@@ -32,6 +31,7 @@ import no.unit.nva.model.ResearchProject;
 import no.unit.nva.model.ResourceOwner;
 import no.unit.nva.model.additionalidentifiers.AdditionalIdentifierBase;
 import no.unit.nva.model.associatedartifacts.AssociatedArtifactList;
+import no.unit.nva.model.associatedartifacts.file.File;
 import no.unit.nva.model.funding.Funding;
 import no.unit.nva.model.funding.FundingList;
 import no.unit.nva.publication.model.business.importcandidate.ImportCandidate;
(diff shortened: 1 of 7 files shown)


publication-commons/src/main/java/no/unit/nva/publication/model/storage/FileDao.java
```diff
@@ -164,12 +164,16 @@ protected User getOwner() {
     @Override
     @JsonProperty(BY_TYPE_AND_IDENTIFIER_INDEX_PARTITION_KEY_NAME)
     public String getByTypeAndIdentifierPartitionKey() {
-        return KEY_PATTERN.formatted(TYPE, getIdentifier());
+        return KEY_PATTERN.formatted(Resource.TYPE, getResourceIdentifier());
```
truhacevkir, line 167: Updating keys in order to be able to fetch Resource and Files by single DynamoDB query.

publication-commons/src/main/java/no/unit/nva/publication/service/impl/ResourceService.java
```diff
@@ -269,6 +271,42 @@ public Resource getResourceByIdentifier(SortableIdentifier identifier) throws No
...
+
+        var entries = client.query(queryRequest).getItems().stream()
+                          .map(map -> parseAttributeValuesMap(map, Dao.class))
+                          .toList();
+
+        var resource = extractResource(entries);
+        var files = extractFiles(entries);
+
+        return resource.map(res -> res.copy().withFiles(files).build());
+    }
```
truhacevkir, line 291: Fetching Resource with files by single call. Returning Optional instead of throwing ApiGateway NotFoundException, so we handle specific behavior in handlers/services.

publication-commons/src/main/java/no/unit/nva/publication/model/business/Resource.java
```diff
@@ -225,6 +227,12 @@ public Resource fetch(ResourceService resourceService) throws NotFoundException
         return resourceService.getResourceByIdentifier(this.getIdentifier());
     }
 
+    // TODO: Implementation in this method should be used every place we fetch resource and publication after we have
+    //  migrated files
+    public Optional<Resource> fetchResourceWithFiles(ResourceService resourceService) {
+        return resourceService.getResourceAndFilesByIdentifier(this.getIdentifier());
+    }
+
```
truhacevkir, line 235: Will rename this method to "fetch" once we have migrated files, so fetching Resource will always serve files.

publication-commons/src/main/java/no/unit/nva/publication/model/business/Resource.java
```diff
@@ -77,6 +77,8 @@ public class Resource implements Entity {
     @JsonProperty
     private AssociatedArtifactList associatedArtifacts;
     @JsonProperty
+    private List<File> files;
+    @JsonProperty
```
truhacevkir, line 81: Will add following field to Publication when it is needed.

brinxmat (approved):

PR #2001, opened 2025-01-16
skeleton for delete file handler

truhacevkir:
Skeleton handler + template function + openapi spec for DeleteFileHandler

### DIFF
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -927,6 +927,71 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Problem'
+  /{publicationIdentifier}/file/{fileIdentifier}:
+    delete:
+      summary: Delete file
+      description: Deletes file metadata
+      tags:
+        - external
+      security:
+        - CognitoUserPool: [
+          'https://api.nva.unit.no/scopes/backend',
+          'https://api.nva.unit.no/scopes/frontend',
+          'aws.cognito.signin.user.admin',
+          'https://api.nva.unit.no/scopes/third-party/file-upload'
+        ]
+      parameters:
+        - in: path
+          name: publicationIdentifier
+          description: Identifier of the Publication.
+          required: true
+          schema:
+            type: string
+        - in: path
+          name: fileIdentifier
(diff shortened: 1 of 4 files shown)


docs/openapi.yaml
```diff
@@ -927,6 +927,71 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Problem'
+  /{publicationIdentifier}/file/{fileIdentifier}:
+    delete:
+      summary: Delete file
+      description: Deletes file metadata
```
LarsV123, line 933: Does the endpoint delete the file itself (ultimately from S3) or the metadata we have stored about the file? This can be interpreted as either.

truhacevkir: Deletes metadata only (database entry). The plan is to have event handler which consumes "File deleted event" and removes it from S3.

publication-file/src/test/java/no/unit/nva/publication/file/upload/DeleteFileHandlerTest.java
```diff
@@ -0,0 +1,59 @@
...
+    @BeforeEach
+    void setUp() {
+        super.init();
+        resourceService = getResourceServiceBuilder().build();
+        handler = new DeleteFileHandler(resourceService);
+        output = new ByteArrayOutputStream();
+    }
+
+    @Test
+    void shouldReturnAccepted() throws IOException {
```
brinxmat, line 40: Can you add a test that does the same for a non-existing file? And potentially a non-existing file on a non-existing publication?

LarsV123: Is that the correct/desired behaviour when/if we know that the file/publication doesn't exist? Should it be `404` instead?

truhacevkir: I will implement functionality and add all the tests once I will start with implementation, it is just a test for skeleton code to satisfy jacoco.

brinxmat (approved):

PR #2002, opened 2025-01-16
update file endpoint part one

truhacevkir:
Adding business logic to UpdateFileHandler.
- Adding interface MutableFileMetadata to File and UpdateFileRequest which controls fields that can be updated on file
- Not handling file type yet -> will handle in next pr as it involves som refactoring.


docs/openapi.yaml
```diff
@@ -4793,9 +4789,29 @@ components:
...
+        - identifier
       properties:
+        type:
+          type: string
+          enum:
+            - UpdateFileRequest
         identifier:
           type: string
+        license:
+          type: string
```
brinxmat, line 4803: There is a restriction here, which could be an enum, but we need the enum (and to keep it updated)

brinxmat (approved):

PR #2004, opened 2025-01-17
feature(file-permissions): Add public role #NP-48292

torbjokv:
- Implement access for "everyone else"
- Add tests for degree and embargo

https://sikt.atlassian.net/wiki/spaces/NVAP/pages/3481993334/File+access+control


publication-commons/src/main/java/no/unit/nva/publication/permissions/file/grant/EveryoneGrantStrategy.java
```diff
@@ -16,11 +16,14 @@ public EveryoneGrantStrategy(File file,
...
-            // and download is allowed
-            return file instanceof OpenFile;
+        if (file instanceof OpenFile openFile) {
+            return switch (permission) {
+                case READ_METADATA -> true;
+                case DOWNLOAD -> openFile.fileDoesNotHaveActiveEmbargo();
+                default -> false;
+            };
         }
+
```
truhacevkir, line 26: What about using new **when** condition in switch clause:

```
return switch (file) {
            case OpenFile ignored when permission.equals(FileOperation.READ_METADATA) -> true;
            case OpenFile openFile when permission.equals(FileOperation.DOWNLOAD) -> openFile.fileDoesNotHaveActiveEmbargo();
            default -> false;
        };
```

publication-commons/src/test/java/cucumber/permissions/file/FileAccessReadFeatures.java
```diff
@@ -9,40 +12,64 @@
...
     public void a_file_in_the_state(String string) throws ClassNotFoundException {
-        var clazz = (Class<File>)Class.forName(File.class.getPackageName() + "." + string);
-        var file = File.builder().build(clazz);
+        var file = getFile(string, EMPTY_PROPERTY);
+        scenarioContext.setPublication(randomNonDegreePublication());
+        scenarioContext.setFile(file);
+    }
+
+    @Given("a file of type {string} with property {string}")
+    public void a_file_in_the_state(String fileType, String fileProperty) throws ClassNotFoundException {
```
truhacevkir, line 38: Maybe use camelCase as we do other places.

truhacevkir (approved):

publication-commons/src/main/java/no/unit/nva/publication/permissions/file/grant/EveryoneGrantStrategy.java
torbjokv: pretty cool! but I think in this case it would not grow great as we adapt more complexity and it may become hard to see if we have exhausted all needed possibilities

publication-commons/src/test/java/cucumber/permissions/file/FileAccessReadFeatures.java
torbjokv: yeah, I guess this was the autogenerated style. I'll fix

torbjokv: also renamed the file to 👍 FileAccessFeatures

PR #2005, opened 2025-01-17
delete file event handler

truhacevkir:
Implementing **DeleteFileFromS3EventHandler** which consumes **PublicationService.FileEntry.Delete** and removes S3 object from **nva-resource-storage** bucket.


publication-event-handlers/src/main/java/no/unit/nva/publication/events/bodies/DataEntryUpdateEvent.java
```diff
@@ -113,10 +110,21 @@ public boolean notEmpty() {
...
+            case DoiRequest doiRequest -> DOI_REQUEST_UPDATE_EVENT_TOPIC;
+            case PublishingRequestCase publishingRequestCase -> PUBLISHING_REQUEST_UPDATE_EVENT_TOPIC;
+            case Message message -> MESSAGE_UPDATE_EVENT_TOPIC;
+            case GeneralSupportRequest generalSupportRequest -> GENERAL_SUPPORT_REQUEST_UPDATE_EVENT_TOPIC;
+            case UnpublishRequest unpublishRequest -> UNPUBLISH_REQUEST_UPDATE_EVENT_TOPIC;
+            case FileEntry fileEntry -> this.isDeleteEvent()
+                                            ? FILE_ENTRY_DELETE_EVENT_TOPIC
+                                            : FILE_ENTRY_UPDATE_EVENT_TOPIC;
+            default -> throw new IllegalArgumentException("Unknown entry type: " + type);
+        };
```
truhacevkir, line 124: Refactoring this method and generating two types of events for file -> delete and update event, so there is no need for DeleteFileFromS3EventHandler to consume each update event and check if "newImage" is present.

publication-event-handlers/src/main/java/no/unit/nva/publication/events/handlers/delete/DeleteFileFromS3EventHandler.java
```diff
@@ -0,0 +1,64 @@
...
+import nva.commons.core.JacocoGenerated;
+import nva.commons.core.paths.UnixPath;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import software.amazon.awssdk.services.s3.S3Client;
+
+public class DeleteFileFromS3EventHandler extends DestinationsEventBridgeEventHandler<EventReference, Void> {
+
+    public static final Logger logger = LoggerFactory.getLogger(DeleteFileFromS3EventHandler.class);
+    public static final String FILE_DELETED_MESSAGE =
```
brinxmat, line 22: private

brinxmat (changes requested):

publication-event-handlers/src/test/java/no/unit/nva/publication/events/handlers/delete/DeleteFileFromS3EventHandlerTest.java
```diff
@@ -0,0 +1,89 @@
...
+
+    @BeforeEach
+    public void setUp() {
+        output = new ByteArrayOutputStream();
+        s3Client = new FakeS3Client();
+        handler = new DeleteFileFromS3EventHandler(s3Client);
+    }
+
+    @Test
+    void shouldDeleteFileFromS3WhenFileEntryHasBeenDeleted() throws IOException {
```
LarsV123, line 45: This test doesn't appear to work correctly. If you remove the line `handler.handleRequest(event, output, CONTEXT);`, the test still passes. This indicates to me that the expected file isn't inserted at the expected path, but could be something else.

truhacevkir: Thank you, seems like i forgot to persist file to S3 before deleting it :)

publication-event-handlers/src/main/java/no/unit/nva/publication/events/handlers/delete/DeleteFileFromS3EventHandler.java
```diff
@@ -0,0 +1,62 @@
...
+        var event = getEvent(eventReference);
+        if (event.isDeleteEvent()) {
+            var fileEntry = (FileEntry) event.getOldData();
+            var key = fileEntry.getIdentifier().toString();
+            deleteFile(key);
+            logger.info(FILE_DELETED_MESSAGE, key, fileEntry.getResourceIdentifier());
+        }
+
+        return null;
+    }
```
LarsV123, line 52: Would it make sense to move some of the logic from this handler to a service class, keeping the handler itself as "skinny" as possible? Not necessary to do it now, just a thought.

LarsV123 (approved):

brinxmat (approved):

PR #2010, opened 2025-01-22
Np 48446 update files metadata in update publication handler

truhacevkir:
Taking in use files as own database entries in:
- UpdatePublicationHander
- CreateTicketHandler
- FetchPublicationHandler
- AcceptedPublishingRequestEventHandler

### DIFF
--- a/publication-commons/src/main/java/no/unit/nva/publication/model/business/PublishingRequestCase.java
+++ b/publication-commons/src/main/java/no/unit/nva/publication/model/business/PublishingRequestCase.java
@@ -16,6 +16,7 @@ import com.fasterxml.jackson.annotation.JsonTypeName;
 import java.net.URI;
 import java.time.Instant;
 import java.util.Collections;
+import java.util.List;
 import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
@@ -23,6 +24,8 @@ import no.unit.nva.identifiers.SortableIdentifier;
 import no.unit.nva.model.Publication;
 import no.unit.nva.model.PublicationStatus;
 import no.unit.nva.model.Username;
+import no.unit.nva.model.associatedartifacts.AssociatedArtifact;
+import no.unit.nva.model.associatedartifacts.AssociatedArtifactList;
 import no.unit.nva.model.associatedartifacts.file.File;
 import no.unit.nva.model.associatedartifacts.file.PendingFile;
 import no.unit.nva.publication.exception.InvalidPublicationException;
@@ -33,6 +36,8 @@ import no.unit.nva.publication.service.impl.TicketService;
 import nva.commons.apigateway.exceptions.ApiGatewayException;
 import nva.commons.apigateway.exceptions.ConflictException;
 import nva.commons.core.JacocoGenerated;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 @JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
 @JsonTypeName(PublishingRequestCase.TYPE)
@@ -40,9 +45,11 @@ import nva.commons.core.JacocoGenerated;
 @SuppressWarnings("PMD.GodClass")
 public class PublishingRequestCase extends TicketEntry {
 
+    private static final Logger logger = LoggerFactory.getLogger(PublishingRequestCase.class);
     public static final String RESOURCE_LACKS_REQUIRED_DATA = "Resource does not have required data to be "
                                                               + "published: ";
-
+    private static final String PUBLISHING_FILE_MESSAGE =
+        "Publishing file {} of type {} from approved " + "PublishingRequest {} for publication {}";
     public static final String TYPE = "PublishingRequestCase";
     public static final String MARKED_FOR_DELETION_ERROR =
         "Publication is marked for deletion and cannot be published.";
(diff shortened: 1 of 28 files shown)


publication-commons/src/main/java/no/unit/nva/publication/service/impl/ResourceService.java
```diff
@@ -268,7 +272,15 @@ public Publication getPublication(Publication sampleResource) throws NotFoundExc
...
     public Resource getResourceByIdentifier(SortableIdentifier identifier) throws NotFoundException {
-        return readResourceService.getResourceByIdentifier(identifier);
+        if (shouldUseNewFiles()) {
+            return getResourceAndFilesByIdentifier(identifier).orElseThrow();
+        } else {
+            return readResourceService.getResourceByIdentifier(identifier);
+        }
+    }
+
+    public boolean shouldUseNewFiles() {
```
torbjokv, line 282: Even if this is temporary, I think it would be nice to put this as a standalone FeatureFlags class maybe?

publication-commons/src/main/java/no/unit/nva/publication/model/business/PublishingRequestCase.java
```diff
@@ -283,6 +290,72 @@ public PublishingRequestCase approveFiles() {
...
+        } else {
+            var resource = Resource.resourceQueryObject(getResourceIdentifier()).fetch(resourceService).orElseThrow();
+            var resourceWithUpdatedAssociatedArtifacts = toPublicationWithApprovedFiles(resource).toPublication();
+            resourceService.updatePublication(resourceWithUpdatedAssociatedArtifacts);
+        }
+    }
+
+    public void rejectRejectedFiles(ResourceService resourceService) {
+        if (resourceService.shouldUseNewFiles()) {
+            getFilesForApproval().stream()
```
torbjokv, line 308: maybe we should rename getFilesForApproval to something that does not assume that is should be always approved. getFilesForEvaluation()

publication-commons/src/main/java/no/unit/nva/publication/service/impl/ResourceService.java
```diff
@@ -287,7 +299,15 @@ public Optional<Resource> getResourceAndFilesByIdentifier(SortableIdentifier ide
         var resource = extractResource(entries);
         var files = extractFiles(entries);
 
-        return resource.map(res -> res.copy().withFiles(files).build());
+
+//        resource.ifPresent(res -> res.getAssociatedArtifacts().addAll(files));
```
torbjokv, line 303: should this have a todo label? (or be removed)

publication-commons/src/main/java/no/unit/nva/publication/service/impl/ResourceService.java
```diff
@@ -287,7 +299,15 @@ public Optional<Resource> getResourceAndFilesByIdentifier(SortableIdentifier ide
...
+        resource.ifPresent(r -> {
+            var associatedArtifacts = r.getAssociatedArtifacts();
+            associatedArtifacts.removeIf(File.class::isInstance);
+            associatedArtifacts.addAll(files);
+            r.setAssociatedArtifacts(new AssociatedArtifactList(associatedArtifacts.stream().distinct().toList()));
```
torbjokv, line 308: do you need distinct here? why?

publication-rest/src/main/java/no/unit/nva/publication/update/UpdatePublicationHandler.java
```diff
@@ -305,9 +313,22 @@ private Publication updateMetadata(UpdatePublicationRequest input,
         new PublishingRequestResolver(resourceService, ticketService, userInstance, customer)
             .resolve(existingPublication, publicationUpdate);
 
+        if (resourceService.shouldUseNewFiles()) {
+            Resource.fromPublication(publicationUpdate).getAssociatedArtifacts().stream()
+                .filter(File.class::isInstance)
+                .map(File.class::cast)
+                .forEach(file -> updateFile(existingPublication, file));
```
torbjokv, line 320: Why do you update files here? My assumption was that only publication status would change here? (and that files should keep its statuses on unpublish republish

torbjokv (changes requested):

publication-commons/src/main/java/no/unit/nva/publication/service/impl/ResourceService.java
truhacevkir: What do you mean?

torbjokv: if it can be separated it probably should? Would have the benefit to also mark why its here (its a feature flag). new FeatureFlags(new Environment()).isEnabled("SHOULD_USE_NEW_FILES") or something

publication-commons/src/main/java/no/unit/nva/publication/model/business/PublishingRequestCase.java
```diff
@@ -283,6 +290,72 @@ public PublishingRequestCase approveFiles() {
...
+    private AssociatedArtifact approveFiles(AssociatedArtifact associatedArtifact) {
+        return switch (associatedArtifact) {
+            case PendingFile<?,?> pendingFile when fileIsApproved((File) pendingFile) -> {
+                var approvedFile = pendingFile.approve();
+                logger.info(PUBLISHING_FILE_MESSAGE, pendingFile.getIdentifier(),
+                            pendingFile.getClass().getSimpleName(), getIdentifier(), getResourceIdentifier());
+                yield approvedFile;
+            }
+
+            case null, default -> associatedArtifact;
```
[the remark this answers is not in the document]
truhacevkir: Probably yes, it is a "piece" of old code moved from AcceptedPublishingRequestEventHandler, I will remove it after we have migrated.

publication-rest/src/main/java/no/unit/nva/publication/update/UpdatePublicationHandler.java
truhacevkir: Frontend wants to update files metadata using existing endpoint for updating file, so they do not have to implement massive logic change when taking in use new files.

torbjokv: ah maybe I just looked at the wrong place, I though this was part of unpublish/republish. my bad! all good then :-)

torbjokv (approved):

PR #2013, opened 2025-01-23
removing get publication method

truhacevkir:
Removing use of method `resourceService.getPublication()` in production code. 

Now will all methods that are fetching publication use `resourceService.getResourceByIdentifier()` method which fetches files from database (depending on environment variable).

The plan is to remove use of this method from tests and remove it from resource service, cause underlying logic of this method is the same as in `resourceService.getPublicationByIdentifier()`. (fetching by primary key)


brinxmat (approved):
Nice

PR #2014, opened 2025-01-24
Fixing license and publishing

truhacevkir:
Fixing license, publishing of publication and persisted file.

License:

Following license `https://rightsstatements.org/vocab/InC/1.0/` is being imported like `https://rightsstatements.org/vocab/inc/1.0` because of `formatValidUriLicense()` method in File constructor. Fixing it. 

Publishing:

When publishing publication, we fetch if from database, fetched publication contains persisted files in associatedArtifacts. When persisting updated publication, files end up being stored in both associatedArtifacts and as database entries, which result in duplicated files on publication. Fix -> clean associatedArtifacts from files when updating publication (only using new files). Should consider more elegant way of handling it. 

File:

Persisting file with customer configured RRS by default + returning PendingOpenFile instead of UploadedFile.

### DIFF
--- a/publication-model/src/main/java/no/unit/nva/model/associatedartifacts/file/File.java
+++ b/publication-model/src/main/java/no/unit/nva/model/associatedartifacts/file/File.java
@@ -9,6 +9,7 @@ import com.fasterxml.jackson.annotation.JsonTypeInfo;
 import com.fasterxml.jackson.annotation.JsonTypeInfo.Id;
 import java.net.URI;
 import java.time.Instant;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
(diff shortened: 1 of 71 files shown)


truhacevkir:
Will remove UploadedFile from datamodel when we have migrated if there is no need for it. Frontend does not need it, and I think in backend we can handle "uploaded" file by making a check if oldImage of consumed database event is empty.

publication-model/src/main/java/no/unit/nva/model/associatedartifacts/file/File.java
```diff
@@ -48,24 +44,12 @@ public abstract class File implements JsonSerializable, AssociatedArtifact {
...
-    public static final Map<String, URI> LICENSE_MAP = Map.of("CC BY",
-                                                              URI.create("https://creativecommons.org/licenses/by/4.0"),
-                                                              "CC BY-NC", URI.create(
-            "https://creativecommons.org/licenses/by-nc/4.0"), "CC BY-NC-ND", URI.create(
-            "https://creativecommons.org/licenses/by-nc-nd/4.0"), "CC BY-NC-SA", URI.create(
-            "https://creativecommons.org/licenses/by-nc-sa/4.0"), "CC BY-ND", URI.create(
-            "https://creativecommons.org/licenses/by-nd/4.0"), "CC BY-SA", URI.create(
-            "https://creativecommons.org/licenses/by-sa/4.0"), "CC0", URI.create(
-            "https://creativecommons.org/publicdomain/zero/1.0"), "RightsReserved",
-                                                              URI.create("http://rightsstatements.org/vocab/InC/1.0/"));
```
brinxmat, line 60: I'm tempted to say that there is a validation in the DTO, but we should define what is an acceptable license and (not least) how to inform the code of the list of potential URIs (per institution? How does that work for a shared record?) This probably requires a task in Jira

brinxmat: An no, we shouldn't re-write URIs…just reject the bad ones

brinxmat (approved):

PR #2015, opened 2025-01-27
Persisiting log entries on file changes

truhacevkir:
Persisting Log entries for file for following events:

- File uploaded
- File approved
- File rejected

Using the same pattern as on for resource, we persist FileEvent on file updates which is consumed by PersistLogEntryEventHandler.

Next step will be to create log event on relevant file type update + when file is being deleted. But there is some cases we need to discuss before implementing.

### DIFF
--- a/publication-commons/src/main/java/no/unit/nva/publication/model/business/FileEntry.java
+++ b/publication-commons/src/main/java/no/unit/nva/publication/model/business/FileEntry.java
@@ -1,5 +1,6 @@
 package no.unit.nva.publication.model.business;
 
+import static java.util.Objects.nonNull;
 import static nva.commons.core.attempt.Try.attempt;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonIgnore;
@@ -15,6 +16,10 @@ import no.unit.nva.identifiers.SortableIdentifier;
 import no.unit.nva.model.Publication;
 import no.unit.nva.model.associatedartifacts.file.File;
 import no.unit.nva.model.associatedartifacts.file.PendingFile;
+import no.unit.nva.publication.model.business.publicationstate.FileApprovedEvent;
+import no.unit.nva.publication.model.business.publicationstate.FileEvent;
+import no.unit.nva.publication.model.business.publicationstate.FileRejectedEvent;
+import no.unit.nva.publication.model.business.publicationstate.FileUploadedEvent;
 import no.unit.nva.publication.model.storage.Dao;
 import no.unit.nva.publication.model.storage.FileDao;
 import no.unit.nva.publication.service.impl.ResourceService;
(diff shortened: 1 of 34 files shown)


publication-commons/src/main/java/no/unit/nva/publication/model/business/FileEntry.java
```diff
@@ -33,39 +39,45 @@ public final class FileEntry implements Entity {
...
     private File file;
+    private FileEvent fileEvent;
 
     @JsonCreator
     private FileEntry(@JsonProperty("resourceIdentifier") SortableIdentifier resourceIdentifier,
                       @JsonProperty("createdDate") Instant createdDate,
                       @JsonProperty("modifiedDate") Instant modifiedDate, @JsonProperty("owner") User owner,
                       @JsonProperty("ownerAffiliation") URI ownerAffiliation,
-                      @JsonProperty("customerId") URI customerId, @JsonProperty("file") File file) {
+                      @JsonProperty("customerId") URI customerId, @JsonProperty("file") File file,
```
brinxmat, line 49: We should probably not have called it "customerId" since "customer" would be better. Oh well.

publication-commons/src/main/java/no/unit/nva/publication/model/business/logentry/FileLogEntry.java
```diff
@@ -0,0 +1,76 @@
+package no.unit.nva.publication.model.business.logentry;
+
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import java.time.Instant;
+import no.unit.nva.identifiers.SortableIdentifier;
+import no.unit.nva.publication.service.impl.ResourceService;
+
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
+public record FileLogEntry(SortableIdentifier identifier, SortableIdentifier fileIdentifier,
```
brinxmat, line 9: This signature is very easy to mess up when using; I'd be tempted to force use of the builder

publication-commons/src/main/java/no/unit/nva/publication/model/business/logentry/PublicationLogEntry.java
```diff
@@ -0,0 +1,63 @@
...
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import java.time.Instant;
+import no.unit.nva.identifiers.SortableIdentifier;
+import no.unit.nva.publication.service.impl.ResourceService;
+
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
+public record PublicationLogEntry(SortableIdentifier identifier, SortableIdentifier resourceIdentifier, LogTopic topic,
+                                  Instant timestamp, LogUser performedBy) implements LogEntry {
+
+    public static final String TYPE = "PublicationLogEntry";
```
brinxmat, line 12: Public?

brinxmat (approved):
See comments

publication-commons/src/main/java/no/unit/nva/publication/model/business/logentry/PublicationLogEntry.java
truhacevkir: Yes, it used in LogEntry interface as subtype property.

PR #2018, opened 2025-01-29
debugging file migration

truhacevkir:


truhacevkir:
Adding log message in the case migrations fails (as it did in dev)

publication-commons/src/main/java/no/unit/nva/publication/service/impl/ResourceService.java
```diff
@@ -394,12 +394,17 @@ private static boolean isResourceWithFiles(Entity dataEntry) {
...
 
     private void persistFileEntry(Resource resource, File file) {
-        var userInstance = UserInstance.fromPublication(resource.toPublication());
-        var resourceIdentifier = resource.getIdentifier();
-        var existingFile = FileEntry.queryObject(file.getIdentifier(), resourceIdentifier)
-                               .fetch(this);
-        if (existingFile.isEmpty()) {
-            FileEntry.create(file, resourceIdentifier, userInstance).persist(this);
+        try {
+            var userInstance = UserInstance.fromPublication(resource.toPublication());
```
brinxmat, line 398: I am going to ask — what do we imagine this Util does *today*?

brinxmat (approved):

publication-commons/src/main/java/no/unit/nva/publication/service/impl/ResourceService.java
truhacevkir: What do you mean? :)

PR #2021, opened 2025-01-31
crafting log messages for migrated files

truhacevkir:
Use case: When migrating files, we create new database entry. By default we are persisting "file uploaded" log entry when it happens. It is wrong for migrated files, as they will get uploaded date which is "migration" date.

Solution: Crafting FileApproved and FileUploaded log messages when migrating depending on file type and which dates file has, as after a little research it seems that "old" files in production are missing publishedDate.

### DIFF
--- a/publication-commons/src/test/java/no/unit/nva/publication/service/impl/ResourceServiceTest.java
+++ b/publication-commons/src/test/java/no/unit/nva/publication/service/impl/ResourceServiceTest.java
@@ -77,6 +77,7 @@ import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 import java.util.stream.Stream;
+import no.unit.nva.commons.json.JsonUtils;
 import no.unit.nva.identifiers.SortableIdentifier;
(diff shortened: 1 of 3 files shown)


publication-commons/src/test/java/no/unit/nva/publication/service/impl/ResourceServiceTest.java
```diff
@@ -1736,6 +1740,153 @@ void shouldLogExceptionWhenFailingMigratingFiles() throws BadRequestException {
                      () -> resourceService.refreshResources(List.of(Resource.fromPublication(persistedPublication))));
     }
 
+    @Deprecated
+    @Test
+    void shouldCreateFileEventWithFilePublishingDateWhenMigratingOpenFileWithPublishedDate()
+        throws BadRequestException, JsonProcessingException {
+        var json = """
```
truhacevkir, line 1747: Using json as it is impossible to construct "published" file without publishedDate, and it is probably good thing :)

brinxmat (approved):

PR #2024, opened 2025-02-03
fixing file dao identifier

truhacevkir:
FileDao deserializing issues

### DIFF
--- a/publication-commons/src/main/java/no/unit/nva/publication/model/storage/FileDao.java
+++ b/publication-commons/src/main/java/no/unit/nva/publication/model/storage/FileDao.java
@@ -113,10 +113,9 @@ public final class FileDao extends Dao implements DynamoEntryByIdentifier {
         return getData().getCustomerId();
     }
 
-    @JsonProperty("identifier")
     @Override
     public SortableIdentifier getIdentifier() {
-        return getData().getIdentifier();
+        return identifier;
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


publication-commons/src/main/java/no/unit/nva/publication/model/storage/FileDao.java
```diff
@@ -115,7 +116,7 @@ public URI getCustomerId() {
     @JsonProperty("identifier")
     @Override
     public SortableIdentifier getIdentifier() {
-        return identifier;
+        return getData().getIdentifier();
```
truhacevkir, line 119: This makes identifier be a part of row, found out today that it was not. 
image

brinxmat: Out of interest, is the field in use at all now?

truhacevkir: It is indirectly in use, cause in transactions we use primaryKey() method. 
```
    @Override
    @JsonProperty(PRIMARY_KEY_SORT_KEY_NAME)
    public String getPrimaryKeySortKey() {
        return KEY_PATTERN.formatted(TYPE, getIdentifier());
    }
```

It seems that that json annotations fixes it, but everything works fine when identifier on dao is null. Since all dao's have identifier field in database, I think it is great to have it here as well.

brinxmat (approved):

PR #2029, opened 2025-02-07
removing customer config validation for external client on upload

truhacevkir:
Not validating allowed publication types for file upload when external client uploads file.


brinxmat (approved):
Have a feeling this should be configurable, but OK

PR #2030, opened 2025-02-09
modifying update and creating query object

truhacevkir:
- Updating update method in resourceService to update files in the same transaction as resource
- Returning QueryObject when creating query object for FileEntry.

### DIFF
--- a/publication-commons/src/main/java/no/unit/nva/publication/model/business/QueryObject.java
+++ b/publication-commons/src/main/java/no/unit/nva/publication/model/business/QueryObject.java
@@ -1,0 +1,11 @@
+package no.unit.nva.publication.model.business;
+
+import java.util.Optional;
+import no.unit.nva.publication.service.impl.ResourceService;
+
+public interface QueryObject<T> {
+
+    Optional<T> fetch(ResourceService resourceService);
+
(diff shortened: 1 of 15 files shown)


publication-rest/src/main/java/no/unit/nva/publication/update/UpdatePublicationHandler.java
truhacevkir, line 1: moving file update from here to resourceService

publication-commons/src/main/java/no/unit/nva/publication/service/impl/ResourceService.java
truhacevkir, line 1: Moving getResourceAndFilesByIdentifier() method to readResourceService to make it possible to reuse method in updateResourceService and renaming method to getResourceByIdentifier().

publication-commons/src/main/java/no/unit/nva/publication/model/business/QueryObject.java
truhacevkir, line 1: Adding QueryObject.

torbjokv (approved):
nice! we should probably improve our function names, but no suggestions at the moment

PR #2036, opened 2025-02-10
log messages when importing files

truhacevkir:
- Persisting log entry (setting log event) when importing files from Brage and Scopus.


publication-commons/src/main/java/no/unit/nva/publication/model/business/logentry/FileLogEntry.java
```diff
@@ -31,6 +32,7 @@ public static final class Builder {
         private String fileType;
         private Instant timestamp;
         private LogUser performedBy;
+        private ImportSource importSource;
```
torbjokv, line 35: is it tricky to reuse LogUser? It seems a bit off that all filelogEntries should have importSource

PR #2038, opened 2025-02-11
persisting new files when updating in brage import

truhacevkir:
When merging incoming publication with existing one
And incoming has new file
Persisting that file

### DIFF
--- a/brage-import/src/main/java/no/sikt/nva/brage/migration/lambda/BrageEntryEventConsumer.java
+++ b/brage-import/src/main/java/no/sikt/nva/brage/migration/lambda/BrageEntryEventConsumer.java
@@ -40,7 +40,9 @@ import no.unit.nva.model.exceptions.InvalidIssnException;
 import no.unit.nva.model.exceptions.InvalidUnconfirmedSeriesException;
 import no.unit.nva.model.instancetypes.PublicationInstance;
 import no.unit.nva.publication.external.services.UriRetriever;
+import no.unit.nva.publication.model.business.FileEntry;
 import no.unit.nva.publication.model.business.Resource;
+import no.unit.nva.publication.model.business.UserInstance;
 import no.unit.nva.publication.s3imports.ImportResult;
 import no.unit.nva.publication.service.impl.ResourceService;
 import no.unit.nva.s3.S3Driver;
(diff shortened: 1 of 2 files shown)


brage-import/src/test/java/no/sikt/nva/brage/migration/lambda/BrageEntryEventConsumerTest.java
```diff
@@ -1177,6 +1178,60 @@ void shouldMigrateHiddenFileFromIncomingWhenExistingPublicationDoesNotHaveHidden
                        .anyMatch(HiddenFile.class::isInstance));
     }
 
+    @Test
+    void shouldPersistNewFileWithUserInstanceFromIncomingPublicationWhenMergingPublications()
+        throws IOException, InvalidUnconfirmedSeriesException {
```
truhacevkir, line 1183: This huge test is disgusting. But it is the only to merge two publication via handler and nothing can be extracted.

brage-import/src/main/java/no/sikt/nva/brage/migration/lambda/BrageEntryEventConsumer.java
```diff
@@ -279,8 +281,15 @@ private UriWrapper discardedFilesReportUri(Publication publication, S3Event s3Ev
...
                                                                      PublicationRepresentation representation) {
+
         var importSource = createImportSource(representation);
         var resource = Resource.fromPublication(publicationForUpdate);
+        var updatedFiles = resource.getFiles();
+        var oldFiles = Resource.fromPublication(existinPublication).getFiles();
+        updatedFiles.stream()
+            .filter(file -> oldFiles.stream().noneMatch(f -> f.getIdentifier().equals(file.getIdentifier())))
+            .forEach(file -> FileEntry.create(file, resource.getIdentifier(),
+                                              UserInstance.fromPublication(representation.publication())).persist(resourceService));
```
truhacevkir, line 292: I think it should be done here, as we want fileentry with userinstance of publication we will not keep.

brinxmat (approved):

brage-import/src/main/java/no/sikt/nva/brage/migration/lambda/BrageEntryEventConsumer.java
```diff
@@ -279,13 +284,24 @@ private UriWrapper discardedFilesReportUri(Publication publication, S3Event s3Ev
...
+        updatedFiles.stream()
+            .filter(file -> containsNotFile(file, oldFiles))
+            .forEach(file -> FileEntry.importFileEntry(file, resource.getIdentifier(),
+                                              UserInstance.fromPublication(representation.publication()), importSource).persist(resourceService));
         resource.updateResourceFromImport(resourceService, importSource);
         var newImage = resource.fetch(resourceService).orElseThrow().toPublication();
         return new BrageMergingReport(existinPublication, newImage);
     }
 
+    private static boolean containsNotFile(File file, List<UUID> fileList) {
```
brinxmat, line 301: how about removing the "!" and wrapping the call in Predicate::not?

truhacevkir: Nice, have never seen it

PR #2040, opened 2025-02-12
Np 48476 new files in scopus import

truhacevkir:
Existing flow:
```
When importing import candidate
And import candidate exists
And import candidate is not imported
Then overwrite candidate.
```

When updating import candidate, we in practice overwriting it, but in case there are files on existing and incoming candidate, we should replace them. They will never have identical identifiers as identifiers are set when mapping xml to candidate.

### DIFF
--- a/publication-commons/src/main/java/no/unit/nva/publication/service/impl/UpdateResourceService.java
+++ b/publication-commons/src/main/java/no/unit/nva/publication/service/impl/UpdateResourceService.java
@@ -206,27 +206,30 @@ public class UpdateResourceService extends ServiceWithTransactions {
                     CuratingInstitutionsUtil.getCuratingInstitutionsOnline(importCandidate, uriRetriever));
             }
 
-            var fileEntriesToDelete = Resource.fromImportCandidate(existingImportCandidate).getFileEntries().stream()
-                                          .map(FileEntry::toDao)
-                                          .map(dao -> dao.toDeleteTransactionItem(tableName))
-                                          .toList();
-
-            var fileEntriesToPersist = Resource.fromImportCandidate(importCandidate).getFileEntries().stream()
-                                           .map(FileEntry::toDao)
-                                           .map(fileDao -> fileDao.toPutNewTransactionItem(tableName))
-                                           .toList();
-
             var transactions = new ArrayList<TransactWriteItem>();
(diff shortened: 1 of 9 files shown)


publication-commons/src/main/java/no/unit/nva/publication/service/impl/UpdateResourceService.java
```diff
@@ -206,9 +206,23 @@ public ImportCandidate updateImportCandidate(ImportCandidate importCandidate) th
...
+
+            var transactions = new ArrayList<TransactWriteItem>();
             var resource = Resource.fromImportCandidate(importCandidate);
-            var updateResourceTransactionItem = createPutTransaction(resource);
-            var request = new TransactWriteItemsRequest().withTransactItems(List.of(updateResourceTransactionItem));
+            transactions.add(createPutTransaction(resource));
+            transactions.addAll(fileEntriesToDelete);
+            transactions.addAll(fileEntriesToPersist);
+
+            var request = new TransactWriteItemsRequest().withTransactItems(transactions);
```
brinxmat, line 225: Chatty method. Extract what can be extracted

scopus-import/src/main/java/no/sikt/nva/scopus/update/ScopusUpdater.java
```diff
@@ -41,10 +41,12 @@ public ImportCandidate updateImportCandidate(ImportCandidate importCandidate)
...
-            var persistedImportcandidate = resourceService.getImportCandidateByIdentifier(
-                existingImportCandidate.getIdentifier());
-            persistedImportcandidate.setEntityDescription(importCandidate.getEntityDescription());
-            return persistedImportcandidate;
+            var persistedImportcandidate = resourceService.getImportCandidateByIdentifier(existingImportCandidate.getIdentifier());
+            return persistedImportcandidate.copyImportCandidate()
+                       .withEntityDescription(importCandidate.getEntityDescription())
+                       .withAssociatedArtifacts(importCandidate.getAssociatedArtifacts())
+                       .withAdditionalIdentifiers(importCandidate.getAdditionalIdentifiers())
+                       .build();
```
brinxmat, line 49: Extract

brinxmat (changes requested):

torbjokv (approved):

PR #2051, opened 2025-02-14
adding alias to complete upload request list parts

truhacevkir:
Adding json aliases for `ETag` and `PartNumber` to `CompleteUploadRequestBody`.
Armaz does not use third party library and sends forward values from `/prepare` call, and these values should match, i.e. we have to support both `ETag`, `etag` and `PartNumber` and `partNumber`.


brinxmat (commented):
Weird

brinxmat (approved):

PR #2064, opened 2025-02-20
removing publish publication method from resource service

truhacevkir:
Remove `publishPublication()` method from `resourceService` which has been used in tests only.


publication-commons/src/main/java/no/unit/nva/publication/model/PublishPublicationStatusResponse.java
truhacevkir, line 1: This class has never been used in production code as far as I remember (maybe before tickets came?)

brinxmat (approved):

PR #2068, opened 2025-02-24
commenting out file fetch from xml

truhacevkir:
Commenting out the functionality that fetches files based on URLs from Scopus XML. The reason is that Scopus XML only provides download URLs without associated version and license information. We do not want to import files when we are unsure if they can be published.

For files fetched via DOI, if the license is missing, we import them as internal files without a license. This allows the importer to decide whether the file should be published or not.

### DIFF
--- a/brage-import/src/main/resources/customers.json
+++ b/brage-import/src/main/resources/customers.json
@@ -436,7 +436,7 @@
       "test": "d102baef-5861-43e2-bb21-42a1f6479aef",
       "dev": "ff6fd53f-30ff-4c47-955e-f201fe5d9e0c"
     },
-    "shortName": "TOI"
+    "shortName": "TØI"
   },
   {
     "name": "kristiania",
--- a/brage-import/src/main/resources/customers.json
+++ b/brage-import/src/main/resources/customers.json
@@ -436,7 +436,7 @@
       "test": "d102baef-5861-43e2-bb21-42a1f6479aef",
       "dev": "ff6fd53f-30ff-4c47-955e-f201fe5d9e0c"
     },
-    "shortName": "T\u00d8I"
+    "shortName": "TOI"
   },
   {
     "name": "kristiania",
--- a/scopus-import/src/main/java/no/sikt/nva/scopus/conversion/files/ScopusFileConverter.java
+++ b/scopus-import/src/main/java/no/sikt/nva/scopus/conversion/files/ScopusFileConverter.java
@@ -27,16 +27,10 @@ import java.time.LocalDate;
 import java.time.ZoneId;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.Objects;
 import java.util.Optional;
 import java.util.TreeSet;
-import java.util.UUID;
 import java.util.stream.Collector;
 import no.scopus.generated.DocTp;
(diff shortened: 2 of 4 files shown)


scopus-import/src/main/java/no/sikt/nva/scopus/conversion/files/ScopusFileConverter.java
```diff
@@ -458,21 +360,94 @@ private ScopusFile saveFile(ScopusFile scopusFile) {
...
-                   .map(UpwOpenAccessType::getUpwOaLocations)
-                   .map(UpwOaLocationsType::getUpwOaLocation)
-                   .orElse(List.of());
-    }
+    // The functionality to fetch files from XML is currently disabled because the XML data lacks
+    // license information for the files. This code will remain commented out until a decision
+    // is made regarding how to handle files without license information.
+    // Once a solution is implemented, this code can be revisited and potentially re-enabled.
+
+    //    private Optional<List<AssociatedArtifact>> fetchFilesFromXml(DocTp docTp) {
```
truhacevkir, line 368: Do not like commenting it out that way, but I think this is something we very likely will take in use once PO has decided what files we want to import from xml.

brage-import/src/main/resources/customers.json
```diff
@@ -436,7 +436,7 @@
       "test": "d102baef-5861-43e2-bb21-42a1f6479aef",
       "dev": "ff6fd53f-30ff-4c47-955e-f201fe5d9e0c"
     },
-    "shortName": "T\u00d8I"
+    "shortName": "TOI"
```
truhacevkir, line 439: I think this is right :)

brinxmat: TØI

scopus-import/src/main/java/no/sikt/nva/scopus/conversion/files/ScopusFileConverter.java
```diff
@@ -162,17 +135,16 @@ private static String extractFileExtensionFromContentType(String value) {
...
+                   .map(Paths::get)
+                   .map(Path::getFileName)
+                   .map(Path::toString);
     }
 
     private static String extractFileNameFromContentDisposition(String contentType) {
-        return contentType.split(FILENAME_CONTENT_TYPE_HEADER_VALUE)[1]
-                   .split(CONTENT_TYPE_DELIMITER)[0]
-                   .replace(QUOTE, StringUtils.EMPTY_STRING);
+        return contentType.split(FILENAME_CONTENT_TYPE_HEADER_VALUE)[1].split(CONTENT_TYPE_DELIMITER)[0].replace(QUOTE,
```
brinxmat, line 146: You probably want to change this — even though it was just reformatting — since the splits may result in out-of-bounds issues

scopus-import/src/main/java/no/sikt/nva/scopus/conversion/files/ScopusFileConverter.java
brinxmat: You can create a PR to reinstate it, but remove it here

brinxmat (approved):

PR #2070, opened 2025-02-24
adding log messages when fetching publication channel

truhacevkir:
We should find out why all channels from Scopus are not being fetched from channel register. First step is to find out if channel register is failing or we receive multiple results when searching for channel by channel name/issn/isbn. 

Wednesday will show what we are struggling with :)

### DIFF
--- a/scopus-import/src/main/java/no/sikt/nva/scopus/conversion/PublicationChannelConnection.java
+++ b/scopus-import/src/main/java/no/sikt/nva/scopus/conversion/PublicationChannelConnection.java
@@ -1,17 +1,21 @@
 package no.sikt.nva.scopus.conversion;
 
+import static java.net.HttpURLConnection.HTTP_OK;
 import static no.sikt.nva.scopus.conversion.PublicationChannel.PUBLISHER;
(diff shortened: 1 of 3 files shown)


scopus-import/src/main/java/no/sikt/nva/scopus/conversion/PublicationChannelConnection.java
```diff
@@ -53,16 +58,25 @@ private static PublicationChannelResponse toPublicationChannelResponse(String re
...
                    .filter(Objects::nonNull)
                    .filter(PublicationChannelConnection::containsSingleResult)
                    .map(PublicationChannelConnection::getId)
                    .findFirst();
     }
 
+    private String handleResponse(HttpResponse<String> response) {
+        if (HTTP_OK != response.statusCode()) {
+            logger.error("Publication channels API responded with {} when searching for {}", response.statusCode(),
+                         response.request().uri());
```
brinxmat, line 75: I hope that the escaping here should is very robust before we request

truhacevkir: UriWrapper is handling escaping yes.

PR #2073, opened 2025-02-25
creating doi with the same owner as pub req owner

truhacevkir:
Next step in bug fixing.

UserInstance from ticket is missing mandatory fields for ticket, fixing it. 
We need in future to remove all possible methods that craft tickets without mandatory fields and create validation on ticket creation.

### DIFF
--- a/publication-commons/src/main/java/no/unit/nva/publication/model/business/UserInstance.java
+++ b/publication-commons/src/main/java/no/unit/nva/publication/model/business/UserInstance.java
@@ -108,7 +108,8 @@ public class UserInstance implements JsonSerializable {
     }
 
     public static UserInstance fromTicket(TicketEntry ticket) {
-        return UserInstance.create(ticket.getOwner(), ticket.getCustomerId());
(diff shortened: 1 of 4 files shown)


publication-commons/src/main/java/no/unit/nva/publication/model/business/UserInstance.java
```diff
@@ -108,7 +108,8 @@ public static UserInstance fromMessage(Message message) {
     }
 
     public static UserInstance fromTicket(TicketEntry ticket) {
-        return UserInstance.create(ticket.getOwner(), ticket.getCustomerId());
+        return new UserInstance(ticket.getOwner().toString(), ticket.getCustomerId(), ticket.getOwnerAffiliation(),
+                                ticket.getResponsibilityArea(), null, List.of(), UserClientType.INTERNAL);
```
truhacevkir, line 112: UserInstance created from ticket is missing important fields such as owner affiliation and responsibility area.

torbjokv (approved):

PR #2076, opened 2025-02-26
removing methods that craft doi requests in tests

truhacevkir:
Removing old methods for crafting DoiRequest so we never use them.

Should use only following method:

`public static DoiRequest create(Resource resource, UserInstance userInstance)`

Similar methods should be removed from PublishingRequest and GeneralSupportCase as well.


publication-commons/src/main/java/no/unit/nva/publication/model/business/DoiRequest.java
```diff
@@ -278,16 +269,10 @@ public PublicationStatus getResourceStatus() {
...
-    public void setResourceStatus(PublicationStatus resourceStatus) {
+    private void setResourceStatus(PublicationStatus resourceStatus) {
         this.resourceStatus = resourceStatus;
     }
 
-    public void validate() {
-        attempt(this::getResourceIdentifier)
-            .toOptional()
-            .orElseThrow(() -> new IllegalStateException(TICKET_WITHOUT_REFERENCE_TO_PUBLICATION_ERROR));
-    }
```
truhacevkir, line 289: I dont think this validation is any good. 
We want to validate that all required fields are set on DoiRequest, but for now it is not possible, cause a lot of tests still use ticket crafting method from TicketEntry which creates DoiRequests that in production is not valid DoiRequest.

I think we want validation like this on ticket creation, but for now it requires even more refactoring:

```
    private void validate() {
        Objects.requireNonNull(getResponsibilityArea());
        Objects.requireNonNull(getOwnerAffiliation());
        Objects.requireNonNull(getOwner());
        Objects.requireNonNull(getCustomerId());
        Objects.requireNonNull(getResourceIdentifier());
        Objects.requireNonNull(getResourceStatus());
        Objects.requireNonNull(getStatus());
    }
```

brinxmat (approved):

PR #2077, opened 2025-02-26
redirect when unpublished publication

truhacevkir:


publication-rest/src/main/java/no/unit/nva/publication/fetch/FetchPublicationHandler.java
```diff
@@ -117,10 +117,14 @@ protected String processInput(Void input, RequestInfo requestInfo, Context conte
...
 
     private String producePublicationResponseWhenUnpublished(RequestInfo requestInfo, Resource resource)
-        throws GoneException {
-        return userCanUpdateResource(requestInfo, resource)
-                   ? createPublicationResponse(requestInfo, resource)
-                   : produceRemovedPublicationResponse(resource, requestInfo);
+        throws GoneException, UnsupportedAcceptHeaderException {
+        return shouldRedirectToDuplicate(requestInfo, resource) || !userCanUpdateResource(requestInfo, resource)
+                   ? produceRemovedPublicationResponse(resource, requestInfo)
+                   : producePublicationResponse(requestInfo, resource);
```
truhacevkir, line 123: 1. We should redirect when user is asking for redirect.
2. We should return tombstone without files when user does not want to redirect or user has no righta to see files for unpublished resource.
3. We should serve unpublished resource with files when user has rights and does not want to redirect.

publication-commons/src/main/java/no/unit/nva/publication/model/business/Resource.java
```diff
@@ -200,6 +200,8 @@ public Resource delete(UserInstance userInstance, Instant currentTime) {
                    .withCreatedDate(getCreatedDate())
                    .withPublishedDate(getPublishedDate())
                    .withModifiedDate(currentTime)
+                   .withDuplicateOf(getDuplicateOf())
+                   .withPublicationNotes(getPublicationNotes())
                    .withResourceEvent(DeletedResourceEvent.create(userInstance, currentTime))
```
truhacevkir, line 205: Keeping these two fields when deleting unpublished publication

torbjokv (approved):

PR #2078, opened 2025-02-26
removing fromPublication method from GSR

truhacevkir:
Removing `GeneralSupportRequest.fromPublication(publication)` method. Should always use `create` method.


brinxmat (approved):
Lovely!

PR #2081, opened 2025-02-28
removing null artifact automatically when adding file or link

truhacevkir:
There is a bug in all environments that occurs when a publication contains a NullAssociatedArtifact and an attempt is made to update it with a file. During file upload, the list of associated artifacts is not being mutated. Consequently, if the associatedArtifacts list already contains a NullAssociatedArtifact, fetching the publication from the database fails due to serialization errors with the artifacts.

### DIFF
--- a/publication-model/src/main/java/no/unit/nva/model/associatedartifacts/AssociatedArtifactList.java
+++ b/publication-model/src/main/java/no/unit/nva/model/associatedartifacts/AssociatedArtifactList.java
@@ -1,6 +1,8 @@
 package no.unit.nva.model.associatedartifacts;
 
+import static java.util.Objects.isNull;
 import static java.util.Objects.nonNull;
+import static java.util.function.Predicate.not;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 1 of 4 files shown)


publication-model/src/main/java/no/unit/nva/model/associatedartifacts/AssociatedArtifactList.java
```diff
@@ -13,14 +15,25 @@
...
-    public static final String NULL_OBJECT_MUST_BE_SINGLETON_IN_LIST = "AssociatedArtifactLists containing "
-            + "NullAssociatedArtifact must contain only this element as a singleton";
     private final List<AssociatedArtifact> associatedArtifacts;
     
     @JsonCreator
     public AssociatedArtifactList(List<AssociatedArtifact> artifacts) {
-        throwExceptionIfNullObjectIsNotOnlyElementInList(artifacts);
-        this.associatedArtifacts = nonNull(artifacts) ? artifacts : Collections.emptyList();
+        if (isNull(artifacts)) {
+            this.associatedArtifacts = Collections.emptyList();
```
brinxmat, line 23: Remove this and add the check to the getter

brinxmat: Or do we want the empty list in ddb?

publication-model/src/test/java/no/unit/nva/model/associatedartifacts/AssociatedArtifactListTest.java
```diff
@@ -0,0 +1,21 @@
...
+import static no.unit.nva.model.testing.associatedartifacts.AssociatedArtifactsGenerator.randomOpenFile;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import java.util.List;
+import org.junit.jupiter.api.Test;
+
+class AssociatedArtifactListTest {
+
+    @Test
+    void shouldRemoveNullAssociatedArtifactFromListWhenListContainsNotOnlyNullAssociatedArtifact() {
```
brinxmat, line 12: What happens if you have two?

brinxmat (changes requested):

publication-model/src/main/java/no/unit/nva/model/associatedartifacts/AssociatedArtifactList.java
truhacevkir: It was empty list before also. We do not have getter here, we are in constructor.  Or what do you think?
```
    public AssociatedArtifactList(List<AssociatedArtifact> artifacts) {
        throwExceptionIfNullObjectIsNotOnlyElementInList(artifacts);
        this.associatedArtifacts = nonNull(artifacts) ? artifacts : Collections.emptyList();
   }
```

publication-model/src/test/java/no/unit/nva/model/associatedartifacts/AssociatedArtifactListTest.java
truhacevkir: Both will be removed. Or do we want to handle case:

When there is only NullAssociatedArtifacts in list, remove all keeping only one artifacts?

publication-model/src/main/java/no/unit/nva/model/associatedartifacts/AssociatedArtifactList.java
brinxmat: For me, I don't think. persisting an empty or null value makes sense if we can deserialize it to an empty list. by adding it to a getter

publication-model/src/test/java/no/unit/nva/model/associatedartifacts/AssociatedArtifactListTest.java
brinxmat: I think it makes sense to say n+ > one

brinxmat (approved):

PR #2082, opened 2025-02-28
fix(file-permissions): also include owner-curator in addition to cont…

torbjokv:
…ributor-curator when evaluating if user is curator or not #NP-48683

### DIFF
--- a/publication-commons/src/main/java/no/unit/nva/publication/permissions/file/FileStrategyBase.java
+++ b/publication-commons/src/main/java/no/unit/nva/publication/permissions/file/FileStrategyBase.java
@@ -53,7 +53,7 @@ public class FileStrategyBase {
         return file.getOwnerAffiliation().equals(userTopLevelOrg);
     }
 
-    protected boolean hasTopLevelRelationForCurrentResource() {
+    private boolean hasTopLevelRelationForResourceContributors() {
         var userTopLevelOrg = userInstance.getTopLevelOrgCristinId();
 
         logger.info("checking if resource top level affiliation {} for user {} is equal to {}.",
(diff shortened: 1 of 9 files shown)


publication-commons/src/main/java/no/unit/nva/publication/permissions/file/FileStrategyBase.java
```diff
@@ -65,8 +65,16 @@ protected boolean hasTopLevelRelationForCurrentResource() {
         return resource.getCuratingInstitutions().stream().anyMatch(org -> org.id().equals(userTopLevelOrg));
     }
 
+    private boolean hasCuratorRelation() {
+        return hasTopLevelRelationForResourceContributors() || hasTopLevelRelationForResourceOwner();
```
torbjokv, line 69: this is the fix, other stuff is just a try to make tests more explicit

publication-commons/src/test/resources/features/file/file_access_write.feature
```diff
@@ -8,53 +9,60 @@ Feature: File metadata write and file delete permissions
...
+      | PendingOpenFile     | File curators for other contributors   | Not Allowed |
+      | PendingOpenFile     | File curator by publication owner at X | Allowed     |
+      | PendingOpenFile     | Everyone else                          | Not Allowed |
+      | PendingOpenFile     | External client                        | Not Allowed |
+      | PendingInternalFile | Publication owner at X                 | Allowed     |
+      | PendingInternalFile | Contributor at X                       | Not Allowed |
+      | PendingInternalFile | Other contributors                     | Not Allowed |
+      | PendingInternalFile | File curator at X                      | Allowed     |
+      | PendingInternalFile | File curators for other contributors   | Not Allowed |
+      | PendingInternalFile | File curator by publication owner at X | Allowed     |
```
torbjokv, line 42: added this test

publication-commons/src/main/java/no/unit/nva/publication/permissions/file/FileStrategyBase.java
truhacevkir: This method check relation only, not "curator" relation, maybe rename it to something similar we have in PublicationPermissionStragies?

This is how exactly the same method is named there:

```
    protected boolean userRelatesToPublicationThroughPublicationOwnerOrCuratingInstitution() {
        return userIsFromSameInstitutionAsPublicationOwner() || userBelongsToCuratingInstitution();
    }
```

truhacevkir (approved):

PR #2084, opened 2025-03-03
persisting published publication log message with ticket owner

truhacevkir:
When consuming publishingRequest ticket and publishing publication, the user who pressed the "publish metadata" button should be set as person on PublicationPublished log message, and that user is a ticket owner :)


torbjokv (approved):
Looks good!

PR #2085, opened 2025-03-03
file imported log message in brage import

truhacevkir:
When updating publication with new file from Brage, log event should be FileImported. Now there is persisted FileUploaded event because `FileEntry.persist()` overrides value that is being set in `FileEntry.importFileEntry()` method.

### DIFF
--- a/publication-commons/src/main/java/no/unit/nva/publication/model/business/FileEntry.java
+++ b/publication-commons/src/main/java/no/unit/nva/publication/model/business/FileEntry.java
@@ -92,8 +92,8 @@ public final class FileEntry implements Entity, QueryObject<FileEntry> {
         return (FileEntry) fileDao.getData();
     }
 
-    public static FileEntry importFileEntry(File file, SortableIdentifier identifier, UserInstance userInstance,
-                                            ImportSource importSource) {
+    public static FileEntry createFromImportSource(File file, SortableIdentifier identifier,
+                                                   UserInstance userInstance, ImportSource importSource) {
         var fileEntry = create(file, identifier, userInstance);
         fileEntry.setFileEvent(FileImportedEvent.create(userInstance.getUser(), Instant.now(), importSource));
         return fileEntry;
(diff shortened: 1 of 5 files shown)


publication-commons/src/main/java/no/unit/nva/publication/model/business/FileEntry.java
```diff
@@ -92,8 +92,8 @@ public static FileEntry fromDao(FileDao fileDao) {
         return (FileEntry) fileDao.getData();
     }
 
-    public static FileEntry importFileEntry(File file, SortableIdentifier identifier, UserInstance userInstance,
-                                            ImportSource importSource) {
+    public static FileEntry createFromImportSource(File file, SortableIdentifier identifier,
+                                                   UserInstance userInstance, ImportSource importSource) {
         var fileEntry = create(file, identifier, userInstance);
         fileEntry.setFileEvent(FileImportedEvent.create(userInstance.getUser(), Instant.now(), importSource));
         return fileEntry;
```
truhacevkir, line 99: We have to set that FileImportEvent both here and in importNew() methods cause that one is used in resourceService when creating transaction for file when persisting files and resource in single call.

publication-commons/src/main/java/no/unit/nva/publication/service/impl/ResourceService.java
```diff
@@ -223,7 +223,7 @@ private Resource insertImportedResource(Resource resource, ImportSource importSo
 
         var userInstance = UserInstance.fromPublication(resource.toPublication());
         var fileTransactionWriteItems = resource.getFiles().stream()
-                                            .map(file -> FileEntry.importFileEntry(file, resource.getIdentifier(), userInstance, importSource))
+                                            .map(file -> FileEntry.createFromImportSource(file, resource.getIdentifier(), userInstance, importSource))
```
truhacevkir, line 226: I meant this method in that other comment :)

torbjokv (approved):

PR #2090, opened 2025-03-05
Np 48665 place scopus entries on queue

truhacevkir:
Sending Scopus files to import to SQS. Next step will be to consume sqs message instead of S3Event in ScopusHandler.


template.yaml
```diff
@@ -2296,25 +2295,51 @@ Resources:
...
+                    sqs_client.send_message(
+                        QueueUrl=scopus_queue,
+                        MessageBody=fname,
+                        MessageAttributes={
+                            'uri': {
+                                'DataType': 'String',
+                                'StringValue': s3_uri
+                            }
+                        }
+                    )
```
truhacevkir, line 2342: I put some effort to write it in java, but is was not so straightforward and code was ugly, so I decided simply to extend existing inline python code :)

template.yaml
```diff
@@ -2296,25 +2295,51 @@ Resources:
...
+                    s3_uri = f's3://{xml_bucket}/{fname}'
+                    sqs_client.send_message(
+                        QueueUrl=scopus_queue,
+                        MessageBody=fname,
+                        MessageAttributes={
+                            'uri': {
+                                'DataType': 'String',
+                                'StringValue': s3_uri
+                            }
+                        }
```
brinxmat, line 2341: Is the indentation here correct?

truhacevkir: Yes, it is.

PR #2093, opened 2025-03-06
Np 48665 consume scopus entries from queue

truhacevkir:
Making ScopusFunction consume SQS event instead of S3Event. 

This improvement makes that we can process xml files received from Scopus when it suits best for us. 

Changes:

- Handler consumes SQS event
- Setting ScopusFunction MaximumConcurrency to 5, so there is max 5 concurrent lambdas processing candidates.
- Setting ScopusFunction BatchSize to 1, so we process 1 candidate per invocation.

### DIFF
--- a/publication-event-handlers/src/main/java/no/unit/nva/publication/events/handlers/delete/DeleteImportCandidateEventConsumer.java
+++ b/publication-event-handlers/src/main/java/no/unit/nva/publication/events/handlers/delete/DeleteImportCandidateEventConsumer.java
@@ -30,7 +30,7 @@ public class DeleteImportCandidateEventConsumer
     public static final String EVENTS_BUCKET = new Environment().readEnv("EVENTS_BUCKET");
     public static final String API_HOST = new Environment().readEnv("API_HOST");
     public static final String SEARCH = "search";
-    public static final String IMPORT_CANDIDATES_2 = "import-candidates2";
+    public static final String IMPORT_CANDIDATES = "import-candidates";
     public static final String CONTENT_TYPE = "application/json";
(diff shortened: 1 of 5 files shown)


publication-event-handlers/src/main/java/no/unit/nva/publication/events/handlers/delete/DeleteImportCandidateEventConsumer.java
```diff
@@ -109,7 +109,7 @@ private Optional<String> getResponseBody(URI uri) {
     private URI constructUri(String scopusIdentifier) {
         return UriWrapper.fromHost(API_HOST)
                    .addChild(SEARCH)
-                   .addChild(IMPORT_CANDIDATES_2)
+                   .addChild(IMPORT_CANDIDATES)
```
truhacevkir, line 112: import-candidates2 does not exist longer.

template.yaml
```diff
@@ -1360,20 +1360,18 @@ Resources:
...
-            Events: 's3:ObjectCreated:*'
-            Filter:
-              S3Key:
-                Rules:
-                  - Name: suffix
-                    Value: '.xml'
+            Queue: !GetAtt ScopusImportQueue.Arn
+            BatchSize: 1
+            ScalingConfig:
+              MaximumConcurrency: 5  # Limit the number of concurrent invocations from this SQS queue
```
truhacevkir, line 1375: No OnFailureConfig, we store failing reports to S3 bucket.

brinxmat (approved):

torbjokv (approved):
Nice!

PR #2098, opened 2025-03-18
completing support case when all messages are deleted

truhacevkir:


messages/src/main/java/no/unit/nva/publication/messages/delete/DeleteMessageHandler.java
```diff
@@ -22,35 +29,73 @@ public class DeleteMessageHandler extends ApiGatewayHandler<Void, Void> {
...
+        throws NotFoundException {
+        var ticket = ticketService.fetchTicketByIdentifier(ticketIdentifier);
+        if (ticket instanceof GeneralSupportRequest) {
+            var messages = ticket.fetchMessages(ticketService);
+            if (allMessagesAreDeleted(messages)) {
+                var resource = getResource(ticket.getResourceIdentifier());
+                ticket.complete(resource.toPublication(), userInstance).persistUpdate(ticketService);
+            }
+        }
+    }
```
truhacevkir, line 91: It would be nice if we could fetch both tickets and messages by ticketIdentifier by single database transaction, but it is not possible.

PR #2099, opened 2025-03-19
Np 48757 remove filename encoding on upload

truhacevkir:
Decoding filename after it has been stored to S3/ before storing to database.

### DIFF
--- a/publication-file/src/main/java/no/unit/nva/publication/file/upload/FileService.java
+++ b/publication-file/src/main/java/no/unit/nva/publication/file/upload/FileService.java
@@ -2,7 +2,6 @@ package no.unit.nva.publication.file.upload;
 
 import static no.unit.nva.model.associatedartifacts.RightsRetentionStrategyConfiguration.RIGHTS_RETENTION_STRATEGY;
 import static no.unit.nva.publication.file.upload.config.MultipartUploadConfig.BUCKET_NAME;
-import static no.unit.nva.publication.file.upload.restmodel.CreateUploadRequestBody.LAST_ASCII_CODEPOINT;
 import com.amazonaws.services.s3.AmazonS3;
 import com.amazonaws.services.s3.AmazonS3ClientBuilder;
 import com.amazonaws.services.s3.model.GetObjectMetadataRequest;
--- a/publication-file/src/main/java/no/unit/nva/publication/file/upload/FileService.java
+++ b/publication-file/src/main/java/no/unit/nva/publication/file/upload/FileService.java
@@ -2,6 +2,7 @@ package no.unit.nva.publication.file.upload;
 
 import static no.unit.nva.model.associatedartifacts.RightsRetentionStrategyConfiguration.RIGHTS_RETENTION_STRATEGY;
 import static no.unit.nva.publication.file.upload.config.MultipartUploadConfig.BUCKET_NAME;
+import static no.unit.nva.publication.file.upload.restmodel.CreateUploadRequestBody.LAST_ASCII_CODEPOINT;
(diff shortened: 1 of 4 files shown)


publication-file/src/main/java/no/unit/nva/publication/file/upload/FileService.java
```diff
@@ -179,7 +180,10 @@ private static void validateUploadPermissions(UserInstance userInstance, Resourc
     private static String toFileName(String contentDisposition) {
         var pattern = Pattern.compile(FILE_NAME_REGEX);
         var matcher = pattern.matcher(contentDisposition);
-        return matcher.matches() ? matcher.group(1) : contentDisposition;
+        var unicodeUnescaper = new UnicodeUnescaper();
+        return matcher.matches()
+                   ? unicodeUnescaper.translate(matcher.group(1))
+                   : unicodeUnescaper.translate(contentDisposition);
```
truhacevkir, line 186: This solves the problem to Armaz. Filename in S3 object metadata will be encoded (as it is now), but filename in database will have its original value.

brinxmat: Would be tempted to do something to avoid thread unsafe stuff and the issues regexes introduce with something a bit simpler:

Pull this into its own class.

```
import org.apache.commons.text.translate.UnicodeUnescaper;

public class ClassName {
    private static final String PREFIX = "filename=\"";
    private static final String SUFFIX = "\"";
    private static final UnicodeUnescaper UNESCAPER = new UnicodeUnescaper();

    private ClassName() {
        // NO-OP
    }

    public String unescape(String value) {
        return UNESCAPER.translate(extractFilename(value));
    }

    private static String extractFilename(String input) {
        return (input.startsWith(PREFIX) && input.endsWith(SUFFIX))
                   ? input.substring(PREFIX.length(), input.length() - SUFFIX.length())
                   : input;
    }
}
```

truhacevkir: Moving this logic to final Filename class:

```
    private Filename() {

    }

    public static String toContentDispositionValue(String value) {
        var escapedFilename = ESCAPER.translate(value);
        return String.format(CONTENT_DISPOSITION_TEMPLATE, escapedFilename);
    }

    public static String fromContentDispositionValue(String value) {
        var filename = getFilenameFromContentDisposition(value);
        return UNESCAPER.translate(filename);
    }

    private static String getFilenameFromContentDisposition(String value) {
        return (value.startsWith(PREFIX) && value.endsWith(SUFFIX))
                   ? value.substring(PREFIX.length(), value.length() - SUFFIX.length())
                   : value;
    }
```

publication-file/src/main/java/no/unit/nva/publication/file/upload/Filename.java
```diff
@@ -0,0 +1,34 @@
...
+
+import org.apache.commons.text.translate.UnicodeEscaper;
+import org.apache.commons.text.translate.UnicodeUnescaper;
+
+public final class Filename {
+
+    private static final String CONTENT_DISPOSITION_TEMPLATE = "filename=\"%s\"";
+    private static final String PREFIX = "filename=\"";
+    private static final String SUFFIX = "\"";
+    public static final int LAST_ASCII_CODEPOINT = 127;
```
brinxmat, line 11: Is there a reason why this is public?

truhacevkir: Nope, will fix!

PR #2100, opened 2025-03-19
new model for log

truhacevkir:
Tasks:
https://sikt.atlassian.net/browse/NP-48603
https://sikt.atlassian.net/browse/NP-48602

New model for log messages as described in tasks.

Main change:

- Instead of constructing Person and Organization in log message based on NVA user in NVA database, we use CristinPerson and CristinOrganization to construct log message. 
- Log message topic/ action can now be performed by user or institution (LogAgent) -> important for migration jobs, cause there are no users involved, so we avoid persist "dummy" empty object. 
- We first fetch user from users-and-roles table -> than we fetch CristinPerson and CristinOrganization from Cristin and using these to construct log message. 
- Renaming LogUser and LogOrganization to Person and Organization when storing in database and serving via API.

### DIFF
--- a/publication-log/src/main/java/no/unit/nva/publication/log/service/LogEntryService.java
+++ b/publication-log/src/main/java/no/unit/nva/publication/log/service/LogEntryService.java
@@ -26,8 +26,8 @@ import org.slf4j.LoggerFactory;
 public class LogEntryService {
 
     public static final String PERSISTING_LOG_ENTRY_MESSAGE = "Persisting log entry for event {} for resource {}";
-    public static final String PERSISTING_FILE_LOG_ENTRY_MESSAGE =
-        "Persisting log entry for event {} for file {} for resource {}";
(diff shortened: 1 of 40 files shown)


publication-log/src/main/java/no/unit/nva/publication/log/service/LogEntryService.java
```diff
@@ -74,43 +79,52 @@ private void persistLogEntry(DoiRequest doiRequest) {
...
+            var organization = logOrganizationFromCristinId(resourceEvent.institution());
+            return resourceEvent.toLogEntry(resource.getIdentifier(), organization);
         } else {
-            var user = createUser(resourceEvent);
+            var user = createUser(resourceEvent.user(), resourceEvent.institution());
             return resourceEvent.toLogEntry(resource.getIdentifier(), user);
         }
     }
 
+    private LogOrganization logOrganizationFromCristinId(URI organizationId) {
```
brinxmat, line 97: I feel this code doesn't really belong here.

brinxmat (approved):

PR #2101, opened 2025-03-20
Np 48547 craft log for publication

truhacevkir:
Adding back code for crafting log entries in production.

Following log entries should be persisted in production when relevant:

- PublicationCreated
- PublicationPublished
- PublicationImport (NVE)
- FileUploaded
- FileApproved
- FileImported (NVE)
- DoiRequested
- DoiAssigned


publication-commons/src/main/java/no/unit/nva/publication/service/impl/ResourceService.java
```diff
@@ -100,6 +112,7 @@ public class ResourceService extends ServiceWithTransactions {
     private static final String SEPARATOR_ITEM = ",";
     private static final String SEPARATOR_TABLE = ";";
     private static final Logger logger = LoggerFactory.getLogger(ResourceService.class);
+    public static final String NVE_IMPORTED_RESOURCE_OWNER = "nve@5948.0.0.0";
```
brinxmat, line 115: public?

brinxmat (approved):

PR #2104, opened 2025-03-21
migrating rightsholder as legalNote on file

truhacevkir:
Instead of migrating rightsHolder from Brage as rightsHolder migrating it as legalNote for file when accessCode from Brage is missing. 

It makes that "Copyright..." string will be presented for each migrated file from Brage (when present).


brage-import/src/test/java/no/sikt/nva/brage/migration/mapper/BrageNvaMapperTest.java
```diff
@@ -156,6 +156,21 @@ void shouldMapInsperaAndWiseflowIdentifierToAdditionalIdentifiers()
         assertEquals(insperaIdentifier, insperaAdditionalIdentifier.value());
     }
 
+    @Test
+    void shouldMapRightsHolderToFileLegalNotWhenAccessCodeIsNotPresent()
```
brinxmat, line 160: LegalNot(e)

brinxmat (approved):

PR #2106, opened 2025-03-24
Combine expansion handlers

truhacevkir:
Combining ResourceExpansionHandler and ExpandedDataEntriesPersistenceHandler. 

We are replacing this:

image

 With this:

image

This should improve indexing time even more. 

Logic before:
ExpansionHandler: Expand document -> Write to S3 -> Send forward
PersistanceHandler: Read file from S3 -> wrap expanded document into IndexDocument -> persist to S3 persistedResources -> send forward

Now: 
ExpansionHandler: Expand document -> wrap expanded document into IndexDocument ->  persist to S3 persistedResources -> Send forward


truhacevkir:
The only issue here is that we disconnect AnalyticsHandler ExpansionHandler, at the same time AnalyticsHandler has been disabled since I started at project, so I think it is fine :)

image

publication-event-handlers/src/main/java/no/unit/nva/publication/events/handlers/persistence/AnalyticsIntegrationHandler.java
```diff
@@ -23,6 +22,7 @@
 
 public class AnalyticsIntegrationHandler extends DestinationsEventBridgeEventHandler<EventReference, EventReference> {
     
+    public static final String EXPANDED_ENTRY_UPDATED_EVENT_TOPIC = "PublicationService.ExpandedDataEntry.Update";
```
brinxmat, line 25: Needs to be public?

brinxmat (approved):

PR #2107, opened 2025-03-25
Attempting to fix ticket bug

truhacevkir:


publication-rest/src/main/java/no/unit/nva/publication/update/PublishingRequestResolver.java
```diff
@@ -73,18 +73,11 @@ private void handlePublishingRequest(Publication oldImage, Publication newImage)
         if (thereAreNoPendingFiles(newImage)) {
             autoCompletePendingPublishingRequestsIfNeeded(newImage, pendingPublishingRequests);
             return;
-        }
-        if (updateHasPendingFileChanges(oldImage, newImage)) {
+        } else {
             updateFilesForApproval(oldImage, newImage, pendingPublishingRequests);
```
truhacevkir, line 77: This is what solves issue with deleted pending file not being removed from publishing request. Because after file has been removed, existing fetched publication files are equal to updated file, when file types are not changed.

torbjokv (approved):

PR #2108, opened 2025-03-25
Np 48782 set message as unread on creation

truhacevkir:
When creating ticket with a message or sending a new message
Then sender is not set as viewedBy on that message.

This makes that when owner sends a message on ticket, he/she will get a notification about the message.


publication-commons/src/main/java/no/unit/nva/publication/model/business/GeneralSupportRequest.java
```diff
@@ -62,7 +63,7 @@ public static GeneralSupportRequest create(Resource resource, UserInstance userI
         generalSupportRequest.setModifiedDate(Instant.now());
         generalSupportRequest.setStatus(TicketStatus.PENDING);
         generalSupportRequest.setIdentifier(SortableIdentifier.next());
-        generalSupportRequest.setViewedBy(ViewedBy.addAll(userInstance.getUser()));
+        generalSupportRequest.setViewedBy(Set.of());
```
brinxmat, line 66: Here as elsewhere, Collections::emptySet

brinxmat (changes requested):
Just one general comment about using Set::of rather than Collections::emptySet, which indicates that the intention is an empty set, rather than a set that has missing elements. Both are immutable, so I would probably check that setting a value actually works — we probably have a test for this, but maybe it's worth verifying that we don't set weird things in the test set up so the initial state isn't actually ever tested.

publication-commons/src/main/java/no/unit/nva/publication/model/business/GeneralSupportRequest.java
truhacevkir: Will do it. We run following for every ticket on creation:
```
   protected TicketEntry() {
        viewedBy = ViewedBy.empty();
    }
```

brinxmat (approved):

PR #2114, opened 2025-03-28
log message for publication update

truhacevkir:
Creating LogEntry with topic PublicationUpdated which will be persisted each time we do an update on publication.

Next step should be to stop backend from updating database entry when there is no change in publication, so we do not updates without effective changes, otherwise there will be persisted PublicationUpdated log message for each "Update publication API call"

### DIFF
--- a/publication-commons/src/main/java/no/unit/nva/publication/model/business/publicationstate/ResourceEvent.java
+++ b/publication-commons/src/main/java/no/unit/nva/publication/model/business/publicationstate/ResourceEvent.java
@@ -10,10 +10,11 @@ import no.unit.nva.publication.model.business.logentry.LogAgent;
 import no.unit.nva.publication.model.business.logentry.PublicationLogEntry;
 
 @JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
-@JsonSubTypes({@JsonSubTypes.Type(CreatedResourceEvent.class), @JsonSubTypes.Type(PublishedResourceEvent.class),
-    @JsonSubTypes.Type(UnpublishedResourceEvent.class), @JsonSubTypes.Type(DeletedResourceEvent.class),
-    @JsonSubTypes.Type(RepublishedResourceEvent.class), @JsonSubTypes.Type(ImportedResourceEvent.class),
-    @JsonSubTypes.Type(DoiReservedEvent.class), @JsonSubTypes.Type(MergedResourceEvent.class)})
+@JsonSubTypes({@JsonSubTypes.Type(CreatedResourceEvent.class), @JsonSubTypes.Type(UpdatedResourceEvent.class),
+    @JsonSubTypes.Type(PublishedResourceEvent.class), @JsonSubTypes.Type(UnpublishedResourceEvent.class),
+    @JsonSubTypes.Type(DeletedResourceEvent.class), @JsonSubTypes.Type(RepublishedResourceEvent.class),
+    @JsonSubTypes.Type(ImportedResourceEvent.class), @JsonSubTypes.Type(DoiReservedEvent.class),
+    @JsonSubTypes.Type(MergedResourceEvent.class)})
 public interface ResourceEvent {
 
     Instant date();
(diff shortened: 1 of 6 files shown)


publication-commons/src/main/java/no/unit/nva/publication/model/business/publicationstate/ResourceEvent.java
```diff
@@ -10,10 +10,11 @@
...
 @JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
-@JsonSubTypes({@JsonSubTypes.Type(CreatedResourceEvent.class), @JsonSubTypes.Type(PublishedResourceEvent.class),
-    @JsonSubTypes.Type(UnpublishedResourceEvent.class), @JsonSubTypes.Type(DeletedResourceEvent.class),
-    @JsonSubTypes.Type(RepublishedResourceEvent.class), @JsonSubTypes.Type(ImportedResourceEvent.class),
-    @JsonSubTypes.Type(DoiReservedEvent.class), @JsonSubTypes.Type(MergedResourceEvent.class)})
+@JsonSubTypes({@JsonSubTypes.Type(CreatedResourceEvent.class), @JsonSubTypes.Type(UpdatedResourceEvent.class),
+    @JsonSubTypes.Type(PublishedResourceEvent.class), @JsonSubTypes.Type(UnpublishedResourceEvent.class),
+    @JsonSubTypes.Type(DeletedResourceEvent.class), @JsonSubTypes.Type(RepublishedResourceEvent.class),
+    @JsonSubTypes.Type(ImportedResourceEvent.class), @JsonSubTypes.Type(DoiReservedEvent.class),
+    @JsonSubTypes.Type(MergedResourceEvent.class)})
```
brinxmat, line 17: Wondering if we should sort this list alphabetically and do one-statement-per-line separation. so we can see the actual update when the code changes

publication-commons/src/main/java/no/unit/nva/publication/model/business/publicationstate/UpdatedResourceEvent.java
```diff
@@ -0,0 +1,30 @@
...
+import java.net.URI;
+import java.time.Instant;
+import no.unit.nva.identifiers.SortableIdentifier;
+import no.unit.nva.publication.model.business.User;
+import no.unit.nva.publication.model.business.UserInstance;
+import no.unit.nva.publication.model.business.logentry.LogAgent;
+import no.unit.nva.publication.model.business.logentry.LogTopic;
+import no.unit.nva.publication.model.business.logentry.PublicationLogEntry;
+
+public record UpdatedResourceEvent(Instant date, User user, URI institution, SortableIdentifier identifier)
```
brinxmat, line 12: It's a `date`, but it isn't; here as elsewhere

publication-commons/src/main/java/no/unit/nva/publication/model/business/publicationstate/UpdatedResourceEvent.java
```diff
@@ -0,0 +1,30 @@
...
+                                        SortableIdentifier.next());
+    }
+
+    @Override
+    public PublicationLogEntry toLogEntry(SortableIdentifier resourceIdentifier, LogAgent user) {
+        return PublicationLogEntry.builder()
+                   .withResourceIdentifier(resourceIdentifier)
+                   .withIdentifier(identifier)
+                   .withTopic(LogTopic.PUBLICATION_UPDATED)
+                   .withTimestamp(date)
```
brinxmat, line 26: Bingo!

brinxmat (changes requested):

publication-commons/src/main/java/no/unit/nva/publication/model/business/publicationstate/UpdatedResourceEvent.java
truhacevkir: Yeah, we discussed it before. There is a task for it :)
Changing it now will be a large refactor change + we need a migration for migrating this field in database.

brinxmat (approved):

PR #2116, opened 2025-03-31
not emitting event for deleted file entry

truhacevkir:
Following method for DataEntryUpdateEvent throws exception when FileEntry does not have newImage. I think we do not want to throw exception here, we simply do not want to emit event when file has been deleted. 
```
    @JsonProperty("topic")
    public String getTopic() {
        var type = extractDataEntryType();
        return switch (type) {
            case Resource resource -> RESOURCE_UPDATE_EVENT_TOPIC;
            case DoiRequest doiRequest -> DOI_REQUEST_UPDATE_EVENT_TOPIC;
            case PublishingRequestCase publishingRequestCase -> PUBLISHING_REQUEST_UPDATE_EVENT_TOPIC;
            case Message message -> MESSAGE_UPDATE_EVENT_TOPIC;
            case GeneralSupportRequest generalSupportRequest -> GENERAL_SUPPORT_REQUEST_UPDATE_EVENT_TOPIC;
            case UnpublishRequest unpublishRequest -> UNPUBLISH_REQUEST_UPDATE_EVENT_TOPIC;
            case FileEntry fileEntry when hasNewImage() -> fileEntry.getFileEvent() instanceof FileDeletedEvent
                                            ? FILE_ENTRY_DELETE_EVENT_TOPIC
                                            : FILE_ENTRY_UPDATE_EVENT_TOPIC;
            default -> throw new IllegalArgumentException("Unknown entry type: " + type);
        };
    }
```

Solution:

Not emitting event when new image for file entry is null:
```
return dataEntryUpdateEvent.shouldProcessUpdate() ? sendEvent(dataEntryUpdateEvent, context) : DO_NOT_EMIT_EVENT;
        
public boolean shouldProcessUpdate() {
        if (extractDataEntryType() instanceof FileEntry) {
            return hasNewImage();
        } else {
            return nonNull(oldData) || nonNull(newData);
        }
    }
```

### DIFF
--- a/publication-event-handlers/src/main/java/no/unit/nva/publication/events/bodies/DataEntryUpdateEvent.java
+++ b/publication-event-handlers/src/main/java/no/unit/nva/publication/events/bodies/DataEntryUpdateEvent.java
@@ -98,8 +98,12 @@ public class DataEntryUpdateEvent implements JsonSerializable {
     }
     
     @JsonIgnore
-    public boolean notEmpty() {
-        return nonNull(oldData) || nonNull(newData);
+    public boolean shouldProcessUpdate() {
+        if (extractDataEntryType() instanceof FileEntry) {
+            return hasNewImage();
+        } else {
+            return nonNull(oldData) || nonNull(newData);
+        }
     }
     
     @JsonProperty("topic")
@@ -118,7 +122,7 @@ public class DataEntryUpdateEvent implements JsonSerializable {
             default -> throw new IllegalArgumentException("Unknown entry type: " + type);
         };
     }
-    
+
     private Entity extractDataEntryType() {
         return nonNull(newData) ? newData : oldData;
     }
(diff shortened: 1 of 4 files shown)


publication-event-handlers/src/main/java/no/unit/nva/publication/events/bodies/DataEntryUpdateEvent.java
```diff
@@ -98,8 +98,12 @@ && getTopic().equals(that.getTopic())
     }
     
     @JsonIgnore
-    public boolean notEmpty() {
-        return nonNull(oldData) || nonNull(newData);
+    public boolean shouldProcessUpdate() {
```
brinxmat, line 101: I added task: https://sikt.atlassian.net/browse/NP-48858

brinxmat (approved):
See comment, approved

PR #2120, opened 2025-04-01
adding ffi customer id in migration code

truhacevkir:


brinxmat (approved):
Assuming these strings have been triple-checked against actual values

PR #2125, opened 2025-04-03
Np 48801 update publication only when with effective changes

truhacevkir:
Not updating publication when there are no effective changes. For now effective change is any change except updated "date", no matter what kind of date it is. 

Following values are ignored when checking for effective changes - createdDate, publishedDate, modifiedDate, indexedDate.

### DIFF
--- a/publication-commons/src/main/java/no/unit/nva/publication/model/business/Resource.java
+++ b/publication-commons/src/main/java/no/unit/nva/publication/model/business/Resource.java
@@ -18,6 +18,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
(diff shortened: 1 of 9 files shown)


publication-model/src/main/java/no/unit/nva/model/instancetypes/degree/DegreePhd.java
```diff
@@ -20,6 +22,24 @@ public DegreePhd(@JsonProperty(PAGES_FIELD) MonographPages pages,
...
+            return false;
+        }
+        return Objects.equals(getRelated(), degreePhd.getRelated());
+    }
+
+    @JacocoGenerated
+    @Override
+    public int hashCode() {
+        return Objects.hash(super.hashCode(), getRelated());
+    }
```
truhacevkir, line 41: Lack of equals method here made that relatedDocuments has never been verified in assertions in test.

publication-commons/src/main/java/no/unit/nva/publication/model/business/Resource.java
```diff
@@ -223,7 +224,28 @@ public boolean isDegree() {
             .orElse(false);
     }
 
-   private static boolean instanceIsDegree(PublicationInstance<? extends Pages> publicationInstance) {
+    @JsonIgnore
+    public boolean hasAffectiveChanges(Resource resource) {
+        return !(Objects.equals(getIdentifier(), resource.getIdentifier())
```
brinxmat, line 229: Logically, createdDate cannot change…

publication-commons/src/main/java/no/unit/nva/publication/service/impl/UpdateResourceService.java
```diff
@@ -118,13 +118,17 @@ public void updateOwner(SortableIdentifier identifier, UserInstance oldOwner, Us
 
     public Resource updateResource(Resource resource, UserInstance userInstance) {
         var persistedResource = fetchExistingResource(resource.toPublication());
-        resource.setCreatedDate(persistedResource.getCreatedDate());
-        resource.setModifiedDate(clockForTimestamps.instant());
 
-        updateCuratingInstitutions(resource, persistedResource);
+        if (resource.hasAffectiveChanges(persistedResource)) {
+            resource.setCreatedDate(persistedResource.getCreatedDate());
```
brinxmat, line 123: This makes me think that we should have a broader chat about how we (don't) mutate objects

brinxmat (approved):

PR #2135, opened 2025-04-24
Np 49014 persist file approval thesis

truhacevkir:
- Persisting FileApprovalThesis in PublishPublicationHandler
- Moving publishing of publication to PublishingService
- Persisting PublishingRequestCase in PublishPublicationHandler as well, so there will be no need to call CreateTicketHandler to publish publication.

### DIFF
--- a/tickets/src/main/java/no/unit/nva/publication/ticket/FileApprovalThesisDto.java
+++ b/tickets/src/main/java/no/unit/nva/publication/ticket/FileApprovalThesisDto.java
@@ -2,6 +2,9 @@ package no.unit.nva.publication.ticket;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import com.fasterxml.jackson.annotation.JsonTypeInfo.Id;
+import com.fasterxml.jackson.annotation.JsonTypeName;
 import java.net.URI;
 import java.time.Instant;
 import java.util.List;
@@ -16,11 +19,14 @@ import no.unit.nva.publication.model.business.TicketEntry;
 import no.unit.nva.publication.model.business.User;
 import nva.commons.core.JacocoGenerated;
 
+@JsonTypeInfo(use = Id.NAME, property = "type")
+@JsonTypeName(FileApprovalThesisDto.TYPE)
 public class FileApprovalThesisDto extends TicketDto {
 
     private static final String WORKFLOW_FIELD = "workflow";
     private static final String APPROVED_FILES_FIELD = "approvedFiles";
     private static final String FILES_FOR_APPROVAL = "filesForApproval";
+    public static final String TYPE = "FileApprovalThesis";
 
     @JsonProperty(WORKFLOW_FIELD)
     private final PublishingWorkflow workflow;


publication-commons/src/main/java/no/unit/nva/publication/service/impl/PublishingService.java
```diff
@@ -0,0 +1,169 @@
...
+    public void publishResource(SortableIdentifier resourceIdentifier, UserInstance userInstance)
+        throws ApiGatewayException {
+        var resource = getResource(resourceIdentifier);
+        validatePermissions(resource, userInstance);
+
+        publishPublication(userInstance, resource);
+
+        if (!resource.getPendingFiles().isEmpty()) {
+            publishResourceWithPendingFiles(userInstance, resource);
+        }
```
truhacevkir, line 61: Rules for publishing and ticket creation are described here:

https://sikt.atlassian.net/browse/NP-48980

publication-commons/src/main/java/no/unit/nva/publication/service/impl/PublishingService.java
```diff
@@ -0,0 +1,169 @@
...
+            var channelClaim = getChannelClaim(publisher.get());
+            if (channelClaim.isPresent() && isClaimedByDifferentOrganization(channelClaim.get(), userInstance)) {
+                var organizationId = getOrganizationId(channelClaim.get());
+                FilesApprovalThesis.create(resource, userInstance, organizationId, workflow)
+                    .persistNewTicket(ticketService);
+                return;
+            }
+        }
+        FilesApprovalThesis.createForUserInstitution(resource, userInstance, workflow)
+            .persistNewTicket(ticketService);
```
brinxmat, line 133: Simplify this method body. There are two terminal states; pull the publisher check and channel claim out to a named method that returns Optional<ChannelClaim> and if present do ? one : two.

publication-commons/src/main/java/no/unit/nva/publication/service/impl/PublishingService.java
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    private boolean isClaimedByDifferentOrganization(ChannelClaimDto channelClaim, UserInstance userInstance) {
+        return !channelClaim.claimedBy().id().equals(userInstance.getCustomerId());
+    }
+
+    private URI createChannelClaimUri(Publisher publisher) {
+        var channelClaimIdentifier = UriWrapper.fromUri(publisher.getId())
+                                         .replacePathElementByIndexFromEnd(0, StringUtils.EMPTY_STRING)
+                                         .getLastPathElement();
```
brinxmat, line 157: This is because channel claims currently have years?

publication-commons/src/test/java/no/unit/nva/publication/service/impl/PublishingServiceTest.java
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    @Test
+    void shouldPublishResourceWhenPublicationIsPublishable() throws ApiGatewayException {
+        var publication = randomPublication(JournalArticle.class).copy()
+                              .withStatus(PublicationStatus.DRAFT)
+                              .withAssociatedArtifacts(List.of())
+                              .build();
+        var userInstance = UserInstance.fromPublication(publication);
+        publication = Resource.fromPublication(publication).persistNew(resourceService, userInstance);
```
brinxmat, line 71: IS there a reason to reassign publication with a resource?

truhacevkir: Yes, the first one is not persisted, it is ugly because Resource.persistNew() returns publication.
I will extract whole thing to a method, so it look prettier.

publication-commons/src/main/java/no/unit/nva/publication/service/impl/PublishingService.java
```diff
@@ -0,0 +1,169 @@
...
+            return Optional.of(identityServiceClient.getChannelClaim(createChannelClaimUri(publisher)));
+        } catch (NotFoundException exception) {
+            return Optional.empty();
+        } catch (Exception e) {
+            throw new BadGatewayException("Could not fetch channel owner!");
+        }
+    }
+
+    private boolean isClaimedByDifferentOrganization(ChannelClaimDto channelClaim, UserInstance userInstance) {
+        return !channelClaim.claimedBy().id().equals(userInstance.getCustomerId());
```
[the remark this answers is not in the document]
truhacevkir: It can, if we get a claimed channel in response, but the claimer is missing. 
Will wrap in optional.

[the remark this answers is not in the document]
truhacevkir: Updating to:

```
        return Optional.ofNullable(channelClaim)
                   .map(ChannelClaimDto::claimedBy)
                   .map(CustomerSummaryDto::id)
                   .map(id -> !userInstance.getCustomerId().equals(id))
                   .orElse(true);
```

publication-commons/src/main/java/no/unit/nva/publication/service/impl/PublishingService.java
brinxmat: I'd maybe therefore comprehend the structure and optionally remove the unmatching part (to be future-proofed for the same without year or trailing slash)

publication-commons/src/test/java/no/unit/nva/publication/service/impl/PublishingServiceTest.java
brinxmat: It would be enough to just assign a different variable since it isn't a comparable object anyway

publication-commons/src/main/java/no/unit/nva/publication/service/impl/PublishingService.java
truhacevkir: I dont think we can, I have tried :(
I have to handle multiple exceptions, I can only wrap in IfPresent, but it looks ugly:

```
        getPublisher(resource).ifPresent(publisher -> {
            try {
                getChannelClaim(publisher).ifPresent(channelClaim -> {
                    if (isClaimedByDifferentOrganization(channelClaim, userInstance)) {
                        try {
                            FilesApprovalThesis.create(resource, userInstance, getOrganizationId(channelClaim), workflow)
                                .persistNewTicket(ticketService);
                        } catch (ApiGatewayException e) {
                            throw new RuntimeException(e);
                        }
                    }
                });
                FilesApprovalThesis.createForUserInstitution(resource, userInstance, workflow)
                    .persistNewTicket(ticketService);

            } catch (BadGatewayException e) {
                throw new RuntimeException(e);
            } catch (ApiGatewayException e) {
                throw new RuntimeException(e);
            }
        });
```

publication-commons/src/main/java/no/unit/nva/publication/service/impl/PublishingService.java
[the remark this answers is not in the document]
brinxmat: In general, this kind of chain-of-calls tells us there is some missing abstraction since we're delving deep into the friend-object's internals, and is as a consequence a locus for incomprehensible errors at runtime.

publication-commons/src/main/java/no/unit/nva/publication/service/impl/PublishingService.java
```diff
@@ -0,0 +1,169 @@
...
+        }
+
+        var channelClaim = getChannelClaim(publisher.get());
+        if (channelClaim.isEmpty()
+            || EVERYONE.equals(channelClaim.get().channelClaim().constraint().publishingPolicy())) {
+            resource.publish(resourceService, userInstance);
+            return;
+        }
+
+        if (!userInstance.getTopLevelOrgCristinId().equals(channelClaim.get().claimedBy().organizationId())) {
```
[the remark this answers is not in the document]
truhacevkir: Claim is lying in commons, i will wrap in optional.

publication-commons/src/main/java/no/unit/nva/publication/service/impl/PublishingService.java
```diff
@@ -0,0 +1,169 @@
...
+        throws BadGatewayException, ForbiddenException {
+        var publisher = getPublisher(resource);
+        if (publisher.isEmpty()) {
+            resource.publish(resourceService, userInstance);
+            return;
+        }
+
+        var channelClaim = getChannelClaim(publisher.get());
+        if (channelClaim.isEmpty()
+            || EVERYONE.equals(channelClaim.get().channelClaim().constraint().publishingPolicy())) {
```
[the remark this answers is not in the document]
truhacevkir: Will change to:

channelClaim.isEmpty() || everyoneCanPublish(channelClaim.get()) -> 

```
    private static boolean everyoneCanPublish(ChannelClaimDto channelClaim) {
        return Optional.ofNullable(channelClaim)
                   .map(ChannelClaimDto::channelClaim)
                   .map(ChannelClaim::constraint)
                   .map(ChannelConstraint::publishingPolicy)
                   .map(EVERYONE::equals)
                   .orElse(false);
    }
```

publication-commons/src/main/java/no/unit/nva/publication/model/business/FilesApprovalThesis.java
```diff
@@ -25,8 +25,16 @@ public FilesApprovalThesis() {
         super();
     }
 
-    public static FilesApprovalThesis create(Resource resource, UserInstance userInstance,
+    public static FilesApprovalThesis create(Resource resource, UserInstance userInstance, URI organizationId,
                                              PublishingWorkflow workflow) {
+        var fileApproval = createFileApproval(resource, userInstance, organizationId, workflow);
```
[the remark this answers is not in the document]
truhacevkir: I agree, for example workflow in something we fetch "from the side" when instantiating ticket, and in this case we also do the same for organization. Do you have any suggestion how this can be simplified?

publication-commons/src/main/java/no/unit/nva/publication/service/impl/PublishingService.java
truhacevkir: Agree, but here we are consuming publisher id from publication. Will this value also be missing `/{year}` part?

brinxmat (approved):

PR #2138, opened 2025-04-24
making accepted publishing request handler handle FilesApprovalThesis

truhacevkir:
Making AcceptedPublishingRequestEventHandler consume FileApprovalEntry, which is a base class for PublishingRequestCase and FilesApprovalThesis. In will handle publishing of files for thesis when FilesApprovalThesis is approved by curators.

### DIFF
--- a/publication-commons/src/main/java/no/unit/nva/publication/model/FilesApprovalEntry.java
+++ b/publication-commons/src/main/java/no/unit/nva/publication/model/FilesApprovalEntry.java
@@ -61,6 +61,15 @@ public abstract class FilesApprovalEntry extends TicketEntry {
         return this.complete(publication, userInstance).persistNewTicket(ticketService);
     }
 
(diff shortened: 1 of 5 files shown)


publication-commons/src/main/java/no/unit/nva/publication/model/FilesApprovalEntry.java
```diff
@@ -61,6 +61,15 @@ public TicketEntry persistAutoComplete(TicketService ticketService, Publication
         return this.complete(publication, userInstance).persistNewTicket(ticketService);
     }
 
+    public void rejectRejectedFiles(ResourceService resourceService) {
+        getFilesForApproval().stream()
+            .map(PendingFile.class::cast)
+            .forEach(file -> FileEntry.queryObject(file.getIdentifier(), getResourceIdentifier())
```
brinxmat, line 67: Can something be extracted here?

brinxmat (approved):

PR #2142, opened 2025-04-28
Np 48995 allowed operation for approving files approval thesis

truhacevkir:
Introducing AllowedOperation for approving FilesApprovalThesis. Naming it `thesis-approval`, come with suggestions for better name :)
The name to match pattern should be files-approval-thesis-approve, but I mean it sound strange :)

We should probably also introduce new allowed operation for publishing, because right now frontend is using publishing-request-create, but we do not longer create ticket to publish publication. But publishing validation is also using channel-claims, so it should be done after we inject channel-claim to publication.

### DIFF
--- a/publication-commons/src/main/java/no/unit/nva/publication/permissions/publication/grant/CuratorGrantStrategy.java
+++ b/publication-commons/src/main/java/no/unit/nva/publication/permissions/publication/grant/CuratorGrantStrategy.java
@@ -1,5 +1,7 @@
 package no.unit.nva.publication.permissions.publication.grant;
 
+import static nva.commons.apigateway.AccessRight.MANAGE_DEGREE;
+import static nva.commons.apigateway.AccessRight.MANAGE_DEGREE_EMBARGO;
 import static nva.commons.apigateway.AccessRight.MANAGE_DOI;
 import static nva.commons.apigateway.AccessRight.MANAGE_PUBLISHING_REQUESTS;
 import static nva.commons.apigateway.AccessRight.MANAGE_RESOURCES_STANDARD;
@@ -36,6 +38,7 @@ public final class CuratorGrantStrategy extends PublicationStrategyBase implemen
             case PUBLISHING_REQUEST_APPROVE,
                  READ_HIDDEN_FILES -> canManagePublishingRequests();
             case SUPPORT_REQUEST_APPROVE -> hasAccessRight(SUPPORT);
+            case THESIS_APPROVAL -> hasAccessRight(MANAGE_DEGREE) || hasAccessRight(MANAGE_DEGREE_EMBARGO);
             case REPUBLISH, DELETE, TERMINATE -> false;
         };
     }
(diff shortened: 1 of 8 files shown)


publication-commons/src/test/java/no/unit/nva/publication/permissions/publication/CuratorPermissionStrategyTest.java
```diff
@@ -195,5 +196,22 @@ void shouldAllowCuratorOnDegreeWithResourceOwnerAffiliation(PublicationOperation
                                   .create(publication, userInstance)
                                   .allowsAction(operation));
     }
+
+    @Test
+    void shouldAllowCuratorToPerformThesisApprovalWhenDegreeAccessRightIsPresent()
```
joachimjorgensen, line 201: Does it enrich the test if we also test for failure?

publication-model/src/main/java/no/unit/nva/model/PublicationOperation.java
```diff
@@ -19,6 +19,7 @@ public enum PublicationOperation {
     DOI_REQUEST_APPROVE("doi-request-approve"),
     PUBLISHING_REQUEST_CREATE("publishing-request-create"),
     PUBLISHING_REQUEST_APPROVE("publishing-request-approve"),
+    THESIS_APPROVAL("thesis-approval"),
```
joachimjorgensen, line 22: Naming can be difficult :) The other operations is named `...-approve`, while you have `...-approval`. Intentional? 

If the the other case will be `file-approval`, this will be fine. I assume there will be one for file as well?

Or should it be `approve-thesis`?

publication-commons/src/main/java/no/unit/nva/publication/permissions/publication/grant/CuratorGrantStrategy.java
```diff
@@ -36,6 +38,7 @@ public boolean allowsAction(PublicationOperation permission) {
             case PUBLISHING_REQUEST_APPROVE,
                  READ_HIDDEN_FILES -> canManagePublishingRequests();
             case SUPPORT_REQUEST_APPROVE -> hasAccessRight(SUPPORT);
+            case THESIS_APPROVAL -> hasAccessRight(MANAGE_DEGREE) || hasAccessRight(MANAGE_DEGREE_EMBARGO);
```
joachimjorgensen, line 41: Should non-embargo-curator be able to approve embargo-files?

truhacevkir: This requirement comes from discussion with Terje. Degree curator and EmbargoDegree curator should be able to approve files for degrees. I mean EmbargoDegree access right is used for file update only.

publication-model/src/main/java/no/unit/nva/model/PublicationOperation.java
truhacevkir: It is intentional. To match the pattern it should be named as files-approval-thesis-approve, but it sound weird. File's allowed operations are unchanged and are. Files are approved via ticket. 
    READ_METADATA("read-metadata"),
    WRITE_METADATA("write-metadata"),
    DELETE("delete"),
    DOWNLOAD("download");

PR #2143, opened 2025-04-28
Np 48899 hard delete of drafts

ketilaa:


template.yaml
```diff
@@ -2373,6 +2373,41 @@ Resources:
...
+  EventForwardingRule:
+    Type: AWS::Events::Rule
+    Properties:
+      Name: ForwardFromInternalToNvaEventBusRule
+      EventBusName: !Ref InternalBus
+      EventPattern:
+        detail:
+          responsePayload:
+            topic:
+              - "PublicationService.Resource.Deleted"
```
brinxmat, line 2385: I actually checked the value against the java constant…do I get a prize? 🏆 You do, because they match!

brinxmat (commented):
Nice

brinxmat (approved):

publication-commons/src/main/java/no/unit/nva/publication/service/impl/ResourceService.java
```diff
@@ -618,13 +618,12 @@ private List<TransactWriteItem> deleteResourceTransactionItems(List<Dao> daos) t
     }
 
     private void applyDeleteResourceConditions(TransactWriteItem deleteResource) {
-        Map<String, String> expressionAttributeNames = Map.of("#status", STATUS_FIELD_IN_RESOURCE, "#doi",
-                                                              DOI_FIELD_IN_RESOURCE);
+        Map<String, String> expressionAttributeNames = Map.of("#status", STATUS_FIELD_IN_RESOURCE);
```
ketilaa, line 621: We now allow deleting non-published resource with doi. Deletion of drafted doi's will be handled automatically by event listener in doi-registrar.

publication-commons/src/main/java/no/unit/nva/publication/service/impl/ResourceService.java
```diff
@@ -618,13 +618,12 @@ private List<TransactWriteItem> deleteResourceTransactionItems(List<Dao> daos) t
...
         Map<String, AttributeValue> expressionAttributeValues = Map.of(":publishedStatus",
                                                                        new AttributeValue(PUBLISHED.getValue()));
 
         deleteResource.getDelete()
-            .withConditionExpression("#status <> :publishedStatus AND attribute_not_exists(#doi)")
+            .withConditionExpression("#status <> :publishedStatus")
```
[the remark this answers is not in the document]
ketilaa: I see the point. I will change the expression to explicitly check that the status is DRAFT

PR #2144, opened 2025-04-29
persisting files approval thesis on publication update

truhacevkir:
Adding similar logic which is lying in PublishingService to PublishingRequestResolver - persisting FilesApprovalThesis when updating publication with new pending file. 

For now we simply support cases when we add new or remove pending files from publication. Next step will be to support ticket transfer when publication channel is being updated to claimed one.

### DIFF
--- a/publication-rest/src/main/java/no/unit/nva/publication/update/PublishingRequestResolver.java
+++ b/publication-rest/src/main/java/no/unit/nva/publication/update/PublishingRequestResolver.java
@@ -190,8 +190,12 @@ public final class PublishingRequestResolver {
         return Optional.ofNullable(channelClaim)
                    .map(ChannelClaimDto::claimedBy)
                    .map(CustomerSummaryDto::id)
-                   .map(id -> userInstance.getCustomerId().equals(id))
+                   .map(id -> hasTheSameCustomer(userInstance, id))
                    .orElse(true);
+    }
+
+    private static Boolean hasTheSameCustomer(UserInstance userInstance, URI id) {
+        return Optional.ofNullable(userInstance).map(UserInstance::getCustomerId).map(id::equals).orElse(false);
     }
 
     private Optional<ChannelClaimDto> getChannelClaim(Publisher publisher) throws BadGatewayException {


publication-rest/src/main/java/no/unit/nva/publication/update/PublishingRequestResolver.java
```diff
@@ -115,33 +142,91 @@ private TicketEntry persistPublishingRequest(
...
+                    FilesApprovalThesis.create(resource, userInstance, organizationId, workflow)
+                        .withFilesForApproval(files)
+                        .persistNewTicket(ticketService);
+                    return;
+                }
+            }
+            FilesApprovalThesis.createForUserInstitution(resource, userInstance, workflow)
+                .withFilesForApproval(files)
+                .persistNewTicket(ticketService);
+        }
```
truhacevkir, line 166: Ugly, but for now not possible to refactor to make it cleaner.

publication-rest/src/main/java/no/unit/nva/publication/update/PublishingRequestResolver.java
```diff
@@ -88,18 +112,21 @@ private boolean thereAreNoPendingFiles(Publication publicationUpdate) {
...
         return resourceService
                    .fetchAllTicketsForResource(Resource.fromPublication(publication))
-                   .filter(PublishingRequestCase.class::isInstance)
-                   .map(PublishingRequestCase.class::cast)
-                   .filter(ticketEntry -> ticketEntry.hasSameOwnerAffiliationAs(userInstance))
+                   .filter(FilesApprovalEntry.class::isInstance)
+                   .map(FilesApprovalEntry.class::cast)
+                   .filter(ticketEntry -> Resource.fromPublication(publication).isDegree()
+                                              ? !ticketEntry.getFilesForApproval().isEmpty()
+                                              : ticketEntry.hasSameOwnerAffiliationAs(userInstance))
```
brinxmat, line 123: Is this just named method containing `(isDegree() && !ticketEntry.isEmpty()) || ticketEntry.hasSameOwnerAffiliationAs(userInstance)`?

publication-rest/src/main/java/no/unit/nva/publication/update/PublishingRequestResolver.java
```diff
@@ -115,33 +142,91 @@ private TicketEntry persistPublishingRequest(
...
     }
 
-    private void persistPendingPublishingRequest(Publication oldImage, Publication newImage) {
+    private void persistPendingPublishingRequest(Publication oldImage, Publication newImage)
+        throws ApiGatewayException {
         var files = getNewPendingFiles(oldImage, newImage).collect(Collectors.toSet());
         var resource = Resource.fromPublication(newImage);
         var workflow = lookUp(customer.getPublicationWorkflow());
+
+        if (resource.isDegree()) {
```
brinxmat, line 151: Climb every mountain.

publication-rest/src/main/java/no/unit/nva/publication/update/PublishingRequestResolver.java
```diff
@@ -115,33 +142,91 @@ private TicketEntry persistPublishingRequest(
...
                                     .createWithFilesForApproval(resource, userInstance, workflow, files);
 
         attempt(() -> persistPublishingRequest(newImage, publishingRequest));
     }
 
+    private boolean isClaimedByUserOrganization(ChannelClaimDto channelClaim, UserInstance userInstance) {
+        return Optional.ofNullable(channelClaim)
+                   .map(ChannelClaimDto::claimedBy)
+                   .map(CustomerSummaryDto::id)
+                   .map(id -> userInstance.getCustomerId().equals(id))
```
brinxmat, line 177: Can customerId from user instance be null?

brinxmat (changes requested):

publication-rest/src/main/java/no/unit/nva/publication/update/PublishingRequestResolver.java
truhacevkir: Do not think so. We consume `UserInstance` from Authorized user which performs `/update` on publication

publication-commons/src/main/java/no/unit/nva/publication/model/FilesApprovalEntry.java
```diff
@@ -91,8 +93,8 @@ public Set<File> getFilesForApproval() {
         return nonNull(filesForApproval) ? filesForApproval : Set.of();
     }
 
-    public void setFilesForApproval(Set<File> filesForApproval) {
-        this.filesForApproval = filesForApproval;
+    public void setFilesForApproval(Collection<File> filesForApproval) {
```
brinxmat, line 96: Because we have done this and there is only a zero-arg constructor (implicit?).

I'd actually consider initialising the Set in the field and just call `addAll` on this in the setter to avoid nulls.

And replace:
```
    public Set<File> getFilesForApproval() {
        return nonNull(filesForApproval) ? filesForApproval : Set.of();
    }
```

with

```
    public Set<File> getFilesForApproval() {
        return new HashSet<>(filesForApproval);
    }
``` 

This will then prevent mutating the object accidentally in subsequent ops.

This style makes immutable objects, but it might require a helper-constructor for people who want to mutate things (if they do this) so they can chain this object to another i.e. this.copy().setFields(old + new)

brinxmat (approved):

publication-rest/src/main/java/no/unit/nva/publication/update/PublishingRequestResolver.java
brinxmat: I also think "think", which indicates we should always use Optional to indicate things may be null in EVERY case very, very strictly.

publication-commons/src/main/java/no/unit/nva/publication/model/FilesApprovalEntry.java
truhacevkir: If setFiles...() will use addAll() method, we should change whole implementation of how these collections are used and mutated in production code. Cause instead of overwriting collection as we do now, we should then both remove items and add items to it, when file is being added to or removed from publication.

PR #2149, opened 2025-05-05
removing persistence of UnpublishRequest

truhacevkir:
UnpublishRequest is never used. Removing persistence of it. In future we should remove all appearances of it (dao, expansion, search..)


brinxmat (approved):
Nice! You have important knowledge here that is important to maintain. Can you indicate what you write in the PR-description in some way in the code (e.g. with `@deprecated` and a comment) and a task in the ticket system?

PR #2152, opened 2025-05-05
add empty publication channels to resource

truhacevkir:


publication-commons/src/main/java/no/unit/nva/publication/model/business/Resource.java
```diff
@@ -245,6 +248,14 @@ && getStatus() == resource.getStatus()
                  && Objects.equals(getImportDetails(), resource.getImportDetails()));
     }
 
+    public void setPublicationChannels(List<PublicationChannel> publicationChannels) {
+        this.publicationChannels = publicationChannels;
+    }
+
+    public List<PublicationChannel> getPublicationChannels() {
+        return publicationChannels;
```
joachimjorgensen, line 256: Either have a null check here, or add a List.of() in constructor?

publication-commons/src/main/java/no/unit/nva/publication/model/business/Resource.java
```diff
@@ -245,6 +248,14 @@ && getStatus() == resource.getStatus()
                  && Objects.equals(getImportDetails(), resource.getImportDetails()));
     }
 
+    public void setPublicationChannels(List<PublicationChannel> publicationChannels) {
```
joachimjorgensen, line 251: Add method for adding single claim? And remove single claim? Or is it better to replace the list?

joachimjorgensen (approved):

publication-commons/src/main/java/no/unit/nva/publication/model/business/Resource.java
truhacevkir: I think we can add these methods when we will use them :)

PR #2156, opened 2025-05-06
Add curating institutions to publication response for elavated users

joachimjorgensen:


truhacevkir (approved):
Nice, we should remember to remove this code :)

PR #2159, opened 2025-05-07
adding functionality for listing publication channels

truhacevkir:
- Adding functionality for listing publication channels
- Fixing some logic when consuming channel identifier and crafting ChannelType from it.

### DIFF
--- a/publication-commons/src/test/java/no/unit/nva/publication/model/business/publicationchannel/PublicationChannelTest.java
+++ b/publication-commons/src/test/java/no/unit/nva/publication/model/business/publicationchannel/PublicationChannelTest.java
@@ -4,6 +4,7 @@ import static no.unit.nva.model.testing.PublicationGenerator.randomPublication;
 import static no.unit.nva.model.testing.PublicationGenerator.randomUri;
 import static no.unit.nva.testutils.RandomDataGenerator.randomInteger;
 import static no.unit.nva.testutils.RandomDataGenerator.randomString;
+import static nva.commons.core.attempt.Try.attempt;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.mockito.Mockito.mock;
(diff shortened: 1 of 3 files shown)


publication-commons/src/test/java/no/unit/nva/publication/model/business/publicationchannel/PublicationChannelTest.java
```diff
@@ -163,6 +165,38 @@ void shouldThrowExceptionWhenIdentityServiceRespondsWithUnhandledError()
                                        .persistNew(resourceService, UserInstance.fromPublication(publication)));
     }
 
+    @Test
+    void shouldReturnListingResultWhenListingPublicationChannels()
```
brinxmat, line 169: What happens when there aren't any? Or there are many more than 1MB?

joachimjorgensen (approved):

publication-commons/src/test/java/no/unit/nva/publication/model/business/publicationchannel/PublicationChannelTest.java
truhacevkir: I thought about it, but I mean we do not have to provide some number to limit number of items to be scanned/queried, as I understand -> Dynamo will stop scanning when reaching 1 MB limit, not depending on what size limit is sent in request?

```
Limit:
The maximum number of items to evaluate (not necessarily the number of matching items).
If DynamoDB processes the number of items up to the limit while processing the results, 
it stops the operation and returns the matching values up to that point, 
and a key in LastEvaluatedKey to apply in a subsequent operation, so that you can pick up where you left off. 
Also, if the processed dataset size exceeds 1 MB before DynamoDB reaches this limit, 
it stops the operation and returns the matching values up to the limit, 
and a key in LastEvaluatedKey to apply in a subsequent operation to continue the operation. 
```
[](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_Query.html)

truhacevkir: [](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_Query.html)

brinxmat (approved):

PR #2165, opened 2025-05-08
Add migration code for adding publication channels to degrees with publisher

joachimjorgensen:
Will overwrite any existing publication channels.

### DIFF
--- a/publication-commons/src/main/java/no/unit/nva/publication/service/impl/ResourceService.java
+++ b/publication-commons/src/main/java/no/unit/nva/publication/service/impl/ResourceService.java
@@ -7,6 +7,7 @@ import static no.unit.nva.model.PublicationStatus.PUBLISHED;
 import static no.unit.nva.model.PublicationStatus.UNPUBLISHED;
 import static no.unit.nva.publication.model.business.Resource.resourceQueryObject;
 import static no.unit.nva.publication.model.business.publicationchannel.PublicationChannelUtil.createPublicationChannelDao;
+import static no.unit.nva.publication.model.business.publicationchannel.PublicationChannelUtil.toChannelClaimUri;
 import static no.unit.nva.publication.model.storage.DynamoEntry.parseAttributeValuesMap;
 import static no.unit.nva.publication.service.impl.ReadResourceService.RESOURCE_NOT_FOUND_MESSAGE;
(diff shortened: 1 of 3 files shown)


publication-commons/src/test/java/no/unit/nva/publication/service/impl/MigrationTests.java
```diff
@@ -172,6 +190,86 @@ void shouldWorkWithEmptyEntityDescription() {
...
+    }
+
+    @Test
+    void shouldNotPersistNonClaimedPublicationChannelWhenNonDegreeWithPublisher() throws NotFoundException {
+        var publisherId = randomPublisherId(randomUUID());
+        var publisher = new Publisher(publisherId);
+
+        var hardCodedIdentifier = new SortableIdentifier("0183892c7413-af720123-d7ae-4a97-a628-a3762faf8438");
+        var publication = randomPublication(Textbook.class);
+        publication.getEntityDescription().getReference().setPublicationContext(reportWithPublisher(publisher));
```
joachimjorgensen, line 241: TextBook with context Report, not sure if that is a real case. Just want to test a publication that has publisher but is not a degree.

publication-commons/src/main/java/no/unit/nva/publication/service/impl/ResourceService.java
```diff
@@ -305,9 +308,29 @@ public Resource updateResource(Resource resource, UserInstance userInstance) {
...
     // update this method according to current needs.
     public Entity migrate(Entity dataEntry) {
+        if (dataEntry instanceof Resource resource) {
+            persistPublicationChannelForPublisherWhenDegree(resource);
+        }
         return dataEntry;
     }
 
+    private void persistPublicationChannelForPublisherWhenDegree(Resource resource) {
+        resource.getPublisherWhenDegree().ifPresent(publisher -> {
```
brinxmat, line 318: Do you write the Resource to mark it as "refreshed"?

truhacevkir: Version will be updated when Resource will be refreshed/ persisted back in `EventBasedBatchScanHandler`.

truhacevkir (approved):
Looks good, but i would test it in sandbox before merging.

PR #2175, opened 2025-05-14
using uriRetriever instead of identityService client when fetching claim

truhacevkir:


joachimjorgensen (approved):

publication-commons/src/main/java/no/unit/nva/publication/external/services/ChannelClaimClient.java
```diff
@@ -0,0 +1,55 @@
...
+        if (response.statusCode() != HttpStatusCode.OK) {
+            throw new IllegalStateException("Received " + response.statusCode() + " from identity service");
+        }
+        return response;
+    }
+
+    private NotFoundException handleFailure(Failure<?> responseFailure) {
+        var exception = responseFailure.getException();
+        if (exception instanceof NotFoundException) {
+            return new NotFoundException(exception);
```
torbjokv, line 50: return exception

publication-commons/src/main/java/no/unit/nva/publication/external/services/ChannelClaimClient.java
```diff
@@ -0,0 +1,55 @@
...
+        }
+
+        throw new RuntimeException("Something went wrong!");
```
torbjokv, line 53: maybe add , exception to here?

PR #2182, opened 2025-05-15
Implement incremental count on publication creation #NP-47658

torbjokv:
This PR is based in a branch from Kir :-) thanks

If the post does not have a cristin id, on creation, this code will inject a syntetic one starting on 10 000 000 and upwards (cristin is on id 2 xxx xxx xxx today). Later we need to add code for migration and etc.


brinxmat (approved):

torbjokv:
> probably need to handle this with an exception in DataEntryUpdateEvent @truhacevkir ?

we've discussed and since batchscan uses ByCustomerResource(PK2, SK2) and this only populate PK/SK0, it should be fine

PR #2192, opened 2025-05-19
feature(resource-service): add synthetic cristin id on existing resou…

torbjokv:
…rces #NP-47659

### DIFF
--- a/publication-commons/src/main/java/no/unit/nva/publication/model/storage/ResourceDao.java
+++ b/publication-commons/src/main/java/no/unit/nva/publication/model/storage/ResourceDao.java
@@ -49,6 +49,7 @@ public class ResourceDao extends Dao
     implements JoinWithResource, JsonSerializable, DynamoEntryByIdentifier {
     
     public static final String CRISTIN_SOURCE = "Cristin";
+    private static final String NVA_SOURCE = "nva";
     public static final String TYPE = "Resource";
     private static final String BY_RESOURCE_INDEX_ORDER_PREFIX = "a";
     private static final String STATUS_FIELD = "status";
@@ -245,11 +246,15 @@ public class ResourceDao extends Dao
     }
 
     private boolean isCristinIdentifier(AdditionalIdentifierBase identifier) {
-        return nonNull(identifier) && identifier instanceof CristinIdentifier && isCristinSource(identifier);
+        return nonNull(identifier) && identifier instanceof CristinIdentifier && (isCristinSource(identifier) || isNvaSource(identifier));
     }
 
     private static boolean isCristinSource(AdditionalIdentifierBase identifier) {
         return identifier.sourceName().toLowerCase(Locale.ROOT).contains(CRISTIN_SOURCE.toLowerCase(Locale.ROOT));
+    }
+
+    private static boolean isNvaSource(AdditionalIdentifierBase identifier) {
+        return identifier.sourceName().toLowerCase(Locale.ROOT).contains(NVA_SOURCE);
     }
 
     //TODO: All AdditionalIdentifiers with Cristin source should be migrated to CristinIdentifier's
--- a/publication-commons/src/main/java/no/unit/nva/publication/model/storage/ResourceDao.java
+++ b/publication-commons/src/main/java/no/unit/nva/publication/model/storage/ResourceDao.java
@@ -229,11 +229,15 @@ public class ResourceDao extends Dao
                                                                                           String source) {
         return getResource().getAdditionalIdentifiers().stream().filter(identifierType::isInstance)
                    .map(identifierType::cast)
-                   .filter(identifier -> identifier.sourceName()
-                                             .toLowerCase(Locale.ROOT)
-                                             .contains(source.toLowerCase(Locale.ROOT)))
+                   .filter(identifier -> containsIgnoringCase(identifier.sourceName(), source))
                    .map(AdditionalIdentifierBase::value)
                    .findFirst();
+    }
+
+    private static boolean containsIgnoringCase(String input, String search) {
+        return input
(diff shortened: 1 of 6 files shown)


torbjokv:
wait until @joachimjorgensen and @truhacevkir have migrated prod before taking this in

publication-commons/src/main/java/no/unit/nva/publication/model/storage/ResourceDao.java
```diff
@@ -187,76 +189,64 @@ public List<TicketDao> fetchAllTickets(AmazonDynamoDB client) {
...
-        String cristinIdentifierValue = Optional.ofNullable(getResource().getAdditionalIdentifiers())
-                                            .stream()
-                                            .flatMap(Collection::stream)
-                                            .filter(this::keyEqualsCristin)
-                                            .map(AdditionalIdentifierBase::value)
-                                            .collect(SingletonCollector.collectOrElse(null));
-        return Optional.ofNullable(cristinIdentifierValue);
+        return getAdditionalIdentifier(CristinIdentifier.class, CRISTIN_SOURCE)
+                   .or(() -> getAdditionalIdentifier(AdditionalIdentifier.class, CRISTIN_SOURCE))
+                   .or(() -> getAdditionalIdentifier(CristinIdentifier.class, BRAGE_SOURCE))
```
torbjokv, line 222: So, this is a cristin-id orginating from brage. Is it better or worse to have than not have? @truhacevkir @brinxmat this is mostly used for import for now, so in terms of import, how bad would a wrongly user entered value be here vs giving up all correctly entered values. I guess its ok? all values is still saved to identifiers list, so we can redo this later with a new batch scan.

torbjokv: btw this output is used to populate PK4

brinxmat: IIRC, we have seen bad values from Brage. I don't think it makes any sense at all to include identifiers from Brage as if they are from Cristin. If they are from Cristin, then they will be joined in a record with the correct identifier from Cristin — right, @truhacevkir? We shouldn't accept hearsay :)

brinxmat: It strikes me that we should not call this value "CristinIdentifier" because it isn't what it is — it is an identifier from a numeric sequence…some of which existed in a system somewhere in the distant past…. It ceased to be a CristinIdentifier the second we inserted the data into a key role in the DB

publication-commons/src/main/java/no/unit/nva/publication/model/storage/ResourceDao.java
```diff
@@ -187,76 +189,64 @@ public List<TicketDao> fetchAllTickets(AmazonDynamoDB client) {
...
+                   .or(() -> getAdditionalIdentifier(CristinIdentifier.class, NVA_SOURCE));
+    }
+
+    private <T extends AdditionalIdentifierBase> Optional<String> getAdditionalIdentifier(Class<T> identifierType,
+                                                                                                  String source) {
+        return getResource().getAdditionalIdentifiers().stream().filter(identifierType::isInstance)
+                   .map(identifierType::cast)
+                   .filter(identifier -> identifier.sourceName()
```
brinxmat, line 230: pull filter out "containsIgnoringCase"

publication-commons/src/main/java/no/unit/nva/publication/model/storage/ResourceDao.java
```diff
@@ -187,76 +189,70 @@ public List<TicketDao> fetchAllTickets(AmazonDynamoDB client) {
...
     }
-    
+
     @JsonProperty(RESOURCES_BY_CRISTIN_ID_INDEX_PARTITION_KEY_NAME)
     public String getResourceByCristinIdentifierPartitionKey() {
         return extractCristinIdentifier().isEmpty() ? null
-                   : CRISTIN_IDENTIFIER_INDEX_FIELD_PREFIX + KEY_FIELDS_DELIMITER + extractCristinIdentifier();
+                   : CRISTIN_IDENTIFIER_INDEX_FIELD_PREFIX
+                     + KEY_FIELDS_DELIMITER
+                     + extractCristinIdentifier().orElseThrow();
```
torbjokv, line 198: @truhacevkir I hope moving from CristinIdentifier:Optional[10000026] to CristinIdentifier:10000026 in DB does not imply problems in import :-)

publication-commons/src/main/java/no/unit/nva/publication/model/storage/ResourceDao.java
torbjokv: yes, at least on the "outside" CristinIdentifier should not be exposed as is, but do we want to change it now?, but on the inside of our system cristin will probably live for a while, so I think it is ok within our services and confusing to call it something else. While the synthetic identifiers is not really a cristin identifiler, we still want it to behave as one, so maaaaybe ok? Our goal is hopefully to remove this ID at some point in time

brinxmat (approved):

publication-commons/src/main/java/no/unit/nva/publication/model/storage/ResourceDao.java
truhacevkir: We use "CristinIdentifier" from brage when looking up for existing publication. For now we first create Publication with CristinIdentifier when importing from Brage, and then using "CristinIdentifier" to look up for existing publication. I dont think we want to change it now, sprint before we will migrate :)

truhacevkir (approved):

PR #2198, opened 2025-05-21
NP-49190 Add FilesApprovalThesis to create meassage handler

joachimjorgensen:


messages/src/main/java/no/unit/nva/publication/messages/create/NewCreateMessageHandler.java
```diff
@@ -80,7 +80,7 @@ private static boolean userHasPermissionToCreateMessageForTicket(PublicationPerm
                                                                      TicketEntry ticketDto) {
         return switch (ticketDto) {
             case DoiRequest ignored -> permissionStrategy.allowsAction(DOI_REQUEST_CREATE);
-            case PublishingRequestCase ignored -> permissionStrategy.allowsAction(PUBLISHING_REQUEST_CREATE);
+            case FilesApprovalEntry ignored -> permissionStrategy.allowsAction(PUBLISHING_REQUEST_CREATE);
```
truhacevkir, line 83: It should you use `approve-files` allowed operation

truhacevkir (approved):

messages/src/main/java/no/unit/nva/publication/messages/create/NewCreateMessageHandler.java
joachimjorgensen: You sure? This means that contributor cannot write messages

PR #2200, opened 2025-05-22
fixing equals method for unconfirmed funding

truhacevkir:


publication-model/src/main/java/no/unit/nva/model/funding/UnconfirmedFunding.java
```diff
@@ -67,26 +65,21 @@ public Instant getActiveTo() {
...
     @Override
-    public boolean equals(Object o) {
-        if (this == o) {
-            return true;
-        }
-        if (o == null || getClass() != o.getClass()) {
-            return false;
-        }
-        UnconfirmedFunding unconfirmedFunding = (UnconfirmedFunding) o;
-        return source.equals(unconfirmedFunding.source)
```
truhacevkir, line 78: source.equals(... throw exception when null.

torbjokv (approved):

PR #2205, opened 2025-05-22
Log number of publications found for channel with batch update

joachimjorgensen:


truhacevkir (approved):

publication-event-handlers/src/main/java/no/unit/nva/publication/events/handlers/batch/PublicationChannelsBatchUpdateService.java
```diff
@@ -26,6 +27,8 @@ public void updateChannels(ChannelUpdateEvent event) {
         var identifier = event.getChannelIdentifier();
         var publicationChannels = listAllPublicationChannelsWithIdentifier(identifier);
 
+        logger.info(FOUND_PUBLICATIONS_WITH_CHANNEL_MESSAGE, publicationChannels.size(), identifier);
```
truhacevkir, line 30: I would log all identifiers, so you are sure that the one you are missing is not present in list

PR #2211, opened 2025-05-26
updating metadata imported for event from brage

truhacevkir:
When importing Event from Brage, setting Agent as organization and Event name as main title.


brage-import/src/main/java/no/sikt/nva/brage/migration/mapper/PublicationContextMapper.java
```diff
@@ -282,8 +283,11 @@ private static boolean isReport(Record brageRecord) {
         return NvaType.REPORT.getValue().equals(brageRecord.getType().getNva());
     }
 
-    private static PublicationContext buildPublicationContextWhenEvent() {
-        return new Event.Builder().build();
+    private static PublicationContext buildPublicationContextWhenEvent(Record brageRecord, URI organizationId) {
+        return new Event.Builder()
+                   .withName(extractMainTitle(brageRecord))
+                   .withAgent(Organization.fromUri(organizationId))
+                   .build();
```
truhacevkir, line 290: Not sure if this is right thing to do :)

brinxmat (approved):

PR #2212, opened 2025-05-26
Np 49166 allow student thesis on publications without files through channel ownership

joachimjorgensen:
Channel ownership should grant student thesis curator update access to degrees even if there are no approved files.

Created a specific and minimal test suite, as the entire Context needs to be refactored. We have some false positives.

### DIFF
--- a/publication-commons/src/test/java/cucumber/permissions/publication/PublicationAccessFeatures.java
+++ b/publication-commons/src/test/java/cucumber/permissions/publication/PublicationAccessFeatures.java
@@ -88,4 +88,9 @@ public class PublicationAccessFeatures {
     public void theUserIsFromTheSameOrganizationAsClaimedPublisher() {
         scenarioContext.setUserOrganization(ORGANIZATION);
     }
+
+    @And("publication has no approved files")
+    public void publicationHasNoApprovedFiles() {
+        scenarioContext.setIsMetadataOnly(true);
+    }
 }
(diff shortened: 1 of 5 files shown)


publication-commons/src/test/resources/features/publication/publication_update_partial.feature
```diff
@@ -78,8 +78,8 @@ Feature: Publication update permissions
       | Degree file curator     | MetadataOnly,Degree,Imported | update         | Allowed     |
       | Everyone else           | MetadataOnly,Degree,Imported | update         | Not Allowed |
       | Related external client | MetadataOnly,Degree,Imported | update         | Allowed     |
-      | Publication owner       | MetadataOnly,Degree          | update         | Not Allowed |
-      | Contributor             | MetadataOnly,Degree          | update         | Not Allowed |
+      | Publication owner       | MetadataOnly,Degree          | update         | Allowed     |
+      | Contributor             | MetadataOnly,Degree          | update         | Allowed     |
```
joachimjorgensen, line 82: False positive due to the publication being set up as an student thesis import.

publication-commons/src/test/java/cucumber/permissions/publication/PublicationAccessFeatures.java
```diff
@@ -88,4 +88,9 @@ public void publisherIsClaimedByOrganization() {
     public void theUserIsFromTheSameOrganizationAsClaimedPublisher() {
         scenarioContext.setUserOrganization(ORGANIZATION);
     }
+
+    @And("publication has no approved files")
+    public void publicationHasNoApprovedFiles() {
+        scenarioContext.setIsMetadataOnly(true);
```
joachimjorgensen, line 94: Should ideally be a context method that adds a non approved file to the associated artifacts list, as this one sets an empty list. But that's a bigger refactoring.

ketilaa (approved):

PR #2214, opened 2025-05-28
persisting recovery message when publication channel not found under expansion

truhacevkir:
- Moving RecoveryEntry to commons
- Persisting RecoveryEntry when fetching publication channel under expansion return 404 (still expanding whole resource and sending forward)
- Adding test case for redirects


truhacevkir:
This will make it possible to find publications with publication channels that does not resolve and update them.

PR #2224, opened 2025-06-02
Support manuall update of serial publication

truhacevkir:
Adding support for updating serial publications in ManuallyUpdatePublicationHandler.
Will use it to fix all serial publications that redirect in test.

### DIFF
--- a/publication-event-handlers/src/main/java/no/unit/nva/publication/events/handlers/batch/ManuallyUpdatePublicationUtil.java
+++ b/publication-event-handlers/src/main/java/no/unit/nva/publication/events/handlers/batch/ManuallyUpdatePublicationUtil.java
@@ -34,8 +34,20 @@ public final class ManuallyUpdatePublicationUtil {
 
     private void updateSeriesOrJournal(List<Resource> resources, ManuallyUpdatePublicationsRequest request) {
         resources.stream()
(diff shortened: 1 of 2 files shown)


publication-event-handlers/src/main/java/no/unit/nva/publication/events/handlers/batch/ManuallyUpdatePublicationUtil.java
```diff
@@ -25,10 +27,33 @@ public static ManuallyUpdatePublicationUtil create(ResourceService resourceServi
...
             case PUBLISHER -> updatePublisher(resources, request);
+            case SERIAL_PUBLICATION -> updateSeriesOrJournal(resources, request);
             case LICENSE -> updateLicense(resources, request);
         }
     }
 
+    private void updateSeriesOrJournal(List<Resource> resources, ManuallyUpdatePublicationsRequest request) {
+        resources.stream()
+            .map(resource -> updateSeriesOrJournal(resource, request.oldValue(), request.newValue()))
+            .forEach(resource -> resourceService.updateResource(resource, UserInstance.fromPublication(resource.toPublication())));
```
torbjokv, line 38: maybe do a check where you only update changed resources?

torbjokv (approved):
See suggestions, otherwise good :-)

PR #2225, opened 2025-06-03
fixing refresh method

truhacevkir:
Updating refresh resource method. This method has been using resourceService.updatePublication() method which got hasEffectiveChanges() check. So recovery batch scan has never updated database entries for resources since then.


brinxmat (approved):
Linter on ResourceServiceTest and merge

PR #2230, opened 2025-06-03
updating finder rules

truhacevkir:
Updating finer rules in Brage migration with following changes:

- When searching for duplicate given types is Journal or UnconfirmedJournal, we use both types as search params. 
- Not comparing types when selecting "merge candidates" fetched by CristinIdentifier, Handler and DOI. 
- Updating order of "Finders" which search for merging candidates:
Prioritizing CristinFinder, HandlerFinder and DoiFinder.


brage-import/src/main/java/no/sikt/nva/brage/migration/merger/findexistingpublication/DoiPublicationFinder.java
```diff
@@ -45,7 +45,7 @@ private Optional<PublicationForUpdate> existingPublicationHasSameDoi(
 
         var publicationsByDoi = searchApiFinder.fetchPublicationsByParam(DOI, doi.toString()).stream()
                                     .filter(item ->
-                                                PublicationComparator.publicationsMatch(item,
+                                                PublicationComparator.publicationsMatchIgnoringType(item,
```
brinxmat, line 48: I'd add a not why Type isn't interesting on the method in PubComp

brinxmat (approved):

PR #2235, opened 2025-06-03
NP-49254 Use instance type instead of context when fetching publisher for degree

joachimjorgensen:
There have been some (erroneous?) instances where publication context have been Book and instance type have been a degree type. Instead of checking if publication is of context Degree, the publication have to be checked against the list of degree instance types.


publication-event-handlers/src/test/java/no/unit/nva/publication/events/handlers/batch/UpdatePublicationsInBatchesHandlerTest.java
```diff
@@ -209,8 +210,10 @@ private static InputStream createEvent(ManualUpdateType type, String oldValue, S
 
     private static URI createChannelIdWithIdentifier(String channelIdentifier) {
```
joachimjorgensen, line 211: Not sure why this hasn't failed previously.

PR #2236, opened 2025-06-03
fix(UpdateTicketHandler): Previously moved ticket fails with 403 when…

torbjokv:
… trying to approve as new receiver #NP-49260


ketilaa (approved):
No tests affected?

torbjokv:
> No tests affected?

Nope 😁 probably should have been

PR #2245, opened 2025-06-06
using nva customers only when crafting curating institutions

truhacevkir:
When crafting curating institutions, using contributors organizations affiliated with NVA customer.

### DIFF
--- a/publication-commons/src/main/java/no/unit/nva/publication/model/utils/CustomerService.java
+++ b/publication-commons/src/main/java/no/unit/nva/publication/model/utils/CustomerService.java
@@ -1,0 +1,32 @@
+package no.unit.nva.publication.model.utils;
+
+import static nva.commons.core.attempt.Try.attempt;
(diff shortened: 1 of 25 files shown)


publication-commons/src/main/java/no/unit/nva/publication/model/utils/CustomerService.java
```diff
@@ -0,0 +1,32 @@
...
+    private static final String CONTENT_TYPE = "application/json";
+
+    private final RawContentRetriever uriRetriever;
+
+    public CustomerService(RawContentRetriever rawContentRetriever) {
+        this.uriRetriever = rawContentRetriever;
+    }
+
+    public CustomerList fetchCustomers() {
+        var uri = UriWrapper.fromHost(new Environment().readEnv("API_HOST")).addChild("customer").getUri();
```
brinxmat, line 21: Maybe inject this environment in the constructor, but in a new PR.

brinxmat (approved):

PR #2248, opened 2025-06-10
Np 49261 cascading delete

ketilaa:


publication-commons/src/main/java/no/unit/nva/publication/service/impl/ResourceService.java
```diff
@@ -321,6 +321,27 @@ public Entity migrate(Entity dataEntry) {
         return dataEntry;
     }
 
+    public List<FileEntry> fetchFileEntriesForResource(Resource resource) {
```
torbjokv, line 324: we need this function in addition to getFileEntries() because the dynamodbevent does not include file information and it is not possible to look up the resource the the service after its deleted

publication-commons/src/main/java/no/unit/nva/publication/service/impl/ResourceService.java
```diff
@@ -321,6 +321,27 @@ public Entity migrate(Entity dataEntry) {
...
+        var queryRequest = new QueryRequest()
+                               .withTableName(tableName)
+                               .withIndexName(BY_TYPE_AND_IDENTIFIER_INDEX_NAME)
+                               .withKeyConditionExpression("PK3 = :resourceIdentifier AND begins_with(SK3, :type)")
+                               .withExpressionAttributeValues(
+                                   Map.of(
+                                       ":resourceIdentifier", new AttributeValue().withS(partitionKeyValue),
+                                       ":type", new AttributeValue().withS(FileDao.TYPE)
+                                   )
+                               );
```
brinxmat, line 336: Extract method that tells us what this query does. If I understand it: `createQueryForFilesAssociatedWithResource`

publication-commons/src/main/java/no/unit/nva/publication/service/impl/ResourceService.java
brinxmat: I didn't look, but are we consuming the list and doing List::toStream, or is the list being used as a list?

brinxmat (approved):

PR #2259, opened 2025-06-13
fix(JournalArticle): change of instance type was not detected because…

torbjokv:
… of issue in equals #NP-49123

### DIFF
--- a/publication-model/src/main/java/no/unit/nva/model/instancetypes/journal/JournalArticle.java
+++ b/publication-model/src/main/java/no/unit/nva/model/instancetypes/journal/JournalArticle.java
@@ -89,7 +89,6 @@ public class JournalArticle implements PublicationInstance<Range> {
 
     @Override
     public int hashCode() {
-        var clazz = getClass();
-        return Objects.hash(clazz.getName(), getPages(), getVolume(), getIssue(), getArticleNumber());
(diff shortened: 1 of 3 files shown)


publication-model/src/main/java/no/unit/nva/model/instancetypes/journal/JournalArticle.java
```diff
@@ -77,24 +76,20 @@ public Range getPages() {
...
+               && Objects.equals(getIssue(), that.getIssue())
+               && Objects.equals(getArticleNumber(), that.getArticleNumber());
     }
 
     @Override
-    @JacocoGenerated
     public int hashCode() {
-        return Objects.hash(getPages(), getVolume(), getIssue(), getArticleNumber());
+        var clazz = getClass();
+        return Objects.hash(clazz.getName(), getPages(), getVolume(), getIssue(), getArticleNumber());
```
brinxmat, line 93: I don't think that clazz.getName() is necessary :D

brinxmat (changes requested):

torbjokv:
> Looks good, but I think it is wrong that we test that type of things on handler level.

I agree, maybe this is the kind of test that I could delete after it have done its job. Not sure if it brings a lot of value to later

PR #2260, opened 2025-06-13
NP-49282 Update test scenarios for publication permissions when status is DRAFT

joachimjorgensen:


publication-model/src/main/java/no/unit/nva/model/PublicationOperation.java
```diff
@@ -10,6 +10,7 @@ public enum PublicationOperation {
 
     UPDATE("update"),
     PARTIAL_UPDATE("partial-update"),
+    @Deprecated
```
joachimjorgensen, line 13: `UPDATE_FILES("update-including-files")` seems to be an old allowedOperation that was replaced by allowedOperations on files.

ketilaa (approved):

PR #2264, opened 2025-06-16
NP-49282 Add Gherkin scenarios for permission strategies for channel claims

joachimjorgensen:


truhacevkir (approved):
Looks good, I think gherkin and readability will be better/more flexible if we instead of writing "users institution..." and in future maybe like "contributors institution...", "curators institution.." write something like:

```
And the publication's publisher is claimed by the institution "<Institution>"
And the user at "<Institution>" has the role "<UserRole>"
```

Instead of:

```
And publication has publisher claimed by "users institution"
And the user have the role "<UserRole>"
```

At the same time it adds two extra columns for these parameters. Not sure what is better.

joachimjorgensen:
> Looks good, I think gherkin and readability will be better/more flexible if we instead of writing "users institution..." and in future maybe like "contributors institution...", "curators institution.." write something like:
> 
> ```
> And the publication's publisher is claimed by the institution "<Institution>"
> And the user at "<Institution>" has the role "<UserRole>"
> ```
> 
> Instead of:
> 
> ```
> And publication has publisher claimed by "users institution"
> And the user have the role "<UserRole>"
> ```
> 
> At the same time it adds two extra columns for these parameters. Not sure what is better.

Good point, it is kind of "clunky" to read it the way I've written it. But it makes the context creation much easier to have it like this, so that we don't have to duplicate every institution with roles. And in this scenario, which institution you are from isn't the interesting part, just whether your institution has claimed the channel. For reference, in other scenarios where institution is important; "creating institution", "curating institution", and "non curating institution" are used (but I'm not really fond of the naming of the first one).

PR #2285, opened 2025-06-30
Adding dlq to dynamodb stream consumer

truhacevkir:
Adding dlq config to DynamoDbStream consumer as we did in identity-service. This will allow us to see if events are timing out during cristin-import and batch scan jobs

### DIFF
--- a/template.yaml
+++ b/template.yaml
@@ -1911,7 +1911,6 @@ Resources:
       Environment:
         Variables:
           EVENTS_BUCKET: !Ref NvaEventsBucketsName
-          DLQ_URL: !Ref DynamodbEventFanoutStreamRecordsDLQ
           MAX_ATTEMPT: !Ref EventBridgeMaxAttempt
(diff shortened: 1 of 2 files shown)


template.yaml
```diff
@@ -1924,6 +1923,11 @@ Resources:
             StartingPosition: TRIM_HORIZON
             BatchSize: 1
             Enabled: true
+            MaximumRetryAttempts: 3
+            BisectBatchOnFunctionError: true
+            DestinationConfig:
+              OnFailure:
+                Destination: !GetAtt DynamodbEventFanoutStreamRecordsDLQ.Arn
```
truhacevkir, line 1931: Not setting max event age as events after batch scan and during cristin-import can be up to 12h old.

LarsV123 (approved):
Should we do the same for `ImportCandidateDynamodbStreamToEventBridgeHandler`?

PR #2286, opened 2025-06-30
fix wrong institution is being set on merged resource event

truhacevkir:
When merging existing publication with incoming publication
Then MergedResourceEvent should be persisted with institution from incoming publication.

### DIFF
--- a/publication-commons/src/main/java/no/unit/nva/publication/model/business/Resource.java
+++ b/publication-commons/src/main/java/no/unit/nva/publication/model/business/Resource.java
@@ -426,8 +426,7 @@ public class Resource implements Entity {
         return resourceService.importResource(this, importSource);
     }
 
-    public Resource updateResourceFromImport(ResourceService resourceService, ImportSource importSource) {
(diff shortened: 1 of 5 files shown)


publication-commons/src/main/java/no/unit/nva/publication/model/business/Resource.java
```diff
@@ -426,8 +426,7 @@ public Resource importResource(ResourceService resourceService, ImportSource imp
         return resourceService.importResource(this, importSource);
     }
 
-    public Resource updateResourceFromImport(ResourceService resourceService, ImportSource importSource) {
-        var userInstance = UserInstance.fromPublication(this.toPublication());
+    public Resource updateResourceFromImport(ResourceService resourceService, ImportSource importSource, UserInstance userInstance) {
         this.setResourceEvent(MergedResourceEvent.fromImportSource(importSource, userInstance, Instant.now()));
```
truhacevkir, line 430: This is the main change. UserInstance.fromPublication() serves us institution from existing publication, cause this is the value which is never updated for publication.

torbjokv (approved):

PR #2291, opened 2025-07-01
Testing larger batch size for dynamodbstream

truhacevkir:
- Processing multiple records from DynamoDbEvent per lambda invocation
- Updating BatchSize to 100
- Setting MaximumBatchingWindowInSeconds to 2 seconds

We will run Cristin import with that config to sandbox and test in all other environments in some weeks. In case it appears to be bad idea, we can simply turn previous setting on.

### DIFF
--- a/template.yaml
+++ b/template.yaml
@@ -1922,8 +1922,8 @@ Resources:
           Properties:
             Stream: !GetAtt NvaResourcesTable.StreamArn
             StartingPosition: TRIM_HORIZON
-            BatchSize: 100
-            MaximumBatchingWindowInSeconds: 2
+            BatchSize: 10
+            MaximumBatchingWindowInSeconds: 0
             Enabled: true
             MaximumRetryAttempts: 1
             BisectBatchOnFunctionError: true


template.yaml
```diff
@@ -1918,17 +1918,32 @@ Resources:
       Events:
         FanoutSource:
           Type: DynamoDB
+#==============================PROPERTIES WE WILL TEST=======================================================================
           Properties:
             Stream: !GetAtt NvaResourcesTable.StreamArn
             StartingPosition: TRIM_HORIZON
-            BatchSize: 1
+            BatchSize: 100
+            MaximumBatchingWindowInSeconds: 2
```
torbjokv, line 1926: Suggest setting 1 to keep round trip time as low as possible? like when we do publish on a resource, its important to keep latency low.

Maybe even skip MaximumBatchingWindowInSeconds (keep it at 0) and set BatchSize to 10

torbjokv (approved):
Looking forward to see how this works :) one suggestion for adjusting properties

PR #2295, opened 2025-07-03
Updating resource service in imports

truhacevkir:
There were multiple handlers that have not been updated with customerClient and channelClient in resourceService builder. Removing ResourceServiceBuilder and using constructor instead.


publication-commons/src/main/java/no/unit/nva/publication/service/impl/ResourceService.java
```diff
@@ -110,7 +110,6 @@ public class ResourceService extends ServiceWithTransactions {
     private static final Logger logger = LoggerFactory.getLogger(ResourceService.class);
     private final String tableName;
     private final Clock clockForTimestamps;
-    private final Supplier<SortableIdentifier> identifierSupplier;
```
truhacevkir, line 113: Removing identifier supplier as it is not used everywhere and is tested only to make sure that we do not allow inserting publication with the same identifier twice. Testing it directly in ServiceWithTransaction instead and removing it from ResourceService.

publication-commons/src/main/java/no/unit/nva/publication/service/impl/ResourceService.java
```diff
@@ -119,14 +120,21 @@ public class ResourceService extends ServiceWithTransactions {
...
+                           ChannelClaimClient channelClaimClient,
+                           CustomerService customerService) {
         super(dynamoDBClient);
+
+        requireNonNull(dynamoDBClient, "DynamoDbClient cannot be null");
+        requireNonNull(tableName, "Table name cannot be null");
+        requireNonNull(uriRetriever, "UriRetriever name cannot be null");
+        requireNonNull(channelClaimClient, "ChannelClaimClient name cannot be null");
+        requireNonNull(customerService, "CustomerService name cannot be null");
+
```
truhacevkir, line 135: Not sure this is the best way of validating service on its instantiation, probably better with factory method, but I think it is OK, cause validation rules are simple :)

brinxmat (approved):

PR #2302, opened 2025-07-30
Use jvm toolchain

brinxmat:
Sets JVM toolchain to ensure everyone is building using the same major version of _vendored_ Java

Additionally,
- Remove the legacy (and honestly pointless) source/target version declarations
- Enforce the compilation target option for Java (which actually does work)
- Use configuration cache, which shaves further seconds off the build if no gradle config changes are made

Take a look at the table and information following (I think we do want strong assertion of versions): https://docs.gradle.org/current/userguide/toolchains.html#comparison_table_for_setting_project_toolchains

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -12,6 +12,7 @@ def javaVersion = 21
 java {
     toolchain {
         languageVersion = JavaLanguageVersion.of(javaVersion)
+        vendor = JvmVendorSpec.AMAZON
     }
 }
 


build.gradle
```diff
@@ -8,8 +8,15 @@ allprojects{
     apply plugin: 'nva.publication.api.dynamo-db'
 }
 
-tasks.withType(JavaCompile) {
```
brinxmat, line 11: This did nothing

build.gradle
```diff
@@ -8,8 +8,15 @@ allprojects{
...
-    sourceCompatibility = JavaVersion.VERSION_21
+def javaVersion = 21
+java {
+    toolchain {
+        languageVersion = JavaLanguageVersion.of(javaVersion)
+    }
+}
```
brinxmat, line 16: This says "Use java 21" and the wrapper will download the correct version of java so you don't have to…

build.gradle
```diff
@@ -8,8 +8,15 @@ allprojects{
...
+
+tasks.withType(JavaCompile).configureEach {
+    options.release = javaVersion
```
brinxmat, line 19: This enforces the target version

buildSrc/src/main/groovy/nva.publication.api.java-conventions.gradle
```diff
@@ -28,11 +28,6 @@ dependencies {
     }
 }
 
-java {
-    sourceCompatibility = JavaVersion.VERSION_21
-    targetCompatibility = JavaVersion.VERSION_21
-}
```
brinxmat, line 34: Although setting sourceCompatibility in the previous case didn't have any real effect, we decided to do it twice.

gradle.properties
```diff
@@ -2,3 +2,4 @@ org.gradle.parallel=true
 org.gradle.caching=true
 org.gradle.console=verbose
 org.gradle.warning.mode=all
+org.gradle.configuration-cache=true
```
brinxmat, line 5: This will ensure that gradle configs are cached so we don't spend unnecessary time here

build.gradle
brinxmat: And is vendor-specific, which it is.

brinxmat: Does this now ensure that our builds are now repeatable?

ketilaa (approved):
👍

PR #2303, opened 2025-07-31
Fix gradle-specific, space-assignment deprecation warnings

brinxmat:
This PR removes deprecation warnings observed when building this project, see results at:
- https://scans.gradle.com/s/2rsycguigbwoq/deprecations
- https://scans.gradle.com/s/2rsycguigbwoq/console-log?anchor=0&page=1

Details:
- Fixes various warnings where the deprecated, gradle-specific space-based assignment like `myProp myValue`, which may be replaced by `myProp = myValue` or, `myProp(myValueKey, myValue)`
- Because these deprecated usages were in `root-plugin` and `java-conventions-plugin`, the number of errors was propagated to every project using these plugins.

### DIFF
--- a/buildSrc/src/main/groovy/nva.publication.api.java-conventions.gradle
+++ b/buildSrc/src/main/groovy/nva.publication.api.java-conventions.gradle
@@ -20,7 +20,7 @@ dependencies {
 
     // This is required until Gradle TestSuites are used, note no version is specified as this syncs
     // with JUnit from gradle.
-    testRuntimeOnly(group: 'org.junit.platform', name: 'junit-platform-launcher')
+    testRuntimeOnly 'org.junit.platform:junit-platform-launcher'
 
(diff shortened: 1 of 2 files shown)


buildSrc/src/main/groovy/nva.publication.api.java-conventions.gradle
```diff
@@ -20,7 +20,7 @@ dependencies {
 
     // This is required until Gradle TestSuites are used, note no version is specified as this syncs
     // with JUnit from gradle.
-    testRuntimeOnly(group: 'org.junit.platform', name: 'junit-platform-launcher')
+    testRuntimeOnly 'org.junit.platform:junit-platform-launcher'
```
brinxmat, line 23: The assignment here is not problematic, but we should migrate away from it because we should at some point move to Kotlin as this is the recommended syntax and the syntax above is gradle specific :(

PR #2316, opened 2025-08-14
removing PK1 secondary index

truhacevkir:
Removing ByTypeCustomerStatus secondary index (PK1)


truhacevkir:
Sandbox is migrated and things seems to work as they should.

PR #2322, opened 2025-08-22
fix(resource-service): large number of contributors causes timeout on save #NP-49617

torbjokv:
Fixing issue with publication where we have large amounts of contributors with affiliation where a looping lookup times out. Trying to fix with a conditional cache when contributor count is larger than 100.

Also refactored duplicated test code into FakeCristinUnitsUtil

Code is dependent on this PR: https://github.com/BIBSYSDEV/nva-cristin-service/pull/431


publication-commons/src/main/java/no/unit/nva/publication/model/utils/CuratingInstitutionsUtil.java
```diff
@@ -30,9 +31,28 @@ public CuratingInstitutionsUtil(RawContentRetriever uriRetriever, CustomerServic
...
 
-    public Set<CuratingInstitution> getCuratingInstitutionsOnline(Publication publication) {
-        return getAffiliatedContributors(publication.getEntityDescription()).flatMap(this::toCuratingInstitutionOnline)
-                   .filter(this::isCustomer)
+    public Set<CuratingInstitution> getCuratingInstitutions(EntityDescription entityDescription,
+                                                            CristinUnitsUtil cristinUnitsUtil) {
+        var contributors = getAffiliatedContributors(entityDescription).toList();
+
+        var topLevelMap = contributors.stream()
+                              .flatMap(contributor -> contributors.size() > CONTRIBUTOR_CACHE_THRESHOLD
```
torbjokv, line 39: This is the main change.

template.yaml
```diff
@@ -160,6 +160,7 @@ Globals:
         BACKEND_CLIENT_SECRET_NAME: 'BackendCognitoClientCredentials'
         EXTERNAL_USER_POOL_URI: !Ref ExternalCognitoAuthorizerUrl
         COGNITO_AUTHORIZER_URLS: !Join [",", !Ref CognitoAuthorizerUrls]
+        CRISTIN_UNITS_S3_URI: !Ref CristinUnitsS3ObjectUri
```
torbjokv, line 163: not great putting it here, but its hard to predict where we use ResourceService

PR #2326, opened 2025-09-02
feat: Enable backup of Scopus buckets

LarsV123:
Ticket: https://sikt.atlassian.net/browse/NP-49604

Changes:
- Adds `ScopusZipBucket` to the backup schedule
- Removes retention policy from `ScopusZipBucket` as suggested (may want to add this back later, but not sure)

### DIFF
--- a/template.yaml
+++ b/template.yaml
@@ -2780,7 +2780,7 @@ Resources:
               NoncurrentDays: !Ref S3NoncurrentVersionExpirationDays
       Tags:
           - Key: IncludedInBackup
-            Value: "false"
+            Value: "true"
       VersioningConfiguration:
           Status: Enabled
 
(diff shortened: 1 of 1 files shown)


template.yaml
```diff
@@ -2799,7 +2799,7 @@ Resources:
               NoncurrentDays: !Ref S3NoncurrentVersionExpirationDays
       Tags:
           - Key: IncludedInBackup
-            Value: "false"
+            Value: "true"
```
truhacevkir, line 2802: Do we need both? Zip bucket contains files from xml bucket in zip file. I would ask JEG if we simply want to keep all zip files in zip bucket + include in backup. (remove DeleteContentAfter3Days policy on ScopusZipBucket)

truhacevkir (approved):

truhacevkir (approved):

PR #2327, opened 2025-09-02
adding third party system to userinstance

truhacevkir:
Adding third party system to UserInstance and defaulting to null for now. 

Plan is to propogate this value from System header which external clients must send in on all write requests. Value will be used to populate log for actions external clients do.


publication-commons/src/main/java/no/unit/nva/publication/model/business/ThirdPartySystem.java
```diff
@@ -0,0 +1,25 @@
+package no.unit.nva.publication.model.business;
+
+import static java.util.Arrays.stream;
+import com.fasterxml.jackson.annotation.JsonValue;
+
+public enum ThirdPartySystem {
+    WISE_FLOW("WISEflow"), INSPERA("Inspera"), OTHER("Other");
```
truhacevkir, line 7: Not sure this is the best name for this enum. Come with suggestions if you have ideas.

tickets/src/test/java/no/unit/nva/publication/ticket/TicketTestLocal.java
```diff
@@ -67,9 +61,9 @@ protected TicketEntry createPersistedDoiRequestWithOwnerAffiliation(Publication
         return DoiRequest.create(Resource.fromPublication(publication), userInstance).persistNewTicket(ticketService);
     }
 
-    public static UserInstance userInstanceWithTopLevelCristinOrg(URI ownerAffiliation) {
-        return new UserInstance(randomString(), randomUri(), ownerAffiliation, randomUri(), randomUri(),
-                                List.of(), UserClientType.INTERNAL);
+    public static UserInstance userInstanceWithTopLevelCristinOrg(URI topLevelOrgCristinId) {
+        return new UserInstance(randomString(), randomUri(), topLevelOrgCristinId, randomUri(), randomUri(),
+                                List.of(), UserClientType.INTERNAL, null);
```
truhacevkir, line 66: I mean this is the right name of this variable here, correct me if I am wrong.

PR #2328, opened 2025-09-03
adding etag to publication response and validation on PUT

truhacevkir:


publication-rest/src/test/java/no/unit/nva/publication/fetch/FetchPublicationHandlerTest.java
```diff
@@ -514,10 +515,42 @@ void shouldReturnUnpublishedPublicationWithFilesWhenUserHasRightsToUpdatedPublic
...
+
+        assertEquals(gatewayResponse.getHeaders().get(ACCESS_CONTROL_EXPOSE_HEADERS), ETAG);
+        assertEquals(String.valueOf(expectedEtag), gatewayResponse.getHeaders().get(ETAG));
+    }
+
+    @Test
+    void shouldReturnNotModifiedWhenProvidingIfNoneMatchHeaderWithETagMatchingCurrentVersionOfPublication()
+        throws ApiGatewayException, IOException {
+        var publication = createPublication();
+        var version = Resource.fromPublication(publication).fetch(publicationService).orElseThrow().getVersion();
```
torbjokv, line 541: create a test that sends garbage in version? should return latest object probably?

publication-rest/src/test/java/no/unit/nva/publication/update/UpdatePublicationHandlerTest.java
```diff
@@ -1954,6 +2034,27 @@ void partialUpdateShouldNotValidateFilesNotPresentInRequest()
         assertEquals(HTTP_OK, response.getStatusCode());
     }
 
+    @Test
+    void shouldReturnPreconditionFailedWhenIfMatchHeaderIsProvidedButVersionOfResourcesIsOutdated()
```
torbjokv, line 2038: do you test the positive case also where update is done with the correct etag?

PR #2332, opened 2025-09-05
Np 49730 remove deprecated update files allowed operation

truhacevkir:
- Removing two deprecated allowed operations (Frontend does not use them)
- Refactoring whole Rrs applier and finder in order to make is easier to navigate and fixing bug.


publication-commons/src/main/java/no/unit/nva/publication/rightsretention/RightsRetentionsValueFinder.java
```diff
@@ -1,126 +0,0 @@
...
-    private boolean isValid(RightsRetentionStrategy rrs) {
-        return getAllowedConfigurations(rrs).contains(rrs.getConfiguredType())
-               || isAllowedToOverrideRrs(rrs, permissionStrategy);
-    }
-
-    private boolean isAllowedToOverrideRrs(RightsRetentionStrategy rrs,
-                                           PublicationPermissions permissionStrategy) {
-        if (nonNull(permissionStrategy) && rrs instanceof OverriddenRightsRetentionStrategy) {
-            return permissionStrategy.isPublishingCuratorOnPublication();
-        }
```
truhacevkir, line 102: Bug is here -> isPublishingCuratorOnPublication() is used only here and is based on deprecated allowed operation. 
Impossible to fix this without large refactoring, cause this check is based on PublicationPermissions, but new one on FilePermission and FileEntry.

torbjokv (approved):

PR #2333, opened 2025-09-08
persist recovery message when fetching affiliation fails

truhacevkir:
Send message to recovery queue when could not fetch affiliation from Cristin under expansion.

Little refactor which will: 
- Send message to recovery queue when channel and organization return other status codes than 200
(before we threw an exception that failed on lambda invocation and message has been persisted based on "failed invocation" and not "failing channel")

### DIFF
--- a/expansion/src/main/java/no/unit/nva/expansion/model/IndexDocumentWrapperLinkedData.java
+++ b/expansion/src/main/java/no/unit/nva/expansion/model/IndexDocumentWrapperLinkedData.java
@@ -30,7 +30,6 @@ import java.util.Objects;
 import java.util.Optional;
 import no.unit.nva.auth.uriretriever.RawContentRetriever;
 import no.unit.nva.commons.json.JsonUtils;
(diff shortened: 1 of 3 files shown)


expansion/src/main/java/no/unit/nva/expansion/model/IndexDocumentWrapperLinkedData.java
```diff
@@ -47,8 +46,6 @@
 public class IndexDocumentWrapperLinkedData {
 
     private static final Logger logger = LoggerFactory.getLogger(IndexDocumentWrapperLinkedData.class);
-    private static final String CRISTIN_VERSION = "; version=2023-05-26";
-    private static final String MEDIA_TYPE_JSON_LD_V2 = APPLICATION_JSON_LD.toString() + CRISTIN_VERSION;
```
truhacevkir, line 51: It seems like cristin-api serves the same response when sending 
`application/json` and `application/ld+json; version=2023-05-26`

PR #2345, opened 2025-09-21
feature(event-handlers): add sqs based batch job handler for processing DynamoDB resources

torbjokv:
# DynamoDB Batch Job Handler System

**Motivation:** We want to allow specialized one-off jobs like a targeted "batch scan"/reindex in a structured way with DLQ, "rate limiting" and with logs. Idea is to have a sqs work queue where we just dump pairs of db keys and task identifiers.

**Alternative implementation considerations:** Not using SQS was considered with the advantage of allowing us to batch 700 items in each write transaction, but was deemed impractical since we want to dump the whole job to sqs and allow to adjust the speed, not bringing down our application or not mess up by running from laptop and something stops it unexpected.

## Architecture
- **SQS Queue**: Receives batch work items (up to 10 per batch)
- **Lambda Handler**: Processes messages in batches grouped by job type
- **DLQ**: Failed messages after 3 retries
- **DynamoDB Transactions**: Atomic batch updates

### 1. DynamodbResourceBatchJobHandler
Main Lambda handler that:
- Groups SQS messages by job type
- Executes entire batches as single transactions
- Reports individual message failures for retry
- Adds error metadata for debugging

### 2. BatchWorkItem
Record class representing a work item:
```java
record BatchWorkItem(
    DynamodbResourceBatchDynamoDbKey dynamoDbKey,
    String jobType,
    JsonNode parameters
)
```

### 3. DynamodbResourceBatchDynamoDbKey
Supports both primary keys and GSI queries:
```java
record DynamodbResourceBatchDynamoDbKey(
    String partitionKey,
    String sortKey,
    String indexName  // null for primary key, GSI name for GSI queries
)
```

### 4. ReindexRecordJob
Implementation that updates version fields to trigger reindexing:
- **Primary Key Support**: Direct updates using PK0/SK0
- **GSI Query Support**: Resolves GSI queries to primary keys first
- **Batch Transactions**: All updates in single transaction
- **Pagination**: Handles large GSI query results

### Queue Message Format
```json
{
  "dynamoDbKey": {
    "partitionKey": "Resource:14d49ab7-4d1d-464d-b732-54b5c46ce6cc:34322@20754.0.0.0",
    "sortKey": "Resource:01995cfe7bce-c1b5a50c-c2a9-4502-8fa6-3790cdb0bfee"
  },
  "jobType": "REINDEX_RECORD",
  "parameters": {}
}
```

For GSI queries:
```json
{
  "dynamoDbKey": {
    "partitionKey": "Resource:01995cfe7bce-c1b5a50c-c2a9-4502-8fa6-3790cdb0bfee",
    "sortKey": "Resource:01995cfe7bce-c1b5a50c-c2a9-4502-8fa6-3790cdb0bfee",
    "indexName": "ResourcesByIdentifier"
  },
  "jobType": "REINDEX_RECORD",
  "parameters": {}
}
```

### Adding New Job Types

1. Implement `DynamodbResourceBatchJobExecutor`:
```java
public class MyCustomJob implements DynamodbResourceBatchJobExecutor {
    @Override
    public void executeBatch(List<BatchWorkItem> workItems) {
        // Process entire batch
        // Use transactions for atomicity
    }

    @Override
    public String getJobType() {
        return "MY_CUSTOM_JOB";
    }
}
```

2. Register in `DynamodbResourceBatchJobHandler.JOBS` array:
```java
private static final DynamodbResourceBatchJobExecutor[] JOBS = {
    new ReindexRecordJob(),
    new MyCustomJob()  // Add your job here
};
```

### DIFF
--- a/template.yaml
+++ b/template.yaml
@@ -2939,13 +2939,13 @@ Resources:
     Properties:
       QueueName: DynamodbResourceBatchJobDlq
       MessageRetentionPeriod: 1209600
-      VisibilityTimeoutSeconds: 60
+      VisibilityTimeout: 60
 
   DynamodbResourceBatchJobWorkQueue:
     Type: AWS::SQS::Queue
     Properties:
       QueueName: DynamodbResourceBatchJobWorkQueue
-      VisibilityTimeoutSeconds: 360 # Should be at least 6 times the Lambda timeout
+      VisibilityTimeout: 360 # Should be at least 6 times the Lambda timeout
       MessageRetentionPeriod: 1209600 # 14 days
       ReceiveMessageWaitTimeSeconds: 20
       RedrivePolicy:
--- a/template.yaml
+++ b/template.yaml
@@ -2932,3 +2932,61 @@ Resources:
                 responsePayload:
                   topic:
                     - PublicationService.Resource.Deleted
+
+  # DynamoDB Batch Job Processing Resources
+  DynamodbResourceBatchJobDLQ:
+    Type: AWS::SQS::Queue
+    Properties:
+      QueueName: DynamodbResourceBatchJobDlq
+      MessageRetentionPeriod: 1209600
+      VisibilityTimeoutSeconds: 60
+
+  DynamodbResourceBatchJobWorkQueue:
+    Type: AWS::SQS::Queue
+    Properties:
+      QueueName: DynamodbResourceBatchJobWorkQueue
+      VisibilityTimeoutSeconds: 360 # Should be at least 6 times the Lambda timeout
+      MessageRetentionPeriod: 1209600 # 14 days
+      ReceiveMessageWaitTimeSeconds: 20
+      RedrivePolicy:
+        deadLetterTargetArn: !GetAtt DynamodbResourceBatchJobDLQ.Arn
+        maxReceiveCount: 3
+      RedriveAllowPolicy:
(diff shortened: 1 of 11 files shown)


template.yaml
```diff
@@ -2931,4 +2931,56 @@ Resources:
...
+  DynamodbResourceBatchJobFunction:
+    Type: AWS::Serverless::Function
+    Properties:
+      Handler: no.unit.nva.publication.events.handlers.batch.dynamodb.DynamodbResourceBatchJobHandler::handleRequest
+      CodeUri: publication-event-handlers
+      Policies:
+        - arn:aws:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole
+        - !GetAtt ResourceServiceAccessManagedPolicy.PolicyArn
+        - !GetAtt EventsLambdaManagedPolicy.PolicyArn
+      Timeout: 60 # if changing, also change the visibility timeout of the SQS queue
```
LarsV123, line 2965: nitpick: These timeouts and retention periods etc should probably be pinned in a top-level parameter with a default value, because I assume we usually want them to be the same for all queues/lambdas.

torbjokv: hmmm I'm not sure. maybe all in some groups, like endpoints, handlers, etc, but still I'm not sure. Default lambda should probably be 20 like it is now and this is an exception?

PR #2363, opened 2025-10-01
combining file approval and ticket completing to one transaction

truhacevkir:
**Issue Description**

**Bug:** 
FilesForApproval disappear from ticket when curator approves publishing request. (PUT ticket/{ticketIdentifier}
**Root Cause Analysis:**
While we couldn't reproduce the bug directly, code analysis reveals a race condition in the ticket completion process. Currently, we execute two separate database transactions on the same ticket:

Transaction 1: Move files from filesForApproval to approvedFiles
Transaction 2: Update ticket status to "completed" and clear filesForApproval

**The Problem:** 
Transaction 2 likely fetches a stale version of the ticket (before Transaction 1's updates are committed), then clears the original filesForApproval list—effectively deleting files that were just approved.
Solution: Merge both operations into a single atomic transaction. Since we always approve files when completing tickets, this consolidation eliminates the race condition.

### DIFF
--- a/publication-commons/src/main/java/no/unit/nva/publication/model/FilesApprovalEntry.java
+++ b/publication-commons/src/main/java/no/unit/nva/publication/model/FilesApprovalEntry.java
@@ -56,12 +56,18 @@ public abstract class FilesApprovalEntry extends TicketEntry {
     @Override
     public FilesApprovalEntry complete(Publication publication, UserInstance userInstance) {
         var completed = (FilesApprovalEntry) super.complete(publication, userInstance);
-        completed.setApprovedFiles(getFilesForApproval().stream()
-                                 .filter(file -> publication.getFile(file.getIdentifier()).filter(PendingFile.class::isInstance).isPresent())
-                                 .map(this::toApprovedFile)
-                                 .collect(Collectors.toSet()));
+        completed.setApprovedFiles(getPendingFilesToApprove(publication));
         completed.setFilesForApproval(Set.of());
         return completed;
+    }
+
+    private Set<File> getPendingFilesToApprove(Publication publication) {
+        return getFilesForApproval().stream()
+                   .filter(file -> publication.getFile(file.getIdentifier())
+                                       .filter(PendingFile.class::isInstance)
+                                       .isPresent())
+                   .map(this::toApprovedFile)
+                   .collect(Collectors.toSet());
     }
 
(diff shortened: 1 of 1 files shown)


tickets/src/main/java/no/unit/nva/publication/ticket/update/UpdateTicketHandler.java
```diff
@@ -248,39 +238,6 @@ private void updateStatus(UpdateTicketRequest ticketRequest, RequestUtils reques
...
-                   .map(File::getIdentifier)
-                   .map(String::valueOf)
-                   .collect(Collectors.joining(","));
-    }
-
-    private static boolean filesForApprovalAreNotApprovable(FilesApprovalEntry ticket) {
-        return ticket.getFilesForApproval().stream()
-                   .map(PendingFile.class::cast)
-                   .anyMatch(PendingFile::isNotApprovable);
-    }
```
truhacevkir, line 282: This check is also made in datamodel level when checking if file is approvable. Removing it from here, the only difference is that we will now serve bad request instead of conflict via API.

publication-rest/src/main/java/no/unit/nva/publication/update/PublishingRequestResolver.java
```diff
@@ -117,7 +117,6 @@ private TicketEntry persistPublishingRequest(
         throws ApiGatewayException {
         return customerAllowsPublishingMetadataAndFiles()
                    ? publishingRequest
-                         .approveFiles()
```
truhacevkir, line 120: Complete + approveFiles is always done at the same time, moving `approveFiles()` directly to` comlete()` method

publication-commons/src/main/java/no/unit/nva/publication/model/FilesApprovalEntry.java
```diff
@@ -56,7 +56,11 @@ public abstract class FilesApprovalEntry extends TicketEntry {
     @Override
     public FilesApprovalEntry complete(Publication publication, UserInstance userInstance) {
         var completed = (FilesApprovalEntry) super.complete(publication, userInstance);
-        completed.emptyFilesForApproval();
+        completed.setApprovedFiles(getFilesForApproval().stream()
+                                 .filter(file -> publication.getFile(file.getIdentifier()).filter(PendingFile.class::isInstance).isPresent())
```
brinxmat, line 60: pull a method out

brinxmat (approved):

PR #2369, opened 2025-10-03
feature(batch-job): load full db scan as batch job on sqs

torbjokv:


publication-event-handlers/src/main/java/no/unit/nva/publication/events/handlers/batch/dynamodb/LoadDynamodbResourceBatchJobHandler.java
```diff
@@ -0,0 +1,226 @@
...
+                                      .withTableName(tableName)
+                                      .withLimit(SCAN_PAGE_SIZE)
+                                      .withAttributesToGet(PRIMARY_KEY_PARTITION_KEY_NAME, PRIMARY_KEY_SORT_KEY_NAME);
+
+        if (lastEvaluatedKey != null && !lastEvaluatedKey.isEmpty()) {
+            scanRequest.withExclusiveStartKey(lastEvaluatedKey);
+        }
+
+        return scanRequest;
+    }
```
truhacevkir, line 163: BatchScan based on PK0 iterates through IdEntries and LogEntries as well, I think this batch will take about 2x the time of current batch scan (which is iterating through PK2)

template.yaml
```diff
@@ -2987,4 +2986,36 @@ Resources:
     Type: AWS::Logs::LogGroup
     Properties:
       LogGroupName: !Sub /aws/lambda/${DynamodbResourceBatchJobFunction}
+
+  LoadDynamodbResourceBatchJobHandler:
+    Type: AWS::Serverless::Function
+    Properties:
+      Description: 'Iterates through all DynamoDB records and queues them for batch processing. Example EventBridge input: {"detail": {"jobType": "REINDEX_RECORD"}}'
```
truhacevkir, line 2993: Do we have consumer for this queue?

publication-event-handlers/src/main/java/no/unit/nva/publication/events/handlers/batch/dynamodb/LoadDynamodbResourceBatchJobHandler.java
```diff
@@ -0,0 +1,226 @@
...
+            .supplyAsync(() -> batches.parallelStream()
+                .mapToInt(this::sendBatchToQueue)
+                .sum(), forkJoinPool)
+            .exceptionally(e -> {
+                logger.error("Failed to process batches in parallel, falling back to sequential", e);
+                return batches.stream()
+                    .mapToInt(this::sendBatchToQueue)
+                    .sum();
+            })
+            .join();
```
truhacevkir, line 124: What about trying parallel scan here? In order to make it way faster?

. Use Parallel Scans for Large Tables

**DynamoDB allows you to divide a scan into multiple segments that run concurrently**
Specify TotalSegments and Segment parameters to distribute the workload
This is most effective when you need to process the entire table quickly
Example: Split into 4-8 segments depending on table size and provisioned throughput

PR #2370, opened 2025-10-03
Np 49958 set published date

truhacevkir:
Setting published date when external client creates publication with status published.


publication-rest/src/test/java/no/unit/nva/publication/create/CreatePublicationHandlerTest.java
```diff
@@ -227,6 +228,21 @@ void shouldPersistProvidedStatusWhenMachineUserPersistsPublicationWithAnyStatus(
         assertThat(publicationResponse.getStatus(), is(equalTo(status)));
     }
 
+    @Test
+    void shouldSetPublishedDateWhenMachineUserPersistsPublicationWithStatusPublished()
+        throws Exception {
```
truhacevkir, line 233: There are test cases that verify that not machine user is not able to create publication with other statuses than DRAFT.

brinxmat (approved):

PR #2374, opened 2025-10-08
Fix contributor node identities

brinxmat:
- Remove JSON-workflow for contributor count, contributor preview
- Use semantics


torbjokv (approved):

truhacevkir (approved):
I think contributor preview is used in search document that external use, but okey to remove?

PR #2509, opened 2026-03-24
Fix:json-ld context maps alternativeAbstracts correctly and license as IRI

brinxmat:
- Alternative abstracts was mapped to alternativeTitle, which is not correct
- Now mapped to alternativeAbstract, which is correct
- License is added — as an IRI, it was missing.


truhacevkir (approved):

torbjokv (approved):
looks good

LarsV123 (approved):

PR #2516, opened 2026-04-20
perf: Enable SnapStart for all lambdas

LarsV123:
Enables SnapStart for all lambdas to reduce cold start problems.

### DIFF
--- a/template.yaml
+++ b/template.yaml
@@ -149,6 +149,11 @@ Globals:
     Runtime: java21
     Timeout: 20
     MemorySize: 1800 # 1 vCPU
+    AutoPublishAlias: live
+    DeploymentPreference:
+      Type: AllAtOnce
+    SnapStart:
+      ApplyOn: PublishedVersions
     Environment:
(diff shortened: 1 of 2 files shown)


template.yaml
```diff
@@ -149,6 +149,11 @@ Globals:
     Runtime: java21
     Timeout: 20
     MemorySize: 1800 # 1 vCPU
+    AutoPublishAlias: live
+    DeploymentPreference:
+      Type: AllAtOnce
+    SnapStart:
+      ApplyOn: PublishedVersions
```
torbjokv, line 144: I would try to first document if this helps or not in sandbox or somewhere else (e2e?) before we do this. There is complexity and deployment "penalties" attached to versioning of lambdas. We also sometimes experience some weird deployment and rollback issues.

LarsV123: Tested yesterday and it seems to work in E2E + sandbox.

brinxmat (approved):

PR #2518, opened 2026-04-24
fix: Prevent automated deletion of resources

LarsV123:
Prevents important resources from being deleted automatically (e.g. if stack deleted or resource modified). See https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-attribute-deletionpolicy.html for details.

Changes:

- [fix: Set policies to prevent automatic deletion of tables/buckets](https://github.com/BIBSYSDEV/nva-publication-api/pull/2518/commits/a98aab1038ce8911308a8f77a0228334431b9cb3)
- [refactor: Set default values as explicit policies](https://github.com/BIBSYSDEV/nva-publication-api/pull/2518/commits/35597ae65f0b3e339181dca807ae818e58fc3e70)

### DIFF
--- a/template.yaml
+++ b/template.yaml
@@ -177,7 +177,7 @@ Resources:
   NvaResourcesTable:
     Type: AWS::DynamoDB::Table
     UpdateReplacePolicy: Retain
-    DeletionPolicy: Delete # DeletionPolicy should be added on main branch stacks through use of stack policy
+    DeletionPolicy: Retain
     Properties:
       TableName: !Sub nva-resources-${AWS::StackName}
       BillingMode: PAY_PER_REQUEST
@@ -252,7 +252,7 @@ Resources:
   NvaImportCandidatesTable:
     Type: AWS::DynamoDB::Table
     UpdateReplacePolicy: Retain
-    DeletionPolicy: Delete # DeletionPolicy should be added on main branch stacks through use of stack policy
+    DeletionPolicy: Retain
     Properties:
       TableName: !Sub nva-import-candidates-${AWS::StackName}
       BillingMode: PAY_PER_REQUEST
@@ -2775,6 +2775,8 @@ Resources:
 
   BrageMigrationInputBucket:
     Type: AWS::S3::Bucket
+    UpdateReplacePolicy: Retain
+    DeletionPolicy: Retain
     Properties:
       BucketName: !Sub "${BrageMigrationInputBucketName}-${AWS::AccountId}"
       LifecycleConfiguration:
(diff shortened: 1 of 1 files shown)


template.yaml
```diff
@@ -177,7 +177,7 @@ Resources:
   NvaResourcesTable:
     Type: AWS::DynamoDB::Table
     UpdateReplacePolicy: Retain
-    DeletionPolicy: Delete # DeletionPolicy should be added on main branch stacks through use of stack policy
```
LarsV123, line 180: As far as I can tell, we don't have such a stack policy.

template.yaml
```diff
@@ -2861,8 +2871,13 @@ Resources:
 
   ScopusImportReportBucket:
     Type: AWS::S3::Bucket
+    UpdateReplacePolicy: Delete
+    DeletionPolicy: Delete
     Properties:
       BucketName: !Sub "${ScopusImportReportBucketName}-${AWS::AccountId}"
+      Tags:
+        - Key: IncludedInBackup
+          Value: "false"
```
LarsV123, line 2880: I'm explicitly setting the default values here and for `ScopusXmlBucket` to make it clear that these are unprotected on purpose. Given that they are not included in the backup, I assume we do not need to protect the resources from deletion.

template.yaml
torbjokv: not sure if this helps very much as we have both snapshots and backups. But it will make it harder to rebuild environments?

LarsV123: I think it will require us to manually delete these resources if we want to fully delete and recreate the stack. This shouldn't take too much time and we don't do it very often.

We should be able to restore from backups if something is deleted by accident, but it's better if we don't need to. Deleting these resources is such an exceptional case that I think it's fine to require explicit manual action.

torbjokv: with PIT enabled, it will also do an automatic snapshot on delete (https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/PointInTimeRecovery_Howitworks.html#howitworks-pitr-deleting-table). If retained and we redeploy all the stacks, we still have to recreate the data into a new table that matches the name in the new stack and I see this as just extra steps. While we have not deleted environments lately, it is something we should do more of to be on top of our deployment routines.  It is already way too painful to clean and reconfigure a new environment, and that is the reasoning for my push-back. To help me understand, what does the actual table give us that the snapshot dont?

brinxmat (approved):

torbjokv (approved):

PR #2523, opened 2026-05-04
fix(publication-model): validate SoftwareSourceCode at publish, not on save

torbjokv:
## Summary

Allow drafts of `SoftwareSourceCode` publications to be saved with a missing `softwareVersion`, and validate the field at `/publish`. Failures are returned to the frontend as RFC 9457 structured errors with JSON Pointers, so a UI can highlight every offending field at once.

- Drop the eager `@JsonProperty(required = true)` / `requireNonNull` on `SoftwareSourceCode.softwareVersion` so drafts deserialize and persist
- Add `PublishValidator` — a small capability interface in `publication-model` that types opt into when they have publish-time rules. `SoftwareSourceCode` implements it and requires a non-blank `softwareVersion`, with the JSON Pointer `#/entityDescription/reference/publicationInstance/softwareVersion`
- Unify `Resource.publish()` validation into a single `validateForPublish()` pass that aggregates universal Resource rules (status, mainTitle) and per-instance rules (`PublishValidator`) into one `List<ValidationError>` — frontend gets every problem at once
- Universal checks now emit structured errors with pointers (`#/status`, `#/entityDescription/mainTitle`) instead of a generic "not publishable" message
- `Resource.publish()` throws domain `ValidationException` (in `publication-model`, enhanced to carry `List<ValidationError>`); the HTTP layer translates it. No api-gateway exception types in the business model
- `PublishPublicationHandler` translates `ValidationException` → `BadRequestException(message, errors)`; nva-commons serializes the list as a top-level `errors` extension on the Problem JSON
- Bump `nvaCommonsVersion` 2.6.1 → 2.7.0 for the new `BadRequestException(String, List<ValidationError>)` constructor and `WithErrors` interface

https://sikt.atlassian.net/browse/NP-51198

### DIFF
--- a/publication-commons/src/main/java/no/unit/nva/publication/model/business/Resource.java
+++ b/publication-commons/src/main/java/no/unit/nva/publication/model/business/Resource.java
@@ -75,6 +75,7 @@ import no.unit.nva.publication.service.impl.ResourceService;
 import no.unit.nva.publication.service.impl.TicketService;
 import nva.commons.apigateway.exceptions.BadRequestException;
 import nva.commons.apigateway.exceptions.NotFoundException;
+import nva.commons.apigateway.exceptions.ValidationError;
 import nva.commons.core.JacocoGenerated;
 
 @SuppressWarnings({
@@ -90,6 +91,7 @@ public class Resource implements Entity {
   public static final URI NOT_IMPORTANT = null;
   public static final List<PublicationStatus> PUBLISHABLE_STATUSES =
       List.of(DRAFT, PUBLISHED_METADATA, UNPUBLISHED);
+  public static final String RESOURCE_IS_NOT_PUBLISHABLE_MESSAGE = "Resource is not publishable!";
 
   @JsonProperty private SortableIdentifier identifier;
   @JsonProperty private PublicationStatus status;
@@ -431,7 +433,8 @@ public class Resource implements Entity {
         .toOptional();
   }
 
-  public Resource publish(ResourceService resourceService, UserInstance userInstance) {
+  public Resource publish(ResourceService resourceService, UserInstance userInstance)
+      throws BadRequestException {
     var resource = fetch(resourceService);
     if (resource.isPresent() && resource.get().isNotPublished()) {
       return resource.get().publish(userInstance, resourceService);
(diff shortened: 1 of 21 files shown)


publication-commons/src/main/java/no/unit/nva/publication/model/business/Resource.java
```diff
@@ -90,6 +93,12 @@ public class Resource implements Entity {
   public static final URI NOT_IMPORTANT = null;
   public static final List<PublicationStatus> PUBLISHABLE_STATUSES =
       List.of(DRAFT, PUBLISHED_METADATA, UNPUBLISHED);
+  public static final String RESOURCE_IS_NOT_PUBLISHABLE_MESSAGE = "Resource is not publishable!";
+  public static final String STATUS_NOT_PUBLISHABLE_MESSAGE =
+      "publication cannot be published in current status";
+  public static final String STATUS_POINTER = "#/status";
+  public static final String MAIN_TITLE_REQUIRED_MESSAGE = "mainTitle is required";
+  public static final String MAIN_TITLE_POINTER = "#/entityDescription/mainTitle";
```
brinxmat: Are these public for a reason?

torbjokv: nope, just slipped through. bad existing pattern was just continued. will fix.

publication-commons/src/main/java/no/unit/nva/publication/model/business/Resource.java
```diff
@@ -446,21 +463,44 @@ private Resource publish(UserInstance userInstance, ResourceService resourceServ
...
+    errors.addAll(collectInstanceValidationErrors());
+    return errors;
+  }
+
+  private boolean mainTitleIsMissing() {
+    return Optional.ofNullable(this.getEntityDescription())
+        .map(EntityDescription::getMainTitle)
+        .isEmpty();
+  }
+
```
brinxmat: If we are going to implement this, I think there is a pattern you should follow:
```
public class Resource implements Validatatable {
  public validate(Validator validator);
}

public interface Validatatable {
  validate(Validator validator);
}

public interface Validator {
  ValidationResult getValidationResult();
}

public class ResourceValidator(Resource resource) {
  public void validate(Resource resource) {
    ...
  }

public ResourceValidationResult getValidationResult() {
   ...
}
}

```

torbjokv: good idea! added a suggestion now. one issue left is the ugly json pointers, but I dont think I will do anything with those right now

publication-commons/src/main/java/no/unit/nva/publication/model/business/Resource.java
```diff
@@ -84,12 +86,14 @@
...
 
   public static final String TYPE = "Resource";
-  public static final URI NOT_IMPORTANT = null;
-  public static final List<PublicationStatus> PUBLISHABLE_STATUSES =
-      List.of(DRAFT, PUBLISHED_METADATA, UNPUBLISHED);
+
+  @JsonIgnore private static final URI NOT_IMPORTANT = null;
+
+  @JsonIgnore
+  private static final String RESOURCE_IS_NOT_PUBLISHABLE_MESSAGE = "Resource is not publishable!";
```
LarsV123, line 96: I don't think you need `@JsonIgnore` on private fields.

LarsV123 (approved):

PR #2531, opened 2026-06-01
feat(template): Grant NIB cross-account read access to Scopus buckets

torbjokv:
Grants the NIB AWS account read-only access to the Scopus XML and zip buckets, so NIB can fetch the source files directly from their own account.

- Add `NibExternalAccountId` parameter resolved from the SSM parameter `/NVA/ExternalAWSAccounts/NIB`
- Add bucket policies on `ScopusXmlBucket` and `ScopusZipBucket` allowing the NIB account `s3:GetObject`, `s3:GetObjectVersion` and `s3:ListBucket`

## Dependency
Depends on BIBSYSDEV/NVA-infrastructure#308, which documents `/NVA/ExternalAWSAccounts/NIB`. The SSM parameter must be created in every environment before this is deployed, otherwise the deploy fails when the parameter is resolved.

https://sikt.atlassian.net/browse/NP-51296

### DIFF
--- a/template.yaml
+++ b/template.yaml
@@ -84,7 +84,7 @@ Parameters:
   NibExternalAccountId:
     Type: AWS::SSM::Parameter::Value<String>
     Default: "/NVA/ExternalAWSAccounts/NIB"
-    Description: AWS account id for NIB, granted read-only access to the Scopus XML bucket
+    Description: AWS account id for NIB, granted read-only access to the selected buckets
   ScopusImportReportBucketName:
     Type: "String"
     Default: "scopus-import-reports"
--- a/template.yaml
+++ b/template.yaml
@@ -81,6 +81,10 @@ Parameters:
     Type: String
     Default: "scopus-xml-files-v2"
     Description: Name of bucket for unzipped scopus xml files
+  NibExternalAccountId:
+    Type: AWS::SSM::Parameter::Value<String>
+    Default: "/NVA/ExternalAWSAccounts/NIB"
+    Description: AWS account id for NIB, granted read-only access to the Scopus XML bucket
   ScopusImportReportBucketName:
     Type: "String"
     Default: "scopus-import-reports"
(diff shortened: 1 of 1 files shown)


joachimjorgensen (approved):

template.yaml
```diff
@@ -81,6 +81,10 @@ Parameters:
     Type: String
     Default: "scopus-xml-files-v2"
     Description: Name of bucket for unzipped scopus xml files
+  NibExternalAccountId:
+    Type: AWS::SSM::Parameter::Value<String>
+    Default: "/NVA/ExternalAWSAccounts/NIB"
+    Description: AWS account id for NIB, granted read-only access to the Scopus XML bucket
   ScopusImportReportBucketName:
```
LarsV123, line 88: This needs to be seeded in all environments before merging, as the stack deploy will fail without it.

template.yaml
```diff
@@ -81,6 +81,10 @@ Parameters:
     Type: String
     Default: "scopus-xml-files-v2"
     Description: Name of bucket for unzipped scopus xml files
+  NibExternalAccountId:
+    Type: AWS::SSM::Parameter::Value<String>
+    Default: "/NVA/ExternalAWSAccounts/NIB"
+    Description: AWS account id for NIB, granted read-only access to the Scopus XML bucket
```
LarsV123: nitpick: Mismatch between description and what it does (it also includes the ZIP bucket)

template.yaml
```diff
@@ -2799,6 +2801,52 @@ Resources:
...
+            Action:
+              - s3:GetObject
+              - s3:GetObjectVersion
+            Resource: !Sub "${ScopusXmlBucket.Arn}/*"
+          - Sid: NibListBucket
+            Effect: Allow
+            Principal:
+              AWS: !Sub "arn:aws:iam::${NibExternalAccountId}:root"
+            Action:
+              - s3:ListBucket
```
LarsV123, line 2824: Do they need `s3:ListBucketVersions` too?

LarsV123 (changes requested):

LarsV123 (approved):

template.yaml
torbjokv: maybe, lets try without first

PR #2534, opened 2026-06-02
refactor(sqs): unify SQS test fakes on the QueueClient abstraction

LarsV123:
Adds a new abstraction layer over the "send message batch" method in the SQS client that better decouples the production code from the AWS SDK and lets us delete a duplicate test fixture.


publication-file/src/main/java/no/unit/nva/publication/file/upload/restmodel/CreateUploadRequestBody.java
torbjokv, line 11: this time its actually what we want? :-)

LarsV123: Git shenanigans, seems to be gone from the diff now...

torbjokv (approved):
Nice! should we look into wrapping in- and outputs also?

PR #2543, opened 2026-07-06
Add text extraction pipeline for files

brinxmat:
This adds several dependencies for text extraction:

- The basic text-extraction worker, which takes content from the resources bucket and extracts text from them
  - PDF
  - Text
  - LaTeX
  - OOXML, Doc
- A seeding workflow that allows us to index keysets batchwise


gradle/libs.versions.toml
```diff
@@ -32,12 +34,17 @@ jaxb-runtime = { group = 'org.glassfish.jaxb', name = 'jaxb-runtime', version.re
 joda-time = { group = 'joda-time', name = 'joda-time', version.ref = 'joda-time' }
 lombok = { group = 'org.projectlombok', name = 'lombok', version.ref = 'lombok' }
 nva-doi-partner-data = { group = 'com.github.BIBSYSDEV', name = 'nva-doi-partner-data', version.ref = 'nvaDoiPartnerData' }
+pdfbox = { group = 'org.apache.pdfbox', name = 'pdfbox', version.ref = 'pdfbox' }
+poi = { group = 'org.apache.poi', name = 'poi', version.ref = 'poi' }
+poi-ooxml = { group = 'org.apache.poi', name = 'poi-ooxml', version.ref = 'poi' }
```
LarsV123: These are available from the shared version catalog. Usage:

```
implementation(nvaCatalog.apache.poi)
implementation(nvaCatalog.apache.poi.ooxml)
```

publication-file-text/src/main/java/no/unit/nva/publication/file/text/SeedTextExtractionHandler.java
```diff
@@ -0,0 +1,181 @@
...
+  }
+
+  @JacocoGenerated
+  private static SqsClient defaultSqsClient() {
+    return SqsClient.builder()
+        .region(
+            new Environment().readEnvOpt(AWS_REGION_ENV).map(Region::of).orElse(Region.EU_WEST_1))
+        .httpClientBuilder(UrlConnectionHttpClient.builder())
+        .build();
+  }
```
LarsV123: Could we instead re-use the existing wrapper `no.unit.nva.publication.queue.SqsMessageSender`? Saves a little bit of boiler-plate.

template.yaml
```diff
@@ -3065,3 +3145,148 @@ Resources:
...
+  TextExtractionHandler:
+    Type: AWS::Serverless::Function
+    Properties:
+      CodeUri: publication-file-text
+      Handler: no.unit.nva.publication.file.text.TextExtractionHandler::handleRequest
+      Role: !GetAtt TextExtractionLambdaRole.Arn
+      Timeout: 900
+      MemorySize: 8192
+      EphemeralStorage:
+        Size: 10240
```
LarsV123: Does SnapStart work with this much ephemeral storage? Docs say no, but could be outdated: https://docs.aws.amazon.com/lambda/latest/dg/snapstart.html#snapstart-runtimes

In any case, these pipelines probably don't benefit from SnapStart, so it could be disabled on both the new handlers

template.yaml
```diff
@@ -3065,3 +3145,148 @@ Resources:
...
+        Rules:
+          - Id: ExpireOldObjectVersions
+            Status: Enabled
+            NoncurrentVersionExpiration:
+              NoncurrentDays: !Ref S3NoncurrentVersionExpirationDays
+      Tags:
+        - Key: IncludedInBackup
+          Value: "true"
+      VersioningConfiguration:
+        Status: Enabled
```
LarsV123: Is it worth making this versioned and including it the backup scheme, given that the contents can be regenerated?

I think a delete operation on a versioned bucket, even if targeting a key that doesn't exist in the bucket, creates a delete marker. These don't get deleted automatically by this rule and could accumulate.

The documentation seems to indicate we should add `ExpiredObjectDeleteMarker: true` to the lifecycle rule to avoid this:

- https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-properties-s3-bucket-rule.html#cfn-s3-bucket-rule-expiredobjectdeletemarker
- https://docs.aws.amazon.com/AmazonS3/latest/userguide/ManagingDelMarkers.html

publication-file-text/README.md
```diff
@@ -0,0 +1,111 @@
+# publication-file-text
+
+Extracts plain text from persisted publication files (PDF, Word, plain text,
+LaTeX) into a dedicated text bucket, so the content can be indexed or analysed
+downstream.
```
LarsV123, line 5: Does this also extract the text from embargoed/restricted files, and is that problematic? Who are the planned downstream consumers?

publication-file-text/README.md
```diff
@@ -0,0 +1,111 @@
...
+  duplicate work, never duplicate data. To recover after a dead-lettered run,
+  fix the cause and upload the CSV again.
+- **Extraction failures**: a message that keeps failing is retried up to 5
+  times by SQS, then moves to `TextExtractionDLQ`, which also has a Slack
+  alarm. A source object that no longer exists is logged and skipped, not
+  retried.
+- **Timeout**: the seeder has the Lambda maximum of 900 seconds per CSV,
+  enough for several hundred thousand keys. For larger runs, split the key
+  list across several CSV files — uploads are independent.
+
```
LarsV123: This might be tricky to debug later, because there seems to be little or no logging of what the handlers do, so we might end up with silent failures and nothing in the logs to explain why it failed. 

Suggestions:

1. Make the handlers aware of their own max runtime and abort gracefully when they get near the timeout limit, i.e. log what it was doing and then throw an exception.
2. Log breadcrumbs to track progress: Log the key of the next file first before starting processing (`TextExtractionHandler`) and log periodic batch progress ( `SeedTextExtractionHandler`).
3. Set a hard cap on the size of input files so that huge CSVs are instantly rejected instead of risking a 5x15 minute execution that times out every time.
4. Add a Slack alarm for unusually high Lambda `Duration` (if that is possible with our setup)

LarsV123 (approved):
No time to review this properly, but the approach seems sound and there's no obvious reason we can't merge it. A test run in a populated environment is the best way to evaluate it anyway.

template.yaml
brinxmat: This actually causes a lot of issues, so versioning is removed

LarsV123 (approved):

PR #2545, opened 2026-07-18
Add support for identifying pdfs that are image-only

brinxmat:
A bulk seed run produced 1 027 BLANK_CONTENT flags. Sampling showed these are predominantly image-only scans: PDFs whose pages are full-page raster images with no text layer (pdffonts empty, pdftotext empty,
  one 300 DPI image per page). Extraction was behaving correctly, but the flag conflated "scanned document that would need OCR" with "genuinely empty document", making the inventory hard to act on.            
                                                                                                                                                                                                                 
  This PR:                                                                                                                                                                                                       
                                                                                                                                                                                                                 
  - Detects image-only PDFs structurally, before parsing. A document is classified image-only when it provably has no text layer: no font resources on any page or nested form XObject (PDF text cannot be drawn 
  without a font), no interactive form fields, and at least one embedded image. Detection is best-effort — documents that cannot be inspected (e.g. encrypted) fall through to ordinary extraction and keep their
  existing classification (PASSWORD_PROTECTED, PARSE_ERROR, …).                                                                                                                                                  
  - Flags them with a new reason, IMAGE_ONLY_CONTENT, carrying the evidence as detail, e.g. Image-only PDF: 60 pages, no embedded fonts, images on 60 pages; text requires OCR. BLANK_CONTENT consequently       
  narrows to its literal meaning: extraction ran and produced nothing.                                                                                                                                           
  - Introduces the OCR seam. Image-only documents are routed to a new port, ImageOnlyPdfProcessor. The default adapter (FlaggingImageOnlyPdfProcessor) records the flag; a future OCR-backed adapter returns     
  extracted text instead — the only wiring change is one line in TextExtractors.defaultExtractors().                                                                                                             
                                                                                                                                                                                                                 
  Notes:                                                                                                                                                                                                         
                                                                                                                                                                                                                 
  - Mixed documents (text plus figures) are unaffected: any real text implies font resources, which short-circuits detection into normal extraction. This is covered by tests, including the adversarial cases   
  (text hidden in a form XObject behind a decoy image; AcroForm documents).                                                                                                                                      
  - The existing 1 027 flags predate this change. Re-seeding those keys (flag listing → CSV → seed bucket) reclassifies the scans as IMAGE_ONLY_CONTENT with per-document evidence and leaves genuinely empty    
  documents as BLANK_CONTENT, cleanly separating the two populations.


publication-file-text/src/main/java/no/unit/nva/publication/file/text/PdfScanSupport.java
```diff
@@ -0,0 +1,129 @@
...
+
+  private PdfScanSupport() {
+    // NO-OP
+  }
+
+  static Optional<PdfScanFingerprint> detectImageOnlyPdf(Path file) {
+    try (var document = Loader.loadPDF(file.toFile())) {
+      return detectImageOnlyPdf(document);
+    } catch (IOException | RuntimeException exception) {
+      LOGGER.warn("Scan detection failed, continuing with text extraction", exception);
```
LarsV123, line 48: nitpick: Might be a bit noisy to log the full stack trace at `WARN` level, but probably fine since it's a one-off job. Should we include the file path in the log message perhaps?

LarsV123 (approved):

PR #2546, opened 2026-08-03
fix: Preserve scan start marker in batch job continuation events

LarsV123:
Ticket: https://sikt.atlassian.net/browse/NP-51512

The `startMarker` in `LoadDynamodbRequest` was a `Map<String, AttributeValue>`, which Jackson cannot serialize. The `NON_EMPTY` mapper behind `JsonSerializable` silently dropped the property, so every continuation event arrived with a null marker and each segment rescanned its first page forever, queueing duplicate work messages indefinitely.

The marker is now a JSON-safe `Map<String, String>`, converted to and from `AttributeValue` at the DynamoDB boundary. Continuation events also carry an iteration counter, and the handler stops the chain with an error log once a segment exceeds 10 million scanned items, so a similar serialization bug can no longer loop unbounded.

New tests cover the JSON round trip through the production mapper and replay continuation events through the handler until the scan completes.

### DIFF
--- a/publication-event-handlers/src/main/java/no/unit/nva/publication/events/handlers/batch/dynamodb/LoadDynamodbResourceBatchJobHandler.java
+++ b/publication-event-handlers/src/main/java/no/unit/nva/publication/events/handlers/batch/dynamodb/LoadDynamodbResourceBatchJobHandler.java
@@ -15,6 +15,7 @@ import java.util.UUID;
 import no.unit.nva.commons.json.JsonUtils;
 import no.unit.nva.events.handlers.EventHandler;
 import no.unit.nva.events.models.AwsEventBridgeEvent;
+import no.unit.nva.publication.model.ScanResultWrapper;
 import no.unit.nva.publication.service.impl.ResourceService;
 import nva.commons.core.CollectionUtils;
 import nva.commons.core.Environment;
@@ -41,6 +42,7 @@ public class LoadDynamodbResourceBatchJobHandler
   private static final String TOTAL_SEGMENTS_ENV = "TOTAL_SEGMENTS";
   public static final String DETAIL_TYPE =
       "PublicationService.DataEntry.LoadDynamodbResourceBatchJob";
+  public static final int MAX_SCANNED_ITEMS_PER_SEGMENT = 10_000_000;
 
   private final SqsClient sqsClient;
   private final EventBridgeClient eventBridgeClient;
@@ -145,7 +147,7 @@ public class LoadDynamodbResourceBatchJobHandler
     var scan =
         resourceService.scanResourcesRaw(
             scanPageSize,
-            input.startMarker(),
+            input.toDynamodbStartMarker(),
             input.types(),
             input.segment(),
             input.totalSegments());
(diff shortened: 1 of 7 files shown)


publication-event-handlers/src/main/java/no/unit/nva/publication/events/handlers/batch/dynamodb/LoadDynamodbResourceBatchJobHandler.java
```diff
@@ -41,6 +42,7 @@ public class LoadDynamodbResourceBatchJobHandler
   private static final String TOTAL_SEGMENTS_ENV = "TOTAL_SEGMENTS";
   public static final String DETAIL_TYPE =
       "PublicationService.DataEntry.LoadDynamodbResourceBatchJob";
+  public static final int MAX_SCANNED_ITEMS_PER_SEGMENT = 10_000_000;
```
torbjokv, line 45: If I have a job with only one segment and our item count today is 12mill++, this may be an issue?

brinxmat (approved):

publication-event-handlers/src/main/java/no/unit/nva/publication/events/handlers/batch/dynamodb/LoadDynamodbResourceBatchJobHandler.java
```diff
@@ -162,8 +164,19 @@ private LoadDynamodbResponse processSegment(LoadDynamodbRequest input, Context c
         workItems.size(),
         messagesQueued);
 
-    if (scan.isTruncated()) {
-      sendEventForNextBatch(input, scan.nextKey(), context);
+    var totalItemsScanned = input.currentItemsProcessed() + scan.scannedCount();
+    if (scan.isTruncated() && totalItemsScanned >= MAX_SCANNED_ITEMS_PER_SEGMENT) {
+      logger.error(
```
torbjokv, line 169: should we throw exception instead so that this shows up as an alarm?

torbjokv (approved):
Looks good! a couple of suggestions added. Also, @brinxmat suggested to me that we created a idempotency-dynamodb-table that included a start.job ID, and all the pagination parameters and a TTL for 7days or so, that could be a tool to stop this issue also. Something we should think about adding in a later PR

publication-event-handlers/src/main/java/no/unit/nva/publication/events/handlers/batch/dynamodb/LoadDynamodbResourceBatchJobHandler.java
LarsV123: Yes. We should make this a request parameter with a lower default that you can override instead, but that was a larger change and I put it off for later: https://sikt.atlassian.net/browse/NP-51513

publication-event-handlers/src/main/java/no/unit/nva/publication/events/handlers/batch/dynamodb/LoadDynamodbResourceBatchJobHandler.java
LarsV123: If we throw, isn't the event just going to get retried automatically? Not ideal.

Could add an alarm to notify us if errors are logged though, is that better?

